PR #1, opened 2021-04-11
various checks

Theo47:
Added some checks on various part of the codes, that you actually have the item/skill that it is trying to use. 
(Basically what i needed in order to be able to run the script successfully from start to end, minus the parts where i solved the problem by just putting some lines into comments or where my solution wasn't general enough)

Also added some support for mumming trunk. Use the meat drop costume on your meat drop familiar. I figured this change was neutral enough that i could include it.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -9,7 +9,6 @@ import {
   getCampground,
   getCounters,
   handlingChoice,
-  haveSkill,
   itemAmount,
   mallPrice,
   myAdventures,
@@ -120,7 +119,7 @@ export function dailyFights() {
 
         // Second round of prof copies with familiar weight on.
         freeFightMood().execute(20);
-      useFamiliar($familiar`Pocket Professor`);
+        useFamiliar($familiar`Pocket Professor`);
         maximizeCached(["Familiar Weight"], { forceEquip: $items`Pocket Professor memory chip` });
         withMacro(
           Macro.trySkill("Lecture on Relativity")
@@ -475,7 +474,7 @@ const freeFightSources = [
 
 const freeKillSources = [
   new FreeFight(
-    () => !get("_gingerbreadMobHitUsed") && haveSkill($skill`Gingerbread Mob Hit`),
+    () => !get("_gingerbreadMobHitUsed") && have($skill`Gingerbread Mob Hit`),
     () =>
       withMacro(Macro.skill("Sing Along").trySkill("Gingerbread Mob Hit"), () =>
         use($item`drum machine`)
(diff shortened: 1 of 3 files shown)


src/fights.ts
```diff
@@ -311,7 +313,7 @@ const freeFightSources = [
     () => get("questL11Worship") !== "unstarted" && get("_drunkPygmyBanishes") === 10,
     () => {
       putCloset(itemAmount($item`bowling ball`), $item`bowling ball`);
-      retrieveItem(10 - get("_drunkPygmyBanishes"), $item`Bowl of Scorpions`);
+      retrieveItem(11 - get("_drunkPygmyBanishes"), $item`Bowl of Scorpions`);
```
phulin, line 316: this was on purpose - the last fight is in the saber code, it's a little awkward but i think this is the best way to do it?

src/fights.ts
```diff
@@ -119,6 +120,7 @@ export function dailyFights() {
 
         // Second round of prof copies with familiar weight on.
         freeFightMood().execute(20);
+      useFamiliar($familiar`Pocket Professor`);
```
phulin, line 123: indentation

src/fights.ts
```diff
@@ -473,7 +475,7 @@ const freeFightSources = [
 
 const freeKillSources = [
   new FreeFight(
-    () => !get("_gingerbreadMobHitUsed"),
+    () => !get("_gingerbreadMobHitUsed") && haveSkill($skill`Gingerbread Mob Hit`),
```
phulin, line 478:
```suggestion
    () => !get("_gingerbreadMobHitUsed") && have($skill`Gingerbread Mob Hit`),
```

src/fights.ts
```diff
@@ -485,7 +487,7 @@ const freeKillSources = [
   ),
 
   new FreeFight(
-    () => clamp(3 - get("_shatteringPunchUsed"), 0, 3),
+    () => (haveSkill($skill`Shattering Punch`) ? clamp(3 - get("_shatteringPunchUsed"), 0, 3) : 0),
```
phulin, line 490:
```suggestion
    () => (have($skill`Shattering Punch`) ? clamp(3 - get("_shatteringPunchUsed"), 0, 3) : 0),
```

src/index.ts
```diff
@@ -137,6 +137,11 @@ function dailySetup() {
     cliExecute("hatter 22");
   }
 
+  if (!get("_mummeryMods").includes("Meat Drop") && have($item`mumming trunk`)) {
+    useFamiliar(meatFamiliar());
+    cliExecute("mummery meat");
+  }
+
```
phulin, line 144: I added this in a recent commit, can you remove?

phulin (changes requested):
These look helpful, thanks!

src/fights.ts
Theo47: The first 10 fights are on lines 275-283, the last fight with saber code is on lines 283-308, and then this part on lines 310-316 seem to be in the case you didn't have a saber.

In this case the macro would fail, because you fetched 10-10=0 bowl of scorpions.

Theo47:
I fixed the rest of the stuff (apart from the drunken pygmy thing)

PR #4, opened 2021-05-06
Assorted fixes

### LINKED ISSUE #3: Immediate failure if no haiku katana is available
After I worked around the error in #2 by setting stashClan to something, I then get the following:
```
> call scripts\garbage-collector\garbo.js

Using adventure value 5000.
Collecting garbage!

Autoattack is disabled.
CCS set to garbo
Refreshing stash contents...
Stash list retrieved.
Pulling items from stash...
Transfer failed for haiku katana
Movement of items failed.
JavaScript exception: Script interrupted. (file:/D:/KoLmafia1/scripts/garbage-collector/garbo.js#5097)
at file:/D:/KoLmafia1/scripts/garbage-collector/garbo.js:5097 (f)
at file:/D:/KoLmafia1/scripts/garbage-collector/garbo.js:5191
at file:/D:/KoLmafia1/scripts/garbage-collector/garbo.js:5137 (withClan)
at file:/D:/KoLmafia1/scripts/garbage-collector/garbo.js:5168 (take)
at file:/D:/KoLmafia1/scripts/garbage-collector/garbo.js:5125 (withStash)
at file:/D:/KoLmafia1/scripts/garbage-collector/garbo.js:4016 (main)
```
This was run on a Pastamancer. I am assuming it wouldn't fail like that on a Seal Clubber but I'll check when I next ascend that account as one.

Phillammon:
- Update libram
- Check we have source terminal before attempting to do source terminal things
- If we lack fury, do not try to furious wallop
- Fallback to a leprechaun-like if we can't find a good rotating familiar, or failing that no familiar at all
- Actually use seal figurines instead of looping forever
- Don't try to use familiars we don't own
- Don't try to use a witchess board we don't own
- Move moveable feast stashing/unstashing where it can do less damage

UPDATE:
- Now moves moveable feast at same time as before, but
- Stash failures no longer lead to immediate abort. Resolves #3


src/combat.ts
```diff
@@ -65,7 +66,7 @@ export class Macro extends LibramMacro {
       get("retroCapeSuperhero") === "robot" &&
       get("retroCapeWashingInstructions") === "kill" &&
       itemType(equippedItem($slot`weapon`)) === "pistol";
-
+
```
phulin, line 69: Please run prettier

src/familiar.ts
```diff
@@ -89,5 +89,11 @@ export function freeFightFamiliar(): Familiar {
     familiarValue.push([$familiar`Stocking Mimic`, mimicValue]);
   }
 
+  for (const familiar of $familiars`Robortender, Hobo Monkey, Cat Burglar, Leprechaun`) {
```
phulin, line 92: Robortender should be a higher value than 1 since it drops candy and junk - maybe 200 is a good estimate?

src/fights.ts
```diff
@@ -107,6 +109,7 @@ export function dailyFights() {
         }
 
         if (
+          SourceTerminal.have() &&
```
phulin, line 112: unnecessary - this is just making sure there is no digitize monster up

src/fights.ts
```diff
@@ -256,7 +256,7 @@ const freeFightSources = [
           : $item`figurine of an ancient seal`;
       retrieveItem(1, figurine);
       retrieveItem(1, $item`seal-blubber candle`);
-      // FIXME: Some problem with mafia tracking here.
+	    use(figurine);
```
phulin, line 259: pls run prettier, tabs are messed up

src/index.ts
```diff
@@ -266,7 +266,7 @@ export function main(argString = "") {
   cliExecute("ccs garbo");
 
   // FIXME: Dynamically figure out pointer ring approach.
-  withStash($items`haiku katana, repaid diaper`, () => {
+  withStash($items`haiku katana, repaid diaper, moveable feast`, () => {
```
phulin, line 269: I don't want to take out the moveable feast for the entire duration of farming if we can avoid it. Can you change withStash to just not pull if it's not available?

phulin (changes requested):

src/index.ts
Phillammon: Addressed as of 15879b0

src/fights.ts
Phillammon: Addressed as of 15879b0

src/fights.ts
Phillammon: Addressed as of 15879b0

src/familiar.ts
Phillammon: Not yet addressed- I feel like I'd want to address this as a separate familiar valuing PR, this is definitely just a bandage to make sure that something sensible comes up. My temptation is to make this a new issue that I can work on at a later date?

src/combat.ts
Phillammon: Addressed as of 15879b0

phulin (approved):

PR #14, opened 2021-05-19
Use dark horse and preternatural greed demon if available

Phillammon:
As title suggests.


gausie (approved):
These both inspire me to do some libram features

PR #59, opened 2021-06-12
Guarantee green taffy only against a KGE

horrible-little-slime:


horrible-little-slime:
Gonna re-do this pull request in a way that doesn't suck now that I've set up some spicy little branches. Sorry!

PR #60, opened 2021-06-14
added latte support using bander-runs

horrible-little-slime:
Checks if you have a latte, and if it has the right modifiers (+40% meat, +5 familiar weight). Uses runs to unlock ingredients if necessary and able, then refills latte to be optimal for barf. Does not currently hunt down carrot.


horrible-little-slime:
burnin down the house

PR #62, opened 2021-06-14
Put digitized embezzlers in guzzlr zones

horrible-little-slime:
This does make canadv.ash a requirement, which is potentially a dealbreaker.

prepWandererZone will return noob cave or the deep dark jungle if you don't have a guzzlr or can't get a good quest

Prioritizes guzzlr quests based on getting to 196/150/30 first, then by guzzlrbuck yield

Rejects anything from the bat hole (screambats), oasis (ultrahydrated), underwater, the bubblin' caldera, or from a charter zone you don't currently have access to. Also rejects anything from down the rabbit hole, as those potions are nontradeable.

Crafts a buttery boy specifically for platinum guzzlr quests. Buys other guzzlr drinks from the mall. 

Noob cave is currently still used for backups, although if the combat rate of your guzzlr zone is 100%, that may be worth replacing as well.


horrible-little-slime:
everything must burn so a new phoenix may rise

PR #63, opened 2021-06-14
Fight protonghosts and voters during the barfturn cycle

horrible-little-slime:
Does not vote right now. Requires saucestorm to kill angry ghosts. Puts voters into noob cave, and obviously only fights free ones.


horrible-little-slime:
some sins can never be forgiven

PR #92, opened 2021-06-18
Create carpe with floundry access and use free pillkeeper

burningbman:
Pillkeeper will get another KGE if possible, otherwise go with inhaler.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -35,6 +35,7 @@ import {
   useSkill,
   visitUrl,
   wait,
+  adventure,
 } from "kolmafia";
 import {
   $class,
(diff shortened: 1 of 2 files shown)


horrible-little-slime:
In CS, you have access to neither KGE nor Inhaler. Another potential pill use could be extending a pricy potion

horrible-little-slime:
Pillkeeper semirare doesn't care about previous semirare being in that zone, so that check isn't necessary. Getting an inhaler early in the day feels suboptimal? because of limited time buffs? But that might not matter that much.

You can use questStep to ensure zone access:
questStep("questL05Goblin") >=1 for the treasury, and questStep("questL10Garbage") >= 9 for the top floor. If we end up adding canAdv, that's probably easier.

Phillammon:
Free pill superceded by #111, Carpe creation not desirable for now

PR #100, opened 2021-06-20
Sanity check valueOfAdventure

Phillammon:
If valueOfAdventure is < 3500, as is default, this will go poorly, so abort out in that situation


Tokoeka (approved):
is a Sanity Check for value of Adventure being ridiculously high also worth it?

PR #110, opened 2021-06-22
put backups in guzzlrzones if able

horrible-little-slime:
Casts musk if you can, then backs up in wanderer zone iff it's a 100% combat rate. My understanding is combatPercent takes into account your various buffs.


Phillammon:
Superceded by #111

PR #111, opened 2021-06-22
Overhaul embezzler logic

Phillammon:
Major upshot of this change is we should now be better at handling envyfish eggs without digitize and starting embezzler chains without VIP key or with the day's photocopy already used, as well as genericizing the interface to make adding more copies later easier.

Hopefully.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "dependencies": {
     "core-js": "^3.8.0",
     "kolmafia": "^1.0.11",
-    "libram": "^0.0.8",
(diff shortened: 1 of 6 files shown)


horrible-little-slime:
After diving through Scotch's log (:lewddeck:), I'd recommend the following:

 - Remove mushroom fights from daily freefights
 - Add a freerun to initialize digitize if you didn't mushroom fight
 - Pass the backup camera requirement to meatoutfit when you go underwater if you're planning on backing up there

PR #114, opened 2021-06-23
Stopped trying to shoot ghosts; radically expanded spell list used for angry ghost voters

horrible-little-slime:


src/index.ts
```diff
@@ -382,7 +383,12 @@ function barfTurn() {
       prepWandererZone(),
       Macro.if_(
         `monsterid ${$monster`Angry ghost`.id}`,
-        Macro.skill("saucestorm").repeat()
+        Macro.trySkill("curse of weaksauce")
```
Phillammon, line 386: could we incorporate a sing along?

horrible-little-slime: Sure, might as well add extract as well

horrible-little-slime: also given that this is the same PR I should perhaps name him

PR #116, opened 2021-06-23
volcano daily tasks support

horrible-little-slime:
Does not currently actually get a fused fuse at end of day with clara's bell, but expects you to. Welcoming advice on how best to do that.


dependencies.txt
```diff
@@ -1,4 +1,5 @@
 https://github.com/Ezandora/Bastille/branches/Release/
 https://github.com/Ezandora/Gain/branches/Release/
 https://github.com/Ezandora/Briefcase/branches/Release/
-https://svn.code.sf.net/p/therazekolmafia/canadv/code/
\ No newline at end of file
+https://svn.code.sf.net/p/therazekolmafia/canadv/code/
+https://github.com/covert-code/KoL-MineVolcano/tree/master
```
Phillammon, line 5: Not sure this requirement is valid

Phillammon (changes requested):

PR #125, opened 2021-06-23
get the olympic pool item if you can and haven't

horrible-little-slime:


src/dailies.ts
```diff
@@ -294,6 +294,13 @@ export function configureMisc() {
     useSkill($skill`Bind Lasagmbie`);
   }
 
+  if (
+    getClanLounge()["Olympic-sized Clan crate"] !== undefined &&
```
Phillammon, line 298: Also need to check for VIP key

PR #140, opened 2021-06-24
use snokebombs, hatred, tranqs, etc when initializing digitize and tuning latte

horrible-little-slime:


horrible-little-slime:
Just as an update for Phil (and whomstever this may concern), this has succeed on my most recent casual leg, and has done a good job initializing digitizes in CS, but it should probably be tested by someone coming fresh out of standard/QT before being pushed.

horrible-little-slime:
Runs worked fine for Beldur!

PR #144, opened 2021-06-25
add multifight support to combat.ts; add fix latte banish in freerun

horrible-little-slime:
Didn't mean for this to end up as two things, I just noticed a missing ! on my way out.


Phillammon:
Merged as part of #149

PR #145, opened 2021-06-26
Added missing parenthesis check for bander ode check/call.

BobSanders-kol:
I was getting an error message about not having a pair of stomping boots (this is true).  After looking at the code, I suspect that the free run logic is missing a parenthesis for the bandersnatch + ode check.  I didn't get to test this fix since I had run out of adv by the time I got around to looking at the error.

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -309,8 +309,8 @@ const freeRuns: freeRun[] = [
 
   new freeRun(
     () =>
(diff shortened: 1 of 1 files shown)


BobSanders-kol:
Ok, so I just installed the linter and I'll double check this in the morning since it's getting late.

PR #156, opened 2021-06-29
Pick bjorn familiars based on drops and enchantment, rather than just enchantment

horrible-little-slime:
Currently approximates value of familiar weight using a linearization based on your meat familiar's leprechaun multiplier.

May overvalue bjorn/crown in the maximizer (it gives a bonus on top of its regular score), but bjorn is pretty much always worth wielding anyways.

Bjorn-familiar list is not exhaustive, but covers most bases.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -264,7 +264,7 @@ export function main(argString = "") {
     }
 
     // FIXME: Dynamically figure out pointer ring approach.
-    withStash($items`haiku katana, repaid diaper`, () => {
+    withStash($items`haiku katana, repaid diaper, buddy bjorn`, () => {
       // 0. diet stuff.
       runDiet();
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -264,7 +264,7 @@ export function main(argString = "") {
     }
 
     // FIXME: Dynamically figure out pointer ring approach.
-    withStash($items`haiku katana, repaid diaper, buddy bjorn`, () => {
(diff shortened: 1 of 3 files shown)


horrible-little-slime:
re: Phill's concern with the professor maximization part, I would use something like
```typescript
maximize("familiar weight, +equip pocket professor memory chip", false);
const expectedLectures = 2 + Math.ceil(Math.sqrt((have($skill`Meteor Lore`) && get("_meteorShowerUses") < 5 ? 20 : 0) + familiarWeight($familiar`pocket professor`) + weightAdjustment()))
const weightNeeded = 1 + Math.pow((expectedLectures - 3), 2) - (have($skill`Meteor Lore`) && get("_meteorShowerUses") < 5 ? 20 : 0);
maximize(`42069 familiar weight max ${weightNeeded}, ${whatever the hell else goes here}`);
```

probably using maximizeCached and/or meatOutfit. I'm not exactly sure how to handle bjorn here, but it shouldn't be too difficult to maximize. We could even just `-back` if the player has a bjorn and fill it in ourselves.

src/index.ts
```diff
@@ -264,7 +264,7 @@ export function main(argString = "") {
     }
 
     // FIXME: Dynamically figure out pointer ring approach.
-    withStash($items`haiku katana, repaid diaper`, () => {
+    withStash($items`haiku katana, repaid diaper, buddy bjorn`, () => {
```
Phillammon, line 267:
```suggestion
    withStash($items`haiku katana, repaid diaper, buddy bjorn, crown of thrones`, () => {
```

PR #157, opened 2021-06-29
Property wrapper

### LINKED ISSUE #35: Set satisfywithNPCs/mall/coinmasters appropriately

horrible-little-slime:
Made an additional function for choice adventures that accepts a map input. Turns out not to be that much easier than just defining the property objects.

Resolves #35


horrible-little-slime:
Note: adding the foldable stuff to meatOutfit is causing issues with the wad of used tape--any advice on how best to proceed here? I could have it fold a wad of used tape after the drum machine fights?

PR #158, opened 2021-06-30
use horrible array methods to decide the best meatFamiliar()

horrible-little-slime:
If you decide this is too intensive to be remotely worth using, that's life! Just wanted to put it on the table.


horrible-little-slime:
Sorts by leprechaun coefficient, then by fairy coefficient.

PR #160, opened 2021-06-30
Profchain only if you can

horrible-little-slime:
After my CS and casual loops, I typically find garbo trying to do the first profchain without enough weight to lecture properly. Now, before each profchain, we'll check whether we've added enough weight to do more lectures. If not, we'll revert to a regular meat familiar and a regular meat outfit, and still mark that particular profchain as complete.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -400,16 +400,14 @@ export function dailyFights() {
           new Requirement([], { forceEquip: $items`Pocket Professor memory chip` }),
         ]);
         if (
-          get("_pocketProfessorLectures") >=
+          get("_pocketProfessorLectures") <
           2 + Math.ceil(Math.sqrt(familiarWeight(myFamiliar()) + weightAdjustment()))
         ) {
-          set("_garbo_meatChain", true);
-        } else {
           withMacro(firstChainMacro(), () =>
             fightSource.run({ location: prepWandererZone(), macro: firstChainMacro() })
           );
-          set("_garbo_meatChain", true);
         }
+       set("_garbo_meatChain", true);
       }
 
       startDigitize();
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -407,7 +407,7 @@ export function dailyFights() {
             fightSource.run({ location: prepWandererZone(), macro: firstChainMacro() })
           );
         }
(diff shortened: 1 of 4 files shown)


src/fights.ts
```diff
@@ -393,10 +399,17 @@ export function dailyFights() {
...
+          get("_pocketProfessorLectures") >=
+          2 + Math.ceil(Math.sqrt(familiarWeight(myFamiliar()) + weightAdjustment()))
+        ) {
+          set("_garbo_meatChain", true);
+        } else {
+          withMacro(firstChainMacro(), () =>
+            fightSource.run({ location: prepWandererZone(), macro: firstChainMacro() })
+          );
+          set("_garbo_meatChain", true);
+        }
```
Phillammon, line 412:
```suggestion
        if (
          get("_pocketProfessorLectures") <
          2 + Math.ceil(Math.sqrt(familiarWeight(myFamiliar()) + weightAdjustment()))
        ) {
          withMacro(firstChainMacro(), () =>
            fightSource.run({ location: prepWandererZone(), macro: firstChainMacro() })
          );
        }
       set("_garbo_meatChain", true);
```

src/fights.ts
```diff
@@ -413,10 +426,17 @@ export function dailyFights() {
...
+          get("_pocketProfessorLectures") >=
+          2 + Math.ceil(Math.sqrt(familiarWeight(myFamiliar()) + weightAdjustment()))
+        ) {
+          set("_garbo_weightChain", true);
+        } else {
+          withMacro(secondChainMacro(), () =>
+            fightSource.run({ location: prepWandererZone(), macro: secondChainMacro() })
+          );
+          set("_garbo_weightChain", true);
+        }
```
Phillammon, line 439:
```suggestion
        if (
          get("_pocketProfessorLectures") <
          2 + Math.ceil(Math.sqrt(familiarWeight(myFamiliar()) + weightAdjustment()))
        ) {
          withMacro(secondChainMacro(), () =>
            fightSource.run({ location: prepWandererZone(), macro: secondChainMacro() })
          );
        }
        set("_garbo_weightChain", true);
```

Phillammon (changes requested):

PR #161, opened 2021-06-30
fold a putty sheet before attempting to stash, if relevant; never equip a spooky putty leotard

horrible-little-slime:


Phillammon:
Closed per discussion in discord

PR #165, opened 2021-07-01
change meatVal to an arrow function

horrible-little-slime:


horrible-little-slime:
Redundant with #166 --we're just too good!

PR #168, opened 2021-07-02
prevent equipping crown as part of freeFightOutfit as well as meatOutfit

horrible-little-slime:


horrible-little-slime:
This seems to work when part of captain-planet, but I'll level with you, I'm only 69% sure I'm doing things the right way here.

PR #170, opened 2021-07-02
cache bjorn choices

horrible-little-slime:
goal here is to reduce execution speed for the outfit functions


horrible-little-slime:
Testing shows this is functional, but because .filter() doesn't mutate the array, it doesn't quite behave as intended. Very interested in solutions to this problem.

Phillammon:
Superceded by #177

PR #186, opened 2021-07-07
pantsgiving support

### LINKED ISSUE #77: No more Pantsgiving handling
I thought it used to run Pantsgiving to get the +1 or +2 fullness during the free fight handling, and it no longer does so. It just does pantogram and leaves those on the entire time.

horrible-little-slime:
Calculate value of pantsgiving assuming horseradish with seasoning and sans-mayozapine; verify we have enough turns to get an extra fullness, and add as a bonus equip.

Add a horseradish() function after just about every fight I can find. There must be a better way to do that, though.

Resolves #77

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -96,6 +96,7 @@ export class Macro extends LibramMacro {
       .externalIf(opsSetup, Macro.skill("Throw Shield").attack())
       .externalIf(katanaSetup, Macro.skill("Summer Siesta").attack())
       .externalIf(capeSetup, Macro.skill("Precision Shot"))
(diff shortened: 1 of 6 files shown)


horrible-little-slime:
As of 0536b00, this is functioning but has two issues that I'd like to resolve before merging:
1. It is not particularly resilient to bugs--if mafia starts miscounting _pantsgivingCount, it may end up burning a bunch of turns on pantsgiving charging without actually getting the fullness from it.
2. I feel like this adds a hefty computational burden each turn, although it doesn't feel noticeably slower. Then again, as my mother always says, "sometimes your feelings are wrong."

PR #221, opened 2021-07-09
adding a config ui

JamesDowney:
Just adding a simple configuration UI ripped straight from autoscend (well, actually from ttpack but there's a lineage there).


Phillammon:
Superceded by #241

PR #225, opened 2021-07-10
Add VIP Clan concept, properly namespace clan prefs

Phillammon:


Tokoeka (approved):
Looks good. 

my first concern of "could this return stash items to the VIP clan" definitely feels unfounded after looping through the logic of the various functions.

PR #227, opened 2021-07-10
stinky-cheese

horrible-little-slime:


horrible-little-slime:
As of 639b003, this appears functional:
 - Folding with more than 1 item is untested, I'll need to borrow 2 additional stinky cheeses to make that work
 - One concern I have is that it can be hard to build up enough cheese charge to make it worth using in barf, and it may be worth rebalancing our weighting of it. But, if I do that, I'd like to discuss with someone what logic we should use when valuing cheese
  - Mafia overcounted my cheesecharge by a little bit, and it may be worth migrating to numericModifier

PR #231, opened 2021-07-12
Add manual tracking of mime army shotglass

pstalcup:


pstalcup:
I did not track if the user has mime army shot glass. Is there any other eason why drunk would be one less after consumption? I can add a check for it

Tokoeka (approved):

gausie:
> I did not track if the user has mime army shot glass. Is there any other eason why drunk would be one less after consumption? I can add a check for it

No but there could be in the future

PR #242, opened 2021-07-13
upgrade libram, update guzzlr and withProperty to work with latest libram

horrible-little-slime:


horrible-little-slime:
This appears to work now

PR #247, opened 2021-07-14
Added hobo monkey stasis

BobSanders-kol:
Added support for hobo monkey stasis.

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -71,7 +71,7 @@ export class Macro extends LibramMacro {
       get("retroCapeWashingInstructions") === "kill" &&
       itemType(equippedItem($slot`weapon`)) === "pistol";
 
-    // TODO: Hobo monkey stasis. VYKEA couch issue. Probably other stuff.
+    // TODO: VYKEA couch issue. Probably other stuff.
     return this.tryHaveSkill("Sing Along")
       .externalIf(
         shouldRedigitize(),
@@ -91,6 +91,13 @@ export class Macro extends LibramMacro {
       .externalIf(
         myFamiliar() === $familiar`Stocking Mimic`,
         Macro.skill("Curse of Weaksauce").while_("!pastround 10", Macro.item("seal tooth"))
+      )
+      .externalIf(
+        myFamiliar() === $familiar`Hobo Monkey`,
+        Macro.while_(
+          `!match "shoulder, and hands you some Meat." && !pastround 20`,
+          Macro.item("seal tooth")
+        )
       )
       .externalIf(sealClubberSetup, Macro.trySkill("Furious Wallop").attack())
       .externalIf(opsSetup, Macro.skill("Throw Shield").attack())
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -5,6 +5,7 @@ import {
   haveSkill,
   inMultiFight,
   itemType,
+  monsterHp,
   myAdventures,
   myClass,
   myFamiliar,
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
Do we want the stasis to occur before the crit? How many rounds of VYKEA do we expect it to survive?

How computationally intensive would it be to use our VYKEA level in conjunction with our ML to predict how many turns we can afford to stasis?

BobSanders-kol:
I don't think we can do the crit before the stasis, since my crits wind up killing the tourists.  Unless there's something I'm missing (there usually is).

And good catch on the couch damage, I hadn't though about that.  Would something like this work:
```javascript
const stasisRounds =
      get("_VYKEACompanionLevel") > 0 ? clamp(monsterHp() / (10 * get("_VYKEACompanionLevel") + 10) - 1, 1, 20): 20;
```
I'd then use stasisRounds instead of the hard coded 20 I have in `!pastround`.

horrible-little-slime:
meatKill isn't called in combat, so you can't use monsterHp(). Instead I'd do `(100 + numericModifier("Monster Level"))*(100 / (100 - Math.min(0.4*numericModifier("Monster Level"), 50)))` as the HP, accounting for the damage reduction ML provides. This may be getting way too complicated, though.

phulin:
I think I'd make stasis conditional on having a Manuel and use the
monsterhpbelow predicate.

BobSanders-kol:
So maybe something like this then:
```javascript
const vykeaMaxDamage = 10 * get("_VYKEACompanionLevel") + 10;
```
```javascript
.externalIf(
        myFamiliar() === $familiar`Hobo Monkey` && have($item`Monster Manuel`),
        Macro.while_(
          `!match "shoulder, and hands you some Meat." && !pastround 20 && monsterhpabove ` +
            (vykeaMaxDamage + 5),
          Macro.item("seal tooth")
        )
      )
```
Pending testing, of course.  And I can put off the statsis for us non-manuel having folks.

The + 5 is just a buffer for our seal tooth damage + some extra HP just in case.

horrible-little-slime:
this is a real nitpick but vykeadamage should be 0 if its level is 0. alternately, check that vykea level isn't 0 during the externalIf.

Other than that, chefkiss

Rinn:
have($item`Monster Manuel`) isn't going to work since the item is consumed on use, mafia should really return it as an entry in getCampground() but it doesn't right now.

BobSanders-kol:
> have($item`Monster Manuel`) isn't going to work since the item is consumed on use, mafia should really return it as an entry in getCampground() but it doesn't right now.

I had only realized the have($item`Monster Manuel`) wouldn't work after finally purchasing one myself.

My last attempt at this feature looked like this:
```typescript
.externalIf(
        myFamiliar() === $familiar`Hobo Monkey` &&
          monsterFactoidsAvailable($monster`garbage tourist`, true) > 0 &&
          !haveEffect($effect`Just the Best Anapests`) &&
          equippedAmount($item`haiku katana`) === 0 &&
          myLocation() === $location`Barf Mountain`,
        Macro.while_(
          `!match "shoulder, and hands you some Meat." && !pastround 20 && monsterhpabove ` +
            (vykeaMaxDamage + 5),
          Macro.item("seal tooth")
        )
      )
```

`monsterFactoidsAvailable` didn't seem to work reliably without also calling `flushMonsterManuelCache();` at some point.  Both of these functions are not documented on the KoLmafia wiki, so using them may be unsupported?  

The extra checks for katana, anapests, and location were due to possibly not being able to correctly parse the `match` statement in the macro.

Considering the Manuel limitations and potential for `match` failure I may close the pull request soon, since I'm not making much headway and these changes seem somewhat controversial?  I welcome any thoughts from others.

Rinn:
I can take care of this in my combat improvement PR #334

PR #249, opened 2021-07-14
add takeStorage to diet acquire function

lewismd13:
checking hagnk's for diet things so it doesn't buy from mall unnecessarily


horrible-little-slime (approved):
I wonder if I have the authority to approve things? Let's find out!

Manny tested this locally.

Issue #199, opened 2021-07-09
[resource] utilize libram summons


docrostov:
garbo should probably try to do minimum-viable-product libram summoning; if you call breakfast before/after garbo you will get some libram summons per your internal mafia preferences but libram stuff can be some decent extra meat depending on how much MP you can generate freely/cheaply, and garbo does generate a lot of "spare" MP. as with tome summons, this could be immensely complicated or super simple, just depends on your goals and usage parameters.

Rinn:
Limited rares
- Bricko: 50% for first eye brick (+2 brick), 33% for second and third, max of _brickoEyeSummons=3
- Taffy: 50% / (_taffyRareSummons+1), max of _taffyYellowSummons=1 and _taffyRareSummons=4
Unlimited
- Bricko: 100% 3x Bricko Brick
- Candy Hearts: equal chance
- Party Favors: 50% / (_favorRareSummons+1) chance of any rare, no lower limit, equal cahnc
- Love songs: equal chance of any
- Dice: equal chance of any
- Resolutions: 2% chance of any rare
- Taffy: equal chance 
Maybe if the estimated price is below some threshold it rotates through the librams just for more variety. Mafia doesn't track the number of times each libram has been cast individually though.

Imo i'd maybe just prioritize getting all the limited rares first since garbo fights brickos and uses a green taffy copy if you have robort.

Rinn:
It might be better if libram had a function that returned an array of expected result chance for a summon

horrible-little-slime:
This is not fully closed, but it can be soon; thanks to @Rinn's libram support in libram, we now burn MP before using PYEC and LTC, and can expand that code to other places.

horrible-little-slime:
Closing this as an issue; it'd be neat to milk more out of it, but we don't have to.

PR #254, opened 2021-07-16
Eat remaining sausage

BobSanders-kol:
I often wind up only eating a few sausages (since I don't have Bind Lasagmbie perm'd/other MP intensive scripts running), so I added code to eat more sausages if we're at the end of the garbo run.  Tested this today and it ran successfully.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,6 +2,7 @@ import {
   booleanModifier,
   buy,
   cliExecute,
+  eat,
   getCampground,
   getCounters,
   guildStoreAvailable,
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
Do we always want to eat 23 sausages? Should we check to see if the cost of creating a sausage is less than the valueOfAdventure?

BobSanders-kol:
We may not want to eat 23 sausages every run, but I figured adding this when myAdventure() === 1 was somewhat safe and we're using, or have "issues" for using, other IotM resources in similar fashion.  I suppose this could be made a config param once those are more fleshed out.  There could be a looping issue I'm unaware of as well.  I'm good with adjusting this if need be, but I figure it's better to use them than lose them?

For valueOfAdventure, we already check if it's below 3500, so 2553 meat for the 23rd sausage should hopefully be net positive.

horrible-little-slime:
Yep I'm extremely bad at math and in my head always thought the sausages went much higher in price that's fine

gausie:
We should incorporate this into some MP libram farming

BobSanders-kol:
> We should incorporate this into some MP libram farming

Looks like scotch has an open issue (#199) for Libram summons.  Sadly I don't own any librams, so I'll let someone else tackle this one.

PR #262, opened 2021-07-17
retrieve seal tooth and the jokester's gun to reduce errors

horrible-little-slime:


src/index.ts
```diff
@@ -259,6 +259,9 @@ export function main(argString = "") {
       ? 1
       : 0;
 
+  retrieveItem($item`seal tooth`);
+  retrieveItem($item`the jokester's gun`);
```
phulin, line 263:
```suggestion
  retrieveItem($item`The Jokester's gun`);
```

src/index.ts
```diff
@@ -259,6 +259,9 @@ export function main(argString = "") {
       ? 1
       : 0;
 
+  retrieveItem($item`seal tooth`);
+  retrieveItem($item`The Jokester's gun`);
```
phulin, line 263: I think these should be in dailySetup, no?

horrible-little-slime: moving to be just below the half a purse. Good catch.

phulin (approved):

PR #264, opened 2021-07-17
Basic potion support.

phulin:
Support for dynamically picking potions during embezzler fights and afterwards.


Phillammon (approved):

src/potions.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+}
+
+export function bathroomFinance(embezzlers: number): void {
+  if (have($effect`Buy! Sell! Buy! Sell!`)) return;
+  // Average gross is 1025 * (embezzlers + 1) / 2 for embezzlers turns.
+  const averageEmbezzlerGross = ((baseMeat + 750) * 2 * (embezzlers + 1)) / 2 / 100;
+  const embezzlerGross = averageEmbezzlerGross * embezzlers;
+  const tourists = 100 - embezzlers;
+  const averageTouristGross = (baseMeat * 2 * (100 + embezzlers + 1)) / 2 / 100;
```
horrible-little-slime, line 204: should this be 100 - embezzlers?

src/potions.ts
```diff
@@ -0,0 +1,211 @@
...
+      if (potion.doublingValue(embezzlers) > 10000) {
+        cliExecute("pillkeeper extend");
+        potion.useAsValuable(embezzlers, true);
+      }
+    }
+  }
+
+  // Only test potions which are reasonably close to being profitable using historical price.
+  const testPotions = meatPotions.filter(
+    (potion) => potion.gross(embezzlers) / potion.price(true) > 0.5
```
horrible-little-slime, line 186: Should this be > 1? Should we use subtraction comparing to 0 instead of division?

src/potions.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  doublingValue(embezzlers: number, historical = false) {
+    return (
+      Math.max(this.net(embezzlers, true, historical), 0) -
+      Math.max(this.net(embezzlers, false, historical), 0)
+    );
+  }
+
+  useAsValuable(embezzlers: number, doubleDuration = false) {
+    const duration = this.effectDuration() * (doubleDuration ? 2 : 1);
```
horrible-little-slime, line 93: Should we consider the value of PYEC extension? LTC extension?

src/potions.ts
```diff
@@ -0,0 +1,211 @@
...
+        "blue"
+      );
+      quantityToUse += additional;
+      embezzlersRemaining = Math.max(embezzlersRemaining - additional * duration, 0);
+    }
+    if (embezzlersRemaining > 0 || (doubleDuration && quantityToUse > 0)) keepGoing = false;
+
+    // How many should we use with non-embezzlers?
+    if (keepGoing && this.net(0, doubleDuration) > 0) {
+      const adventureCap = myAdventures() * 1.04 + 50;
```
horrible-little-slime, line 134: consider dropping the +50 if not runnning `garbo ascend`

src/potions.ts
```diff
@@ -0,0 +1,211 @@
...
+  // TODO: Count free fights (25 meat each for most).
+  const embezzlers = embezzlerCount();
+  const potions = Item.all().filter(
+    (item) => item.tradeable && !banned.includes(item) && itemType(item) === "potion"
+  );
+  const meatPotions = potions
+    .map((item) => new Potion(item))
+    .filter((potion) => potion.bonusMeat() > 0);
+
+  if (have($item`Eight Days a Week Pill Keeper`) && !get("_freePillKeeperUsed")) {
```
horrible-little-slime, line 166: Given that we're (presumably) calling this before doing all of our embezzler fights, we want to make sure to only do this if it's more valuable than another embezzler. First, check if we have treasury access (canAdv or questStep), and if we do, we'll want to see how much meat an additional embezzler is worth.

horrible-little-slime: If we decide this is too hard to calculate I think it's reasonable to assume that you want to get the embezzler if you can. "I think" is pulling a lot of weight here, though.

src/potions.ts
phulin: Yeah, this is the formula used elsewhere - we should have a global thing that tracks this.

src/potions.ts
phulin: This is trying to do a rough filtering using historical price, and then do a precise check later using current price. So it checks to see if it's within a factor of 2 of being profitable.

src/potions.ts
phulin: Later I think

src/potions.ts
phulin: I put in 10,000 as a guess at the value of an additional embezzler. I should add the canAdv check too.

phulin:
Note to self: fix snowcones before merging

horrible-little-slime (approved):

Issue #207, opened 2021-07-09
[resource] rain doh management


docrostov:
garbo currently does use spooky putty for extra embezzlers, but i do not think it applies the weirder usage case, where you have both putty and rain doh and you can get an extra +1 copy by using both. (i have no idea why tptb coded it like this, it is a ridiculous thing). 

one issue to keep in mind here is that rain doh is harder to utilize via stash, so it is possible that this is kind of out of scope, but if the user -has- both already it should probably use it for +1 embezzie. idk. very minor issue (as are most of the resource tagged issues) but possibly worth some hacky code.

PR #266, opened 2021-07-17
Add Rain-Doh black box, add Macro.tryCopier() function

Rinn:
- Macro.tryCopier() checks various conditions on whether or not a copier item or skill should be used
- Add support for Rain-Doh black box
- Use a spooky putty monster if it is invalid after ascending

Ran two days and it seems to be working correctly.

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -59,14 +59,12 @@ export class Macro extends LibramMacro {
     switch (itemOrSkill) {
       case $item`Spooky Putty sheet`:
         return this.externalIf(
-          get("spookyPuttyCopiesMade") + Math.max(1, get("_raindohCopiesMade")) < 6 &&
-            !have($item`Spooky Putty monster`),
+          get("spookyPuttyCopiesMade") + Math.max(1, get("_raindohCopiesMade")) < 6,
           Macro.tryItem(itemOrSkill)
         );
       case $item`Rain-Doh black box`:
         return this.externalIf(
-          get("_raindohCopiesMade") + Math.max(1, get("spookyPuttyCopiesMade")) < 6 &&
-            !have($item`Rain-Doh box full of monster`),
+          get("_raindohCopiesMade") + Math.max(1, get("spookyPuttyCopiesMade")) < 6,
           Macro.tryItem(itemOrSkill)
         );
       case $item`4-d camera`:
@@ -92,7 +90,7 @@ export class Macro extends LibramMacro {
           Macro.tryItem(itemOrSkill)
         );
       case $item`print screen button`:
-        return this.externalIf(!have($item`screencapped monster`), Macro.tryItem(itemOrSkill));
+        return this.tryItem(itemOrSkill);
       case $item`alpine watercolor set`:
         return this.externalIf(
           monster !== $monster`none` && ChateauMantegna.paintingMonster() !== monster,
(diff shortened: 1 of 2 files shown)


src/fights.ts
```diff
@@ -370,6 +373,11 @@ function embezzlerSetup() {
   if (!get("_cameraUsed") && !have($item`shaking 4-d camera`)) {
     retrieveItem($item`4-d camera`);
   }
+
+  // Fix invalid spooky putty monster (caused by ascending)
```
horrible-little-slime, line 377: You may want to examine the spooky putty monster here, to make sure mafia knows what's in there.

Rinn: yeah i'll add that

src/combat.ts
```diff
@@ -53,6 +55,67 @@ export class Macro extends LibramMacro {
     return new Macro().tryHaveSkill(skillOrName);
   }
 
+  tryCopier(itemOrSkill: Item | Skill, monster: Monster = $monster`none`): Macro {
+    switch (itemOrSkill) {
+      case $item`Spooky Putty sheet`:
+        return this.externalIf(
+          get("spookyPuttyCopiesMade") < 5 && !have($item`Spooky Putty monster`),
```
soolar, line 62: This should also check to make sure `spookyPuttyCopiesMade + _raindohCopiesMade < 6`, shouldn't it? And same for below with black box.

horrible-little-slime:
Resolves #207

src/combat.ts
```diff
@@ -53,6 +53,65 @@ export class Macro extends LibramMacro {
...
+      case $item`alpine watercolor set`:
+        return this.tryItem(itemOrSkill);
+      case $item`LOV Enamorang`:
+        return this.externalIf(
+          get("_enamorangs") < 5 && get("enamorangMonster") === $monster`none`,
+          Macro.tryItem(itemOrSkill)
+        );
+      case $skill`Digitize`:
+        return this.externalIf(
+          get("_sourceTerminalDigitizeUses") <
```
horrible-little-slime, line 99: This is redundant with trySkill--digitize shouldn't show up if you are out of digitizes

src/combat.ts
```diff
@@ -53,6 +53,65 @@ export class Macro extends LibramMacro {
...
+          !get("_envyfishEggUsed") && !have($item`envyfish egg`),
+          Macro.tryItem(itemOrSkill)
+        );
+      case $item`print screen button`:
+        return this.tryItem(itemOrSkill);
+      case $item`alpine watercolor set`:
+        return this.tryItem(itemOrSkill);
+      case $item`LOV Enamorang`:
+        return this.externalIf(
+          get("_enamorangs") < 5 && get("enamorangMonster") === $monster`none`,
```
horrible-little-slime, line 94: enamorangMonster will never be `none`--it will be null or undefined, instead.

src/fights.ts
```diff
@@ -379,6 +385,16 @@ function embezzlerSetup() {
...
+  if (have($item`Spooky Putty monster`) && get("spookyPuttyMonster") === $monster`none`) {
```
horrible-little-slime, line 390: A similar thing can happen to rain-doh if you use the copier while wearing papier-mache or other text-obfuscating conditions--is it worth adding a refresher there?

src/combat.ts
Rinn: My thought was to reduce the size of the macro that's being submitted, but yeah you're right it's technically not necessary.

horrible-little-slime (approved):
I am slightly concerned about using putty and the box in the same fight for reasons discussed in discord, but for reasons discussed in discord it is unlikely to actually lead to explosions.

PR #268, opened 2021-07-18
digitized wanderers override sausages, so we don't want to kramco in the middle of that.

horrible-little-slime:


src/fights.ts
```diff
@@ -530,7 +530,11 @@ export function dailyFights() {
         });
         startDigitize();
         nextFight = getEmbezzlerFight();
-        if (kramcoGuaranteed() && (!nextFight || nextFight.name !== "Backup")) doSausage();
+        if (
+          kramcoGuaranteed() &&
+          (!nextFight || (nextFight.name !== "Backup" && nextFight.name !== "Digitize"))
+        )
+          doSausage();
       }
```
phulin, line 538:
```suggestion
        ) {
          doSausage();
        }
      }
```

Tokoeka (approved):

PR #270, opened 2021-07-18
estimate adventure count better--care about having a thumb ring and about the ascend flag

horrible-little-slime:


horrible-little-slime:
This unfortunately makes garbo explode

PR #274, opened 2021-07-19
make distinction between backups and digitizes for prepWandererZone

horrible-little-slime:


src/lib.ts
```diff
@@ -140,50 +141,89 @@ function acceptBestGuzzlrQuest() {
...
-      if (!guzzlrBooze) {
-        return defaultLocation;
-      } else if (!have(guzzlrBooze)) {
-        print(`just picking up some booze before we roll`, "blue");
-        cliExecute(`acquire ${guzzlrBooze}`);
+        use(1, place.potion);
       }
+    });
+    if (!Guzzlr.havePlatinumBooze()) {
+      cliExecute("make buttery boy");
```
phulin, line 209:
```suggestion
      create($item`Buttery Boy`);
```

src/lib.ts
```diff
@@ -140,50 +141,89 @@ function acceptBestGuzzlrQuest() {
...
+    }
+  } else {
+    const guzzlrBooze = Guzzlr.getBooze();
+    if (!guzzlrBooze) {
+      return defaultLocation;
+    } else if (!have(guzzlrBooze)) {
+      print(`just picking up some booze before we roll`, "blue");
+      cliExecute(`acquire ${guzzlrBooze}`);
```
phulin, line 217:
```suggestion
      retrieveItem(guzzlrBooze);
```

src/lib.ts
```diff
@@ -138,50 +141,89 @@ function acceptBestGuzzlrQuest() {
   }
 }
 
-export function prepWandererZone(): Location {
+export enum draggableFight {
+  BACKUP,
+  WANDERER,
+}
+
+export function prepWandererZone(type: draggableFight = draggableFight.WANDERER): Location {
```
phulin, line 149: Can we rename this like "pickWandererZoneAndPrep"? Current name is a little unclear

src/lib.ts
```diff
@@ -975,3 +1016,10 @@ export function kramcoGuaranteed(): boolean {
       5 + 3 * get("_sausageFights") + Math.pow(Math.max(0, get("_sausageFights") - 5), 3)
   );
 }
+
+export function canWander(): boolean {
```
phulin, line 1020: also no idea what this is supposed to be for?

src/lib.ts
```diff
@@ -138,50 +141,89 @@ function acceptBestGuzzlrQuest() {
...
+      print(`just picking up some booze before we roll`, "blue");
+      retrieveItem(guzzlrBooze);
     }
-    return guzzlZone;
   }
+  return guzzlZone;
 }
 
-function guzzlrCheck() {
+function guzzlrCheck(): boolean {
```
phulin, line 223: not sure what this does either per the name

src/lib.ts
horrible-little-slime: aye aye

src/lib.ts
horrible-little-slime: renamed to checkGuzzlrZone. Is that more clear? The intent is that this tells us if we categorically hate a particular guzzlr zone. It does not reject zones that we only hate for wanderers or that we only hate for backups.

horrible-little-slime: I thought about doWeHateOurGuzzlrZone but thought that might be too fanciful. But that's the next option in line for names if checkGuzzlrZone is still unclear

src/lib.ts
horrible-little-slime: this gets used to determine if we are going to fight any wanderers that we might put in guzzlrZones, because that affects our decisions about whether to reject a guzzlr quest. It was suggested to me that we should yoink that out and put it in its own function for legibility, and because it could potentially be useful elsewhere. Just now realizing I should add a little something to make sure you have voters left.

horrible-little-slime:
I am burning this down to make way for a bright, beautiful new future

PR #275, opened 2021-07-19
Use react for relay browser

burningbman:
Use react for rendering the relay browser

### DIFF
--- a/src/relay/App.tsx
+++ b/src/relay/App.tsx
@@ -1,0 +1,34 @@
+import * as React from "react";
+import Setting from "./components/Setting";
+
+class App extends React.Component<Data> {
+  constructor(props: Data) {
+    super(props);
+  }
+  render(): React.ReactNode {
+    const elements = this.props.settings.map((setting) => {
+      return (
+        <Setting
+          value={setting.value}
+          type={setting.type}
+          name={setting.name}
+          description={setting.description}
+        />
+      );
+    });
+    return (
+      <div id="garbageCollectorContainer">
+        <h1> Garbage Collector Configuration </h1>
+        <form action="" method="post">
+          <table>{elements}</table>
+          <input type="submit" name="" value="Save Changes" />
+        </form>
+        <h2>Info</h2>
+        garbo version: {this.props.rev}
+        <br />
+      </div>
+    );
+  }
+}
+
+export default App;
--- a/src/relay/App.tsx
+++ b/src/relay/App.tsx
@@ -1,8 +1,26 @@
 import * as React from "react";
 import Setting from "./components/Setting";
 
-function App(props: Data) {
-  const elements = props.settings.map((setting) => (
+declare global {
+  function getData(callback: (data: Data) => void): void;
+  type SettingProp = {
+    name: string;
+    type: string;
+    description: string;
+    value: string | number;
+  };
+  type Data = {
+    settings: SettingProp[];
+    updatedSettings: UpdatedSetting[];
+  };
+  type UpdatedSetting = {
+    name: string;
+    value: string;
+  };
+}
+
+function App({ settings, updatedSettings }: Data) {
+  const preferences = settings.map((setting) => (
     <Setting
       value={setting.value}
       type={setting.type}
@@ -10,16 +28,21 @@ function App(props: Data) {
       description={setting.description}
     />
   ));
+
+  const updatedPreferences = updatedSettings.map((setting) => (
+    <div>
+      Changing setting {setting.name} to {setting.value}
+    </div>
+  ));
+
   return (
     <div id="garbageCollectorContainer">
       <h1> Garbage Collector Configuration </h1>
+      {updatedPreferences}
       <form action="" method="post">
-        <table>{elements}</table>
+        <table>{preferences}</table>
         <input type="submit" name="" value="Save Changes" />
       </form>
-      <h2>Info</h2>
-      garbo version: {props.rev}
-      <br />
     </div>
   );
 }
--- a/src/relay/App.tsx
+++ b/src/relay/App.tsx
@@ -1,34 +1,27 @@
 import * as React from "react";
 import Setting from "./components/Setting";
 
-class App extends React.Component<Data> {
-  constructor(props: Data) {
-    super(props);
-  }
-  render(): React.ReactNode {
-    const elements = this.props.settings.map((setting) => {
-      return (
-        <Setting
-          value={setting.value}
-          type={setting.type}
-          name={setting.name}
-          description={setting.description}
-        />
-      );
-    });
-    return (
-      <div id="garbageCollectorContainer">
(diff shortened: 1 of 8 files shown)


src/relay/components/Setting.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
+
+class Setting extends React.Component<SettingProp> {
```
phulin, line 3: I strongly recommend using functional components instead of class components. The difference is really night and day.

src/relay_garbo.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  writeln('<head><link rel="stylesheet" href="/garbage-collector/main.css"></head>');
+  writeln('<div id="root"></div>');
+
+  // add script that react calls when loaded to get kol data
+  writeln("<script>");
+  writeln(
+    `let getData = function(callback) {callback(${JSON.stringify({
+      rev: svnInfo("garbage-collector").last_changed_rev,
```
phulin, line 46: Does this work with git-based scripts?

webpack.config.js
```diff
@@ -31,12 +21,60 @@ module.exports = {
...
+      path: path.resolve(__dirname, "KoLmafia", "scripts", "garbage-collector"),
+      filename: "[name].js",
+      libraryTarget: "commonjs",
+    },
+  },
+  sharedConfig
+);
+
+// handle the file creating the garbo UI html file
+const otherRelayConfig = Object.assign(
```
phulin, line 55: I think you want webpack-merge here.

webpack.config.js
```diff
@@ -31,12 +21,60 @@ module.exports = {
         test: /\.(ts|js)x?$/,
         // exclude: /node_modules/,
         loader: "babel-loader",
+        options: { presets: ["@babel/env", "@babel/preset-react"] },
```
phulin, line 24: I don't think we want preset-react for the non-react code.

src/relay/App.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from "react";
+import Setting from "./components/Setting";
+
+class App extends React.Component<Data> {
+  constructor(props: Data) {
+    super(props);
+  }
+  render(): React.ReactNode {
+    const elements = this.props.settings.map((setting) => {
+      return (
```
phulin, line 10: You can drop the braces and return here.

src/relay_garbo.ts
burningbman: I'm not sure. I just copied it from the existing ash relay script. I'm fine removing it. It currently displays 0.

webpack.config.js
burningbman: I'm not familiar with webpack-merge, or webpack for that matter. I spent a little trying to get it to work, but wasn't able to wrap my ahead around what it should look like. If you could show me an example of someone using webpack-merge to handle multiple outputs, I'd be happy to use it.

webpack.config.js
burningbman: Moved to just react config.

src/relay/App.tsx
burningbman: done

src/relay/components/Setting.tsx
burningbman: Done. As the components get more complex, do you switch to class components?

webpack.config.js
burningbman: Turns out I had broken the webpack config doing something earlier. Using webpack-merge now.

src/relay/components/Setting.tsx
phulin: Nope. Those are really the old way of writing components - the new way is to use hooks.

src/relay/index.tsx
```diff
@@ -0,0 +1,22 @@
+declare global {
+  function getData(callback): void;
+  type SettingProp = {
```
phulin, line 3: I'd move these types into App.tsx.

src/relay_garbo.ts
phulin: Yeah, don't think we need it - would be nice to display if it worked, but it doesn't.

src/relay_garbo.ts
```diff
@@ -0,0 +1,54 @@
...
+      writeln(`Changing setting ${field} to ${fields[field]} <br>`);
+      set(field, fields[field]);
+    }
+  });
+
+  // load user prefences into json object to pass to react
+  const settings = [];
+  const lines = fileToArray("garbo_settings.txt");
+  for (const i in lines) {
+    const data = lines[i].split("\t");
```
phulin, line 30:
```suggestion
  for (const line of lines) {
    const [, , name, type, description] = line.split("\t");
```
You want to use for... of to iterate over iterables.

Can you also change the settings file not to have the first two fields? Don't we also want a default value field in that file?

webpack.config.js
```diff
@@ -32,11 +23,79 @@ module.exports = {
         // exclude: /node_modules/,
         loader: "babel-loader",
       },
+      {
+        test: /\.css$/,
+        use: [MiniCssExtractPlugin.loader, "css-loader"],
+      },
```
phulin, line 29: I don't think we need a css loader right now. I think you can strike this and all the MiniCssExtractPlugin stuff.

src/relay/App.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from "react";
+import Setting from "./components/Setting";
+
+function App(props: Data) {
```
phulin, line 4: Typical react style would be to bind the keys in the function signature:
```suggestion
function App({ settings, rev }: Data) {
```

src/relay_garbo.ts
```diff
@@ -0,0 +1,54 @@
+import { fileToArray, formField, formFields, svnInfo, writeln } from "kolmafia";
+import { get, set } from "libram";
+
+export function main(): void {
+  // handle updating values
+  const fields = formFields();
+  Object.keys(fields).forEach((field) => {
```
phulin, line 7:
```suggestion
  for (const field of Object.keys(fields)) {
```
From a high-level perspective, I'd like to include any code that's generating user-visible stuff to be inside React world and not generated server-side. But this is fine for now.

src/relay_garbo.ts
burningbman: I can change the settings file, but fileToArray returns an Object, not an iterable.

webpack.config.js
burningbman: I tried removing all that (and tried it before adding it in) and it wasn't properly including the css. If you can show me how to do it, I'm down.

phulin: OK. I'll take a look tonight.

phulin:
I'll figure out the CSS stuff later. Can you rebase and then we can merge?

PR #280, opened 2021-07-20
changing robort drinks from buy() to retrieveItem()

lewismd13:
Hopefully now we won't buy 123,456 meat bloody noras


horrible-little-slime (approved):
Good short-term solution

PR #281, opened 2021-07-20
use freeruns to prepare noon in gingerbread city

### LINKED ISSUE #204: [resource] use free runs profitably
currently, i don't think that garbo utilizes free runs for bander or boots. i think it would be a relatively easy thing to -place-, but it is not super straightforward what the runs should be used for. 

- first, it should be used to get latte ingredients (i believe this is already coded)
- the best usage is likely gingerbread city candy/cigarettes, for guaranteed profit
- if the user does -not- have gingerbread city, in non-CS aftercore, could be used for UR hunting (thinking especially of knob menagerie or icy peak, but peak requires com% management)
- if the user is in CS aftercore, could be used to UR hunt but -definitely- requires com% management

happy for ideas on this but i think free run usage pairs well with the current famweight push right before you profcopy embezzlers. i think it's a worthy usage but might be annoying to do, as the zones you want to free run in may not be open...

horrible-little-slime:
Also revamps free run code!

Does not attempt to actually do the noon adventure, but should be immune to turtle taming.

Partially resolves #204


phulin:
Needs rebase

Tokoeka (approved):

PR #284, opened 2021-07-22
Add eslint plugin for names.

phulin:


horrible-little-slime (approved):
I could kiss you on the mouth

PR #285, opened 2021-07-22
Add feel nostalgic support

CheeseyPickle:
Probably a really minor optimization, but it's not like it gets used otherwise

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -88,7 +88,7 @@ export class Macro extends LibramMacro {
         Macro.if_("monstername garbage tourist", Macro.trySkill("Gallapagosian Mating Call"))
       )
       .externalIf(
-        get("lastCopyableMonster") == $monster`garbage tourist`,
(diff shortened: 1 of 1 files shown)


src/combat.ts
```diff
@@ -87,6 +87,10 @@ export class Macro extends LibramMacro {
         get("_gallapagosMonster") !== $monster`garbage tourist`,
         Macro.if_("monstername garbage tourist", Macro.trySkill("Gallapagosian Mating Call"))
       )
+      .externalIf(
+        get("lastCopyableMonster") == $monster`garbage tourist`,
```
horrible-little-slime, line 91: Linting issue: make this a ===

CheeseyPickle: Fixed, not used to ts, sorry :(

horrible-little-slime (approved):
Not the strictly most optimal thing we could be doing, but strictly more optimal than what we have now

PR #294, opened 2021-07-25
replace getWorkshed with getCampground as it appears in horseradish()

horrible-little-slime:
Right now, mafia will have a VERY bad time if you use getWorkshed and have no workshed item. I'd like to return to getWorkshed in the long run, but until it's fixed, I'd like garbo to work.


horrible-little-slime:
r20803

PR #298, opened 2021-07-27
prevent GWBTs if headed to the icy peak

horrible-little-slime:


horrible-little-slime:
Trying a different approach at a later date, per discord discussion with Phill

PR #299, opened 2021-07-27
use fewer, larger withChoice and withProperty calls

horrible-little-slime:


horrible-little-slime:
Closed per discord discussion; will use upcoming PropertyManager feature from Libram

PR #301, opened 2021-07-27
restructure barf turn. BONUS: fight guaranteed kramcos in wanderer zones!

horrible-little-slime:


horrible-little-slime:
Tested 2x and no apparent explosions, although I am yet to examine logs to verify that it lacked non-explosive issues

phulin:
Hmmm, don't we want to do kramcos significantly more often than this? Like whenever 1 / E[turns to next kramco] * (~2000 + MPA) is better than the MPA from your offhand?

horrible-little-slime:
The primary advantage here is that it means you're decked out for free
fights during a kramco, instead of a sausage turn being worth 1k + LGR +
bjorn* + pantogram, we end up getting more out of it when it's used
deliberately. That being said, yeah we may or may not actually want to do
this, I just remember Phill was talking about maybe switching.

One thing to remember about kramcos is because encountering them makes
future ones less likely, the naive heuristic is probably overestimating
their value. How many guaranteed kramcos do we expect during a barf day?
How many kramcos do we expect when you use our current algorithm?

src/lib.ts
```diff
@@ -969,9 +969,5 @@ export function pickBjorn(mode: PickBjornMode = PickBjornMode.FREE): BjornedFami
 }
 
 export function kramcoGuaranteed(): boolean {
-  return (
-    have($item`Kramco Sausage-o-Matic™`) &&
-    totalTurnsPlayed() - get("_lastSausageMonsterTurn") + 1 >=
-      5 + 3 * get("_sausageFights") + Math.pow(Math.max(0, get("_sausageFights") - 5), 3)
-  );
+  return have($item`Kramco Sausage-o-Matic™`) && getKramcoWandererChance() === 1;
```
phulin, line 972:
```suggestion
  return have($item`Kramco Sausage-o-Matic™`) && getKramcoWandererChance() >= 1;
```

just in case - floating point errors etc

PR #304, opened 2021-07-27
calculate and cache lgr value

horrible-little-slime:
I do not currently account for volcoino, and doing so would dink with our caching a little bit. Not that much, though. But until someone gives me numbers on how to treat it, no <3


phulin:
merge conflicts

horrible-little-slime:
Merge issues resolv'd

horrible-little-slime:
'twas but a lark

PR #312, opened 2021-08-02
Break bjorn code out, take 2

phulin:


horrible-little-slime:
Needs rebasing, but if this is tested I'll push 'er through the second that's done

phulin:
done

PR #313, opened 2021-08-02
streamline daily tasks

horrible-little-slime:


horrible-little-slime:
obsoleted it myself, mostly

PR #316, opened 2021-08-03
try to freerun to fish for hipster/goth fights

horrible-little-slime:


src/dailies.ts
```diff
@@ -541,3 +542,27 @@ export function gingerbreadPrepNoon(): void {
...
 }
+
+export function hipsterFishing(): void {
+  if (have($familiar`Mini-Hipster`)) {
+    useFamiliar($familiar`Mini-Hipster`);
+  } else if (have($familiar`Artistic Goth Kid`)) {
+    useFamiliar($familiar`Artistic Goth Kid`);
+  } else return;
+
+  while (findRun(false) && get("_gothKidFights") < 7) {
```
Rinn, line 553:
```suggestion
  while (findRun(false) && get("_hipsterAdv") < 7) {
```

`_gothKidFights` is to track the pvp fights acquired, you want `_hipsterAdv`

src/dailies.ts
```diff
@@ -557,7 +564,12 @@ export function hipsterFishing(): void {
     const runSource = findRun(false);
     if (!runSource) return;
     if (runSource.prepare) runSource.prepare();
-    freeFightOutfit([...(runSource.requirement ? [runSource.requirement] : [])]);
+    freeFightOutfit([
+      ...(runSource.requirement ? [runSource.requirement] : []),
+      ...(have($familiar`Mini-Hipster`)
+        ? [new Requirement([], { forceEquip: [bestHipsterEquip] })]
```
Rinn, line 570: passing all 3 with a bonus may be better in case any of the equipment isn't available it will just handle it

horrible-little-slime: You're right, that's way better, implemented.

horrible-little-slime:
Tested and functioning as of 8/4

Phillammon (approved):

PR #319, opened 2021-08-04
yoink a navel ring or GAP during the freerun-using part of dailySetup()

horrible-little-slime:


horrible-little-slime:
tested and functioning as of 8/4; does anyone think we should yoink both of them, just in case, or should prioritizing the navel ring stick around?

Phillammon (approved):

PR #320, opened 2021-08-04
get kgb martinis

horrible-little-slime:


src/dailies.ts
```diff
@@ -541,3 +541,13 @@ export function gingerbreadPrepNoon(): void {
...
   }
 }
+
+export function martini(): void {
+  if (
+    !have($item`Kremlin's Greatest Briefcase`) ||
+    get("_kgbClicksUsed") > 17 ||
+    get("_kgbDispenserUses") >= 3
+  )
+    return;
```
Tokoeka, line 551: is it worth putting {} around the return, just to avoid confusion? I had to parse it a couple of times before the segmentation stuck.

Tokoeka (approved):

horrible-little-slime:
Yeah I'll do that now; it made more sense when the condition was 1 line

PR #321, opened 2021-08-04
Implement latte olfacts, and refill latte for them

CheeseyPickle:


horrible-little-slime:
As written, this will refill with the same ingredient twice if you haven't unlocked two from the set ["cajun", "carrot", "rawhide"]--note how in the existing latte refilling code, we only refill at all if you're able to get cajun & rawhide, because only in that scenario does it beat out half a purse.

CheeseyPickle:
Oh, so that's why we only refill if we have those 2, got it

Phillammon (approved):

PR #322, opened 2021-08-04
only harvest garden if we're going to swap it later

horrible-little-slime:


Phillammon (approved):
We probably want to add a preference for "should we be allowed to mess w/ your garden"

PR #325, opened 2021-08-05
Update babel to 7.15 and thus remove patch files

gausie:
Also update yarn to latest to make this work smoothly.

This builds properly but its having issues with `libram` imports I think? Not quite sure what's going on yet.


horrible-little-slime:
Is the libram imports thing resolved? If so I'll push this through once I'm
out of bed.

PR #327, opened 2021-08-06
ice sculpture as a copy source

Rinn:


src/fights.ts
```diff
@@ -393,6 +409,10 @@ function embezzlerSetup() {
     retrieveItem($item`4-d camera`);
   }
 
+  if (!get("_iceSculptureUsed") && !have($item`ice sculpture`)) {
+    retrieveItem($item`unfinished ice sculpture`);
+  }
+
```
phulin, line 415: Ice sculpture often isn't worth it - can we add a price check somehow?

src/fights.ts
```diff
@@ -404,12 +408,24 @@ function embezzlerSetup() {
...
+  ) {
     retrieveItem($item`4-d camera`);
   }
 
-  if (!get("_iceSculptureUsed") && !have($item`ice sculpture`)) {
+  if (
+    !get("_iceSculptureUsed") &&
+    !have($item`ice sculpture`) &&
+    averageEmbezzlerNet - averageTouristNet >
+      mallPrice($item`snow berries`) + mallPrice($item`ice harvest`) * 3
```
horrible-little-slime, line 426: Is it worth also checking the direct price of a sculpture?

Rinn: it's not tradable

horrible-little-slime (approved):

PR #329, opened 2021-08-06
Upgrade libram and use PropertyManager

horrible-little-slime:
still in testing, but I expect it to work

### DIFF
--- a/src/dailies.ts
+++ b/src/dailies.ts
@@ -48,14 +48,16 @@ import {
   ensureEffect,
   findRun,
   prepWandererZone,
+  PropertyManager,
   questStep,
   Requirement,
   saleValue,
+  setChoice,
   tryFeast,
 } from "./lib";
 import { freeFightOutfit } from "./outfit";
 import { withStash } from "./clan";
-import { set, withChoice, withChoices } from "libram/dist/property";
+import { set, withChoice } from "libram/dist/property";
 import { estimatedTurns } from "./globalvars";
 import { Macro } from "./combat";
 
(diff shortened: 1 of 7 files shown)


src/dailies.ts
```diff
@@ -48,14 +48,16 @@ import {
...
   questStep,
   Requirement,
   saleValue,
+  setChoice,
   tryFeast,
 } from "./lib";
 import { freeFightOutfit } from "./outfit";
 import { withStash } from "./clan";
-import { set, withChoice, withChoices } from "libram/dist/property";
+import { set, withChoice } from "libram/dist/property";
```
phulin, line 60: move this into the bigger libram export

src/dailies.ts
```diff
@@ -144,68 +146,64 @@ export function latte(): void {
...
-            ]);
-            adventureMacro($location`The Black Forest`, runSource.macro);
-            horseradish();
-          }
-        });
+      PropertyManager.setChoices({
+        [923]: 1,
+        [924]: 1,
+        [502]: 2,
+        [505]: 2,
```
phulin, line 153: can you add comments saying what these are?

src/lib.ts
```diff
@@ -32,9 +32,11 @@ import {
   Macro,
   MaximizeOptions,
   property,
-  set,
   SongBoom,
 } from "libram";
+import { PropertiesManager } from "libram/dist/property";
```
phulin, line 37: this should also be exported now

phulin (changes requested):

src/fights.ts
```diff
@@ -99,7 +99,7 @@ import {
 } from "./outfit";
 import { bathroomFinance } from "./potions";
 import { estimatedTurns, log } from "./globalvars";
-import { getString, withChoice, withChoices } from "libram/dist/property";
+import { getString } from "libram/dist/property";
```
phulin, line 102: can't we just use `get<string>`?

src/lib.ts
```diff
@@ -31,19 +31,21 @@ import {
   have,
   Macro,
   MaximizeOptions,
+  PropertiesManager,
   property,
-  set,
   SongBoom,
 } from "libram";
 
+export const PropertyManager = new PropertiesManager();
```
phulin, line 39: variables should be lowercase

horrible-little-slime:
Both choice adventures and regular properties appear to be set, stored, and reset properly as of my testing

PR #331, opened 2021-08-06
Set Snow Suit to nose for drops & +item

### LINKED ISSUE #78: Snow Suit decorations
it uses the mp regen config when it should have the carrot nose config instead.

Rinn:
resolves #78


horrible-little-slime (approved):

Rinn:
should probably give it a bonus if `_carrotNoseDrops` is less than 3

Rinn:
Ah hm, the maximizer is actually switching it to hat for some reason.

PR #332, opened 2021-08-06
drinkSafe instead of drink

Rinn:
Just replacing two calls to mafia drink with drinkSafe

### DIFF
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -190,12 +190,16 @@ function fillLiver() {
   if (myFamiliar() === $familiar`Stooper`) {
     useFamiliar($familiar`none`);
   }
+  try {
+    while (myInebriety() + 1 <= inebrietyLimit() && availableAmount($item`astral pilsner`) > 0) {
+      drinkSafe(1, $item`astral pilsner`);
+    }
+  } catch {
+    print(`Failed to drink leftover astral pilsners.`, "red");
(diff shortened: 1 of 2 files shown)


horrible-little-slime (approved):

horrible-little-slime (changes requested):
I changed my mind, and no longer approve of this; drinkSafe calls acquire, which then checks for a price cap. You need to either edit acquire to not blow up for pilsners, or continue not using drinkSafe for the times we drink pilsners

Rinn:
oh i see what you mean, I'll fix that

src/diet.ts
```diff
@@ -186,16 +179,41 @@ function fillStomach() {
...
+  try {
+    if (
+      !get("_mimeArmyShotglassUsed") &&
+      itemAmount($item`mime army shotglass`) > 0 &&
+      availableAmount($item`astral pilsner`) > 0
+    ) {
+      drinkSafe(1, $item`astral pilsner`);
+    }
+    if (
+      globalOptions.ascending &&
```
horrible-little-slime, line 195: :)

horrible-little-slime (approved):
I'd like to see `fillLiverAstralPilsner();` used where we start filling up with turkey, regardless of whether we're ascending. That may generally be the best place for it.

PR #334, opened 2021-08-06
Combat improvements

Rinn:
- use simpler macro for all non-meat fights
- kill() Macro that nukes if regular attacks miss, hp falls below 25%, or past round 20
- ghost bustin, ensure enough stuns can be used to not get hit
- seal clubbin, use Furious Wallop and smacks
- use $skill & $item instead of strings where appropriate
- inject a macro that can deal with tentacles if you have eldritch attunement
- use Extract, Pocket Crumbs, & porquoise-handled sixgun in meatKill() by checking monster hp and passive damage from vykea + other passive damage sources (monster manuel only)
- use abstraction on machine tunnel monsters to get item if profitable
- tryHaveSkill doesn't need to accept a string, be explicit. Use for all skills that aren't dynamic
- add tryHaveItem to simplify macro more if reusable combat items are unavailable before the fight start
- kill with a sauce spell if a sauceror for weaksauce mp when not using crits for meat
- add cocoabo, boa, & npzr stasis support

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -202,7 +202,11 @@ export class Macro extends LibramMacro {
           Macro.while_("!pastround 3 && !hppercentbelow 25", Macro.item("seal tooth"))
         )
         .externalIf(
-          myFamiliar() === $familiar`Stocking Mimic`,
+          [
+            $familiar`Cocoabo`,
+            $familiar`Ninja Pirate Zombie Robot`,
+            $familiar`Stocking Mimic`,
+          ].some((familiar) => myFamiliar() === familiar),
           Macro.while_("!pastround 10 && !hppercentbelow 25", Macro.item("seal tooth"))
         )
         .externalIf(
@@ -231,7 +235,11 @@ export class Macro extends LibramMacro {
           Macro.while_(`!pastround 3 && monsterhpabove ${passiveDamage}`, Macro.item("seal tooth"))
         )
         .externalIf(
-          myFamiliar() === $familiar`Stocking Mimic`,
+          [
+            $familiar`Cocoabo`,
+            $familiar`Ninja Pirate Zombie Robot`,
+            $familiar`Stocking Mimic`,
+          ].some((familiar) => myFamiliar() === familiar),
           Macro.while_(`!pastround 10 && monsterhpabove ${passiveDamage}`, Macro.item("seal tooth"))
         )
         .externalIf(
@@ -267,7 +275,9 @@ export class Macro extends LibramMacro {
         Macro.while_("!pastround 3 && !hppercentbelow 25", Macro.item("seal tooth"))
       )
       .externalIf(
-        myFamiliar() === $familiar`Stocking Mimic`,
+        [$familiar`Cocoabo`, $familiar`Ninja Pirate Zombie Robot`, $familiar`Stocking Mimic`].some(
+          (familiar) => myFamiliar() === familiar
+        ),
         Macro.while_("!pastround 10 && !hppercentbelow 25", Macro.item("seal tooth"))
       )
       .externalIf(
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -145,20 +145,25 @@ export class Macro extends LibramMacro {
     )
       .tryHaveSkill($skill`Sing Along`)
       .externalIf(
-        !have($effect`On the Trail`),
+        !have($effect`On the Trail`) && have($skill`Transcendent Olfaction`),
         Macro.if_("monstername garbage tourist", Macro.trySkill("Transcendent Olfaction"))
       )
       .externalIf(
-        get("_gallapagosMonster") !== $monster`garbage tourist`,
+        get("_gallapagosMonster") !== $monster`garbage tourist` &&
+          have($skill`Gallapagosian Mating Call`),
         Macro.if_("monstername garbage tourist", Macro.trySkill("Gallapagosian Mating Call"))
       )
       .externalIf(
-        get("_latteMonster") !== $monster`garbage tourist` ||
-          getCounters("Latte Monster", 0, 30).trim() === "",
+        !get("_latteCopyUsed") &&
+          (get("_latteMonster") !== $monster`garbage tourist` ||
+            getCounters("Latte Monster", 0, 30).trim() === "") &&
+          have($item`latte lovers member's mug`),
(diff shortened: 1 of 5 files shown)


src/combat.ts
```diff
@@ -130,52 +145,262 @@ export class Macro extends LibramMacro {
...
+      .externalIf(have($skill`Meteor Lore`), Macro.trySkill($skill`Micrometeorite`))
+      .tryHaveItem($item`Time-Spinner`)
+      .tryHaveItem($item`Rain-Doh indigo cup`)
+      .tryHaveItem($item`Rain-Doh blue balls`)
+      .externalIf(
+        haveEquipped($item`Buddy Bjorn`) || haveEquipped($item`Crown of Thrones`),
+        Macro.while_("!pastround 3 && !hppercentbelow 25", Macro.item("seal tooth"))
+      )
+      .externalIf(
+        myFamiliar() === $familiar`Stocking Mimic`,
```
horrible-little-slime, line 270: Could we add NPZR to this?

src/combat.ts
```diff
@@ -130,52 +145,262 @@ export class Macro extends LibramMacro {
...
+          !(myClass() === $class`Sauceror` && have($skill`Curse of Weaksauce`)),
+          Macro.while_("!pastround 20 && !hppercentbelow 25 && !missed 1", Macro.attack())
+        )
+        // Using while_ here in case you run out of mp
+        .while_("hasskill Saucegeyser", Macro.skill($skill`Saucegeyser`))
+        .while_("hasskill Weapon of the Pastalord", Macro.skill($skill`Weapon of the Pastalord`))
+        .while_("hasskill Cannelloni Cannon", Macro.skill($skill`Cannelloni Cannon`))
+        .while_("hasskill Wave of Sauce", Macro.skill($skill`Wave of Sauce`))
+        .while_("hasskill Saucestorm", Macro.skill($skill`Saucestorm`))
+        .while_("hasskill Lunging Thrust-Smack", Macro.skill($skill`Lunging Thrust-Smack`))
```
horrible-little-slime, line 302: Consider prioritizing LTS over saucegeyser if you're a muscle class

src/combat.ts
```diff
@@ -130,52 +145,262 @@ export class Macro extends LibramMacro {
...
+
+  basicCombat(): Macro {
+    return this.startCombat().kill();
+  }
+
+  static basicCombat(): Macro {
+    return new Macro().basicCombat();
+  }
+
+  ghostBustin(): Macro {
```
horrible-little-slime, line 320: we use love gnats in the actual macro--could we use it to increment stunRounds as well?

src/combat.ts
```diff
@@ -130,52 +145,262 @@ export class Macro extends LibramMacro {
...
+
+    // Lacking multi-round stuns
+    if (stunRounds < 3) {
+      return this.basicCombat();
+    }
+
+    return this.tryHaveSkill($skill`Sing Along`)
+      .tryHaveItem($item`Rain-Doh blue balls`)
+      .externalIf(get("lovebugsUnlocked"), Macro.trySkill($skill`Summon Love Gnats`))
+      .tryHaveSkill(classStun)
```
horrible-little-slime, line 392: Does this  behave nicely when classStun is null?

src/lib.ts
```diff
@@ -466,3 +462,61 @@ export function saleValue(...items: Item[]): number {
...
+      // 25 poison damage (25+12+6+3+1)
+      return 47;
+    case $familiar`Robortender`:
+      return 20;
+  }
+
+  return 0;
+}
+
+export function maxPassiveDamage(): number {
```
horrible-little-slime, line 508: We don't need to consider regular familiar damage because those will only happen in inherently free fights, right?

horrible-little-slime (commented):
Will test tomorrow and potentially give more thorough feedback

src/combat.ts
Rinn: yes, I changed tryHaveSkill to accept null

src/combat.ts
Rinn: Yeah I did: `if (get("lovebugsUnlocked")) stunRounds++;`

src/combat.ts
Rinn: yeah, how many rounds should it stasis for

src/combat.ts
Rinn: i was thinking spells are safer because they always hit, you could be a muscle class and still fail to hit if your moxie is low against a scaling monster. Or if something is physically resistant

src/lib.ts
Rinn: yeah. NPZR damage would need to be considered if stasising, but otherwise we're not using any attack familiars for meat farming

src/combat.ts
Rinn: and really the chance of even getting to the nuke part of this are pretty so I didn't want to over complicate it

horrible-little-slime: Reasonable!

horrible-little-slime (approved):

src/combat.ts
```diff
@@ -130,52 +145,277 @@ export class Macro extends LibramMacro {
...
+    return new Macro().basicCombat();
+  }
+
+  ghostBustin(): Macro {
+    // Only bust ghosts if you have enough stunners to prevent getting hit
+    let stunRounds = 0;
+    let classStun: Skill | null = null;
+    let extraStun: Skill | null = null;
+    if (have($item`Rain-Doh blue balls`)) stunRounds++;
+    if (get("lovebugsUnlocked")) stunRounds++;
```
CheeseyPickle, line 341: Isn't Frost Bite a stun as well? Or is that too MP expensive?

CheeseyPickle: Nvm, I forgot we want to shoot the ghosts :/

CheeseyPickle: Although, should we consider blood bubble/pill keeper?

Rinn: You can't ensure blood bubble hasn't already triggered because it could have blocked an attack before you start shooting because of losing init or from getting hit during the Sing Along round,. As for pill keeper you're probably not going to have any blocking equipment on because the free combat farming equipment is going to be used.

Changing it to use specific equipment is an option but you'd miss out on possible free random item drops from the ring etc, and if equipment was factored in it would probably just be easier to maximize elemental res and DR instead of dealing with blockers.

PR #335, opened 2021-08-06
Don't force memory chip in prof meatchain

CheeseyPickle:
We're forcing the memory chip to be equipped during the prof meat chain... for some reason.
Shouldn't it be better to allow amulet coin to be equipped or something?

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -507,8 +507,32 @@ export function dailyFights(): void {
       }
 
       startDigitize();
-
+      
       // SECOND EMBEZZLER CHAIN
+      if (have($familiar`Pocket Professor`) && !get<boolean>("_garbo_chipChain", false)) {
+        const startLectures = get("_pocketProfessorLectures");
+        const fightSource = getEmbezzlerFight();
+        if (!fightSource) return;
+        useFamiliar($familiar`Pocket Professor`);
+        meatOutfit(true, [
+          ...fightSource.requirements,
+          new Requirement([], { forceEquip: $items`Pocket Professor memory chip` }),
+        ]);
+        if (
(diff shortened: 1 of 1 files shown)


Rinn:
it should probably do as many as possible without the memory chip, then equip the chip and do the 2 more with the meat outfit, then continue on to the familiar weight chain with it equipped and do the remaining

CheeseyPickle:
Yeah, that would be better actually.

horrible-little-slime:
> 
> 
> it should probably do as many as possible without the memory chip, then equip the chip and do the 2 more with the meat outfit, then continue on to the familiar weight chain with it equipped and do the remaining

I disagree; I think having three separate chains is a bad idea insofar as it means getting the same net embezzler total, but fewer run with your lep-analogue

Rinn:
> > it should probably do as many as possible without the memory chip, then equip the chip and do the 2 more with the meat outfit, then continue on to the familiar weight chain with it equipped and do the remaining
> 
> I disagree; I think having three separate chains is a bad idea insofar as it means getting the same net embezzler total, but fewer run with your lep-analogue

I don't understand, this is for the lecture chains where pocket professor is required so you can't use a different familiar. My point for that suggestion is so you can equip the meat outfit + amulet coin until you hit the lecture limit, then switch to meat outfit + memory chip until you hit the cap again, then finally switch to the familiar weight outfit + memory chip to finish. This is functionally the same except you have more turns with the amulet coin.

CheeseyPickle:
Hmm, in that case, doing this at all probably isn't worth it because you lose 2 fights with your meat outfit in exchange for having amulet coin for like 8 fights.

PR #338, opened 2021-08-07
Mayflower Bouquet drops + sporadic +meat bonus

Rinn:
Built off of #331, so that should be merged first

Flower drops are weird, I got the first 4 relatively quick than the 5th took a lot more turns to get (wiki speculates the drop rate falls off), so I added in a pretty aggressive degeneration of the drop rate and floor it at 1% after 5 drops so garbo won't get stuck with using it over something better for an extended amount of time.

### DIFF
--- a/src/outfit.ts
+++ b/src/outfit.ts
@@ -229,6 +229,27 @@ function cheeses(embezzlerUp: boolean) {
       )
     : [];
 }
+function mayflowerBouquet() {
+  // +40% meat drop 12.5% of the time (effectively 5%)
+  // Drops flowers 50% of the time, wiki says 5-10 a day.
+  // Theorized that flower drop rate drops off but no info on wiki.
+  // During testing I got 4 drops then the 5th took like 40 more adventures
+  // so let's just assume rate drops by 11% with a min of 1% ¯\_(ツ)_/¯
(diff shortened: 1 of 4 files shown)


src/outfit.ts
```diff
@@ -237,3 +229,34 @@ function cheeses(embezzlerUp: boolean) {
...
+    (saleValue($item`tin magnolia`) +
+      saleValue($item`upsy daisy`) +
+      saleValue($item`lesser grodulated violet`) +
+      saleValue($item`half-orchid`) +
+      saleValue($item`begpwnia`) / 5) *
+    Math.max(0.01, 0.5 - get("_mayflowerDrops") * 0.11);
+  return new Map<Item, number>([
+    [
+      $item`Mayflower bouquet`,
+      (get("_mayflowerDrops") < 5 ? averageFlowerValue : 0) + sporadicMeatBonus,
```
horrible-little-slime: this should care about what type of fight we're in. for free fights, sporadicMeatBonus is probably 0; for barfTurns, it's higher, and for embezzlers...we should probably force it to not use it for embezzlers

Rinn: yeah you're right, it need to take in the same args as the bjorn logic

horrible-little-slime (approved):

PR #339, opened 2021-08-08
Grab temple adventures

CheeseyPickle:
Grab Hidden Temple adventures if it's unlocked and you're also ascending.

Probably not the most optimal place to put it because it also extends up to 10 effects, but I have no idea how to make it optimal, and right now the adventures are more important.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -71,6 +71,7 @@ import {
   TunnelOfLove,
   Witchess,
 } from "libram";
+import { acquire } from "./acquire";
 import { fillAsdonMartinTo } from "./asdon";
 import { withStash } from "./clan";
 import { Macro, withMacro } from "./combat";
@@ -388,7 +389,8 @@ function embezzlerSetup() {
   if (
     globalOptions.ascending &&
     questStep("questM16Temple") > 0 &&
-    get("lastTempleAdventures") < myAscensions()
+    get("lastTempleAdventures") < myAscensions() &&
+    acquire(1, $item`Stone Wool`, 3 * get("valueOfAdventure") + 100, false) > 0
   ) {
     ensureEffect($effect`Stone-Faced`);
     setChoice(582, 1);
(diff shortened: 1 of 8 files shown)


horrible-little-slime:
Consider adding a price check for stone wool--equilibrium price is going to be slightly higher than 3 * valueOfAdventure

src/fights.ts
```diff
@@ -385,6 +385,12 @@ function embezzlerSetup() {
     }
   });
   if (have($item`License to Chill`) && !get("_licenseToChillUsed")) use($item`License to Chill`);
+  if (globalOptions.ascending && get("questM16Temple") === "finished" && get("lastTempleAdventures") < get("knownAscensions")) {
```
horrible-little-slime, line 388: consider using questStep in lieu of === "finished", and myAscensions() in lieu of get("knownAscensions")

horrible-little-slime (approved):
Very good, although I'm not entirely sure we want to run from baa'baa'bur'ann in the unlikely event we hit her. I think that's up in the air, though, and this is strictly speaking better than what we're doing right now, so I'm sold.

src/fights.ts
```diff
@@ -379,6 +380,24 @@ function embezzlerSetup() {
...
+    acquire(1, $item`stone wool`, 3 * get("valueOfAdventure") + 100, false) > 0
+  ) {
+    ensureEffect($effect`Stone-Faced`);
+    setChoice(582, 1);
+    setChoice(579, 3);
+    const runSource = findRun();
+    if (runSource) {
+      if (runSource.prepare) runSource.prepare();
+      freeFightOutfit([...(runSource.requirement ? [runSource.requirement] : [])]);
+      while (get("lastTempleAdventures") < myAscensions()) {
```
Rinn, line 396: this isn't going to work, you need to get a new runSource every iteration, prepare it, and equip the requirements

Rinn (changes requested):

src/fights.ts
horrible-little-slime: as an addon to this, I think this is one of the situations like with startDigitize() where you may want to add a backup LTB if you can't find a run

PR #342, opened 2021-08-08
Add Heavy Rains support

soolar:
Rain Man embezzlers and lightning strike free kills so far.

TODO:
1) Consider equipping Thor's Pliers for more lightning strikes
2) Probably stop using lightning strikes on worms, saving them for Heavy Rains wanderers instead
3) Probably shift rain man to be later if you don't have manuel, so that you've definitely fought at least one embezzler that life time before using rain man.


soolar:
I'm just going to close this PR for the moment, I might come back to it eventually but I'm ascending again so it'll probably be a while until I'm back to long enough term garbo farming to justify going HR, and even then I'm still not 100% sure the washaway risk and some other problems are worth the benefits of HR.

PR #350, opened 2021-08-09
work better when started with some used spleen

horrible-little-slime:


JamesDowney (approved):
Looks like an improvement to me!

PR #356, opened 2021-08-11
Add command line option to not run barf turns

burningbman:
Allow user to run `garbo nobarf` to only run free fights and initial embezzler fights.
Diet is still performed before running the fights.
Note that some embezzler logic had to be taken out of fights.js in order to prevent a circular dependency between fights.js and globalvars.js. The refactor is a straight copy-paste.

### DIFF
--- a/src/dailies.ts
+++ b/src/dailies.ts
@@ -48,6 +48,7 @@ import { meatFamiliar } from "./familiar";
 import {
   baseMeat,
   ensureEffect,
+  estimatedTurns,
   findRun,
   prepWandererZone,
   propertyManager,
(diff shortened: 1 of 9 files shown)


horrible-little-slime:
I suspect this will still have circular imports; `lib` imports `globalOptions` from `globalVars`, which in turn references `embezzlerCount` from `embezzlers`, which in turn imports `Requirement` and `sum` from `lib`. Have you gotten this to build and run?

burningbman:
I had gotten it to build and run, but that's a good point. I can pull the free runs into their own file like embezzlers?

horrible-little-slime:
I don't see how that would resolve the situation. Arguably the best solution would be to add `sum` and `Requirement` to Libram, which may very well be worth doing.

phulin:
I haven't looked at this in detail, but globalVars probably shouldn't
import anything.

horrible-little-slime:
One solution could be to locally define a function in a variety of places tha tlooks like `globalOptions.noBarf ? embezzlerCount() : estimatedTurns()`

burningbman:
Gonna close this one since the obvious solution is to move estimatedTurns to lib.ts.

PR #360, opened 2021-08-11
Use shocking lick as a free kill

CheeseyPickle:
Use any lick charges you have remaining if you're ascending.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -1193,7 +1193,7 @@ const freeKillSources = [
 
   new FreeFight(
     () =>
-      get("shockingLickCharges") > 0 && globalOptions.ascending ? get("shockingLickCharges") : 0,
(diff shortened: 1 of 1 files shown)


horrible-little-slime (approved):

src/fights.ts
```diff
@@ -1190,6 +1190,18 @@ const freeKillSources = [
       requirements: () => [new Requirement(["100 Item Drop"], {})],
     }
   ),
+
+  new FreeFight(
+    () =>
+      get("shockingLickCharges") > 0 && globalOptions.ascending ? get("shockingLickCharges") : 0,
```
horrible-little-slime, line 1196:
```suggestion
      globalOptions.ascending ? get("shockingLickCharges") : 0,
```

CheeseyPickle: Yeah that probably would be better :/

PR #364, opened 2021-08-12
dont' require pointer finger ring; rehash pointer ring logic in general

horrible-little-slime:


src/outfit.ts
```diff
@@ -106,27 +107,40 @@ export function meatOutfit(
...
-    }
     // TODO: Fix pointer finger ring construction.
-    if (myClass() !== $class`Seal Clubber`) {
-      if (have($item`haiku katana`)) {
-        forceEquip.push($item`haiku katana`);
-      } else if (have($item`unwrapped knock-off retro superhero cape`)) {
-        if (!have($item`ice nine`)) retrieveItem($item`ice nine`);
-        forceEquip.push($item`ice nine`);
+    if (have($item`mafia pointer finger ring`)) {
+      forceEquip.push($item`mafia pointer finger ring`);
```
Rinn, line 112: I'd move the crit equipment logic into a separate function and return a Requirement, then you can merge it if doesn't conflict with the kramco or protonic checks below, that will prevent accidently equipping one item if several are required when the conflict happens.

Rinn (changes requested):

src/index.ts
```diff
@@ -245,6 +234,12 @@ function barfTurn() {
 
     // d. get dressed
     meatOutfit(embezzlerUp, [], underwater);
+    if (
```
Rinn, line 237: You prob don't need this here since the retrocape is being configured in meatOutfit

Rinn (changes requested):

horrible-little-slime:
I feel like crit equipment takes priority over kramco and proton, but
separating that out into a function seems wise regardless.

Maybe we do proton over the crit equips--is a free fight worth over 550
meat?

src/index.ts
```diff
@@ -351,7 +339,7 @@ export function main(argString = ""): void {
       myInebriety() <= inebrietyLimit() &&
       (myClass() !== $class`Seal Clubber` || !have($skill`Furious Wallop`))
     )
-      stashItems.push($item`haiku katana`);
+      stashItems.push(...$items`haiku katana, Operation Patriot Shield`);
```
Tokoeka, line 342: In a similar vein of "we don't pull the haiku katana if we're a seal clubber", is it worth not pulling the haiku katana if (we are a TT & we have / can pull the OPS)? and not pulling the OPS if we can pull a haiku katana otherwise? 

I'm not certain how easy that would be to code as I admittedly don't fully understand the withStash inner workings.

Tokoeka (approved):

src/index.ts
horrible-little-slime: Will address in subsequent PR

PR #367, opened 2021-08-12
don't add bonusequip values of 0

horrible-little-slime:


horrible-little-slime:
just gonna change libram

PR #373, opened 2021-08-13
Move free runs to FreeFight handling

Rinn:
* Move free run fights out of dailies into fights.ts, make them use FreeRunFight to automatically pick and prepare a free run source if the other conditions are met
* Make function to check and fill latte, move to outfit.ts
* Add getting candy from gingerbread noon since its free (I'll add gingerbread midnight support in another PR later since the logic is more complicated)

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -1328,6 +1328,7 @@ const freeRunFightSources = [
     },
     {
       familiar: () => $familiar`Space Jellyfish`,
+      requirements: () => new Requirement([], { forceEquip = $items`Powerful Glove` });
     }
   ),
   new FreeFight(
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -61,6 +61,7 @@ import {
   $location,
   $monster,
   $monsters,
+  $phyla,
   $skill,
   $slot,
   adventureMacro,
@@ -815,7 +816,16 @@ const freeFightSources = [
       ChateauMantegna.have() &&
       !ChateauMantegna.paintingFought() &&
       (ChateauMantegna.paintingMonster()?.attributes?.includes("FREE") ?? false),
-    () => ChateauMantegna.fightPainting()
+    () => ChateauMantegna.fightPainting(),
+    {
+      familiar: () =>
+        have($familiar`Robortender`) &&
+        $phyla`elf, fish, hobo, penguin, constellation`.some(
+          (phylum) => phylum === ChateauMantegna.paintingMonster()?.phylum
+        )
+          ? $familiar`Robortender`
+          : null,
+    }
   ),
 
   new FreeFight(
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -1328,7 +1328,7 @@ const freeRunFightSources = [
     },
     {
       familiar: () => $familiar`Space Jellyfish`,
-      requirements: () => [new Requirement([], { forceEquip = $items`Powerful Glove` })],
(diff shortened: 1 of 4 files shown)


src/fights.ts
```diff
@@ -733,6 +736,44 @@ class FreeFight {
   }
 }
 
+class FreeRunFight {
```
horrible-little-slime, line 739: Should this extend the existing freefight class instead of standing on its own?

src/fights.ts
```diff
@@ -733,6 +736,44 @@ class FreeFight {
...
+        this.options.familiar ? this.options.familiar() ?? freeFightFamiliar() : freeFightFamiliar()
+      );
+      if (runSource.prepare) runSource.prepare();
+      freeFightMood().execute();
+      freeFightOutfit([
+        ...(this.options.requirements ? this.options.requirements() : []),
+        ...(runSource.requirement ? [runSource.requirement] : []),
+      ]);
+      safeRestore();
+      // Abort and force the run function to use the correct macro
```
horrible-little-slime, line 771: this feels like a suboptimal solution

src/outfit.ts
```diff
@@ -93,6 +96,44 @@ export function freeFightOutfit(requirements: Requirement[] = []): void {
   if (haveEquipped($item`Snow Suit`) && get("snowsuit") !== "nose") cliExecute("snowsuit nose");
 }
 
+export function checkLatte(): boolean {
```
horrible-little-slime, line 99: I think "refreshlatte" may be a more apt name--I initially implemented this not to check current ingredients, but to check unlocked ingredients

src/fights.ts
horrible-little-slime: upon review, nevermind, although i'm not quite sure why we want the withMacro wrapper here

Rinn: The withMacro wrapper is taken directly from the other FreeFight class, I needed to pass in the free run macro so entries that do other combat logic first (ex extract jelly) can do that and run away afterwards. I can make the default just use the free run but I was thinking if I made it abort that would be an explicit halt for someone making a new entry instead of wasting a run if hey needed to implement custom combat logic.

src/fights.ts
```diff
@@ -1154,6 +1204,203 @@ const freeFightSources = [
...
+        Macro.while_(
+          "!pastround 28 && hasskill CHEAT CODE: Replace Enemy",
+          Macro.skill("extract jelly").skill("CHEAT CODE: Replace Enemy")
+        )
+          .trySkill("extract jelly")
+          .step(runSource.macro)
+      );
+    },
+    {
+      familiar: () => $familiar`Space Jellyfish`,
```
horrible-little-slime, line 1330:
```suggestion
      familiar: () => $familiar`Space Jellyfish`,
      requirement: new Requirement([], { forceEquip = $items`Powerful Glove` });
```

src/fights.ts
```diff
@@ -1154,6 +1204,204 @@ const freeFightSources = [
...
+      requirements: () => new Requirement([], { forceEquip = $items`Powerful Glove` });
```
Rinn, line 1331:
```suggestion
      requirements: () => [new Requirement([], { forceEquip = $items`Powerful Glove` })],
```

horrible-little-slime:
Sorry for introducing a bunch of merge conflicts, but the big thing is we should refrain from using strings in macros when we can instead use $items and $skills.

Phillammon (approved):

horrible-little-slime:
merged in a different form for reasons

Issue #379, opened 2021-08-15
utilization of vampyre cloake charges


docrostov:
once per combat, you can cast "become a wolf" for 1 turn of a +50% meat buff. if you use all 10 charges during your free fights prior to your platinum yendorian express card, that's 15 turns of 50% meat, all on embezzlers. so that's 7500 meat. it isn't much, but it's free! :V

PR #380, opened 2021-08-15
restructure freeFightOutfit; add cloake to startCombat() and freeFightOutfit

horrible-little-slime:
#379

@docrostov 

I recognize that this is pretty inelegant, but I think the core conceit of what I'm aiming for is good; I'd love advice on how to do that in a less ugly and terrible way. I couldn't use a standard destructuring method because a lot of the fields in maximizeOptions are optional

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -265,10 +265,7 @@ export class Macro extends LibramMacro {
   startCombat(): Macro {
     return this.tryHaveSkill($skill`Sing Along`)
       .tryHaveSkill($skill`Curse of Weaksauce`)
-      .externalIf(
-        haveEquipped($item`vampyric cloake`) && get("_vampyreCloakeFormUses") < 10,
-        Macro.trySkill($skill`Become a Wolf`)
-      )
+      .externalIf(haveSkill($skill`Become a Wolf`), Macro.trySkill($skill`Become a Wolf`))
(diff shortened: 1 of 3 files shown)


src/combat.ts
```diff
@@ -265,6 +265,7 @@ export class Macro extends LibramMacro {
   startCombat(): Macro {
     return this.tryHaveSkill($skill`Sing Along`)
       .tryHaveSkill($skill`Curse of Weaksauce`)
+      .externalIf(haveSkill($skill`Become a Wolf`), Macro.trySkill($skill`Become a Wolf`))
```
burningbman, line 268: Why do we need the externalIf? Isn't trySkill sufficient?

src/outfit.ts
```diff
@@ -51,43 +51,55 @@ export function freeFightOutfit(requirements: Requirement[] = []): void {
   const equipMode =
     myFamiliar() === $familiar`Machine Elf` ? BonusEquipMode.DMT : BonusEquipMode.FREE;
   const bjornChoice = pickBjorn(equipMode);
+  const compiledRequirements = Requirement.merge([new Requirement([], {}), ...requirements]);
```
burningbman, line 54: Don't need the first empty requirement after your change to the merge logic.

Rinn (approved):

PR #384, opened 2021-08-16
make LOV Extraterrestrial chocolate worth only 15k

horrible-little-slime:
While sellbot does theoretically sell it for 20k, 1., we only get 18k from it, and 2., it never sells


horrible-little-slime:
Doing something spicier

PR #385, opened 2021-08-16
implement glitch reward before fighting

lewismd13:
with a check to see if it was done by breakfast first


src/fights.ts
```diff
@@ -811,6 +811,9 @@ const freeFightSources = [
     () => have($item`[glitch season reward name]`) && !get("_glitchMonsterFights"),
     () => {
       retrieveItem($item`[glitch season reward name]`);
+      if (get("_glitchItemImplemented") === false) {
```
horrible-little-slime, line 814:
```suggestion
      if (!get("_glitchItemImplemented")) {
```

PR #391, opened 2021-08-17
Refactor class-chocolates logic

s-k-z:
Calculate expected value of class chocolates. Uses the one with the highest expected value if it's greater than zero.

### DIFF
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -276,19 +276,29 @@ export function runDiet(): void {
   );
   use(loveChocolateEat, $item`LOV Extraterrestrial Chocolate`);
 
-  const choco = new Map([
-    [toInt($class`Seal Clubber`), $item`chocolate seal-clubbing club`],
-    [toInt($class`Turtle Tamer`), $item`chocolate turtle totem`],
-    [toInt($class`Pastamancer`), $item`chocolate pasta spoon`],
-    [toInt($class`Sauceror`), $item`chocolate saucepan`],
-    [toInt($class`Accordion Thief`), $item`chocolate stolen accordion`],
-    [toInt($class`Disco Bandit`), $item`chocolate disco ball`],
+  const chocos = new Map([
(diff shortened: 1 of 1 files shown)


src/diet.ts
```diff
@@ -276,19 +276,28 @@ export function runDiet(): void {
...
-    const count = clamp(3 - used, 0, 3);
-    use(count, item);
+  const classChoco = chocos.get(myClass());
+  const chocExpVal = (remaining: number, item: Item): number => {
+    const advs = [0, 0, 1, 2, 3][remaining + (item === classChoco ? 1 : 0)];
+    return advs * MPA - mallPrice(item);
+  };
+  const chocosRemaining = clamp(3 - get("_chocolatesUsed"), 0, 3);
+  for (let i = chocosRemaining; i > 0; i--) {
+    const chocoVals = Object.values(chocos).map((choc): [Item, number] => [
```
horrible-little-slime, line 294:
```suggestion
    const chocoVals = Array.from(choco.values()).map((choc): [Item, number] => [
```

src/diet.ts
```diff
@@ -276,19 +276,28 @@ export function runDiet(): void {
...
-    const count = clamp(3 - used, 0, 3);
-    use(count, item);
+  const classChoco = chocos.get(myClass());
+  const chocExpVal = (remaining: number, item: Item): number => {
+    const advs = [0, 0, 1, 2, 3][remaining + (item === classChoco ? 1 : 0)];
+    return advs * MPA - mallPrice(item);
+  };
+  const chocosRemaining = clamp(3 - get("_chocolatesUsed"), 0, 3);
+  for (let i = chocosRemaining; i > 0; i--) {
+    const chocoVals = Array.from(choco.values()).map((choc): [Item, number] => [
```
horrible-little-slime, line 294:
```suggestion
    const chocoVals = Array.from(chocos.values()).map((choc): [Item, number] => [
```

that one's on me

PR #397, opened 2021-08-18
add nobarf option

burningbman:
Add nobarf option when running garbo.
Some embezzler code had to be moved from fight.ts to embezzler.ts to prevent circular dependencies when calculating estimatedTurns.

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -693,7 +693,7 @@ export function logMessage(message: string): void {
 }
 
 export function printLog(color: string): void {
-  for (const message in log) {
+  for (const message of log) {
     print(message, color);
   }
 }


horrible-little-slime:
`lib.ts` still imports estimatedTurns(), and `embezzler.ts` still imports `Requirement` and `sum` from `lib.ts`. If we switch to using Libram's `Requirement`, we can remove that import, and then you can define sum() inside embezzlers.ts or replace the uses of sum() with reduce(). 

That being said, I should totally add sum to libram's utils.ts

burningbman:
Switching to libram's Requirement did not go well today. Free fights didn't happen because necessary equipment was not equipped. Will give it another go after this loop, but might just have to put the changes as-is here and see if you can help figure out what is going on with the Requirements.

horrible-little-slime:
Definitely send a lot in the discord; it could be that lodash's merge isn't
deep enough

PR #402, opened 2021-08-20
Free run for gingerbread cigarettes

Rinn:
#204

Could make this more complex and get more expensive items if you have a ton of sprinkles, but this is a good start.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -1379,11 +1379,43 @@ const freeRunFightSources = [
       get("_gingerbreadCityTurns") + (get("_gingerbreadClockAdvanced") ? 5 : 0) === 9,
     () => {
       propertyManager.setChoices({
-        1204: 1, //Gingerbread Train Station Noon random candy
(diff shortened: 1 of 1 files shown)


src/fights.ts
```diff
@@ -1379,11 +1379,43 @@ const freeRunFightSources = [
...
-        1204: 1, //Gingerbread Train Station Noon random candy
+        1204: 1, // Gingerbread Train Station Noon random candy
       });
       adventureMacro($location`Gingerbread Train Station`, Macro.abort());
     }
   ),
+  new FreeRunFight(
+    () =>
+      (get("gingerbreadCityAvailable") || get("_gingerbreadCityToday")) &&
+      get("_gingerbreadCityTurns") + (get("_gingerbreadClockAdvanced") ? 5 : 0) < 19 &&
```
horrible-little-slime, line 1390: I know it's largely redundant, but also make sure that it's past noon.

PR #405, opened 2021-08-21
Use meteor shower on embezzlers near end of day

Rinn:
Use if adventures remaining is less than 150 on any digitized embezzlers for the +lbs.


horrible-little-slime:
is it worth using the digitize prediction function to determine exactly how many embezzlers we expect to encounter?

Rinn:
> is it worth using the digitize prediction function to determine exactly how many embezzlers we expect to encounter?

idk what the change would be to start using it when `5 - get("_meteorShowerUses")` digitizes are remaining

PR #412, opened 2021-08-22
Add a setting to interrupt garbo

Rinn:
I find myself slamming on esc a lot to stop garbo when testing, so here is a way to abort it safely from the relay script instead.

Current points where the abort can happen:

- After a single free fight or kill
- After the entire embezzler meat chain is finished
- After the entire embezzler weight chain is finished
- After any other embezzler fight
- After a single barf fight

### DIFF
--- a/mafia/data/garbo_settings.txt
+++ b/mafia/data/garbo_settings.txt
@@ -1,9 +1,10 @@
 # variable timing, number, variable name, variable type, description
 # each separated by a tab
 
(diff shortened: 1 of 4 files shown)


horrible-little-slime:
will abort() behave nicely with our try...finallys? Will it appropriately return items to the stash and stuff like that?

Rinn:
> will abort() behave nicely with our try...finallys? Will it appropriately return items to the stash and stuff like that?

yes

PR #413, opened 2021-08-23
Add trash and DJ NEP quests.

phulin:
Also adds megawoots quest but doesn't have the free kill logic to complete it yet.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -1693,7 +1693,7 @@ function setNepQuestChoicesAndPrepItems() {
     if (!questStep("_questPartyFair")) {
       setChoice(1324, 2); // Check out the kitchen
       setChoice(1326, 3); // Talk to the woman
-    } else {
+    } else if (get("choiceAdventure1324") !== 5) {
       setChoice(1324, 5);
       print("Found Geraldine!", "blue");
       // Format of this property is count, space, item ID.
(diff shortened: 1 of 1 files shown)


horrible-little-slime (approved):
Will merge once tested

src/fights.ts
```diff
@@ -1600,21 +1607,58 @@ export function freeFights(): void {
...
+    }
+    if (["food", "booze", "woots", "trash", "dj"].includes(quest)) {
+      runChoice(1); // Accept quest
     } else {
-      runChoice(2);
-      setChoice(1324, 5);
+      runChoice(2); // Decline quest
+    }
+  }
+
```
horrible-little-slime, line 1624: Why not just initialize `quest` as a const here?

src/fights.ts
```diff
@@ -1600,21 +1607,58 @@ export function freeFights(): void {
...
+    set("_garbo_nepUsedElectronicsKit", true);
+  }
+  if (get("lastEncounter") === "Gone Kitchin'" && lastDecision() === 3) {
+    print("Found Geraldine!", "blue");
+  }
+  if (get("lastEncounter") === "Forward to the Back" && lastDecision() === 3) {
+    print("Found Gerald!", "blue");
+  }
+
+  if (quest === "food") {
```
horrible-little-slime, line 1638: It looks like we never set 1324 back to 5 for gerald/ine quests--in the event that we get a noncombat early enough to get a second one, shouldn't we set it back to 5 once we've found our food/booze target

PR #416, opened 2021-08-25
enamorang

horrible-little-slime:


Rinn:
Seems fine, my only thoughts is that someone might not want to use an untradeable limited resource like this though.

PR #421, opened 2021-08-28
use universal seasoning

### LINKED ISSUE #420: Use universal seasoning before diet
I don't think it can have a negative effect so why not

horrible-little-slime:
Resolves #420 (blaze it)


horrible-little-slime:
Concern: it may give passive damage effects

PR #429, opened 2021-09-01
Free run from BASIC Elemental with summon mayfly swarm

Rinn:
I'm not even really sure if this is worth doing but maybe someone will get extremely lucky and get an ultra-rare.


soolar:
How does this guarantee you get a basic elemental and not a fruit golem or knob goblin mutant?

And why not just use the necklace at barf mountain and get a few more funfunds instead?

PR #430, opened 2021-09-01
Handle beaten up from mining

### LINKED ISSUE #343: Gets beaten up during volcano mining sometimes, and that's a bad thing!
Appears to cause problems with future daily tasks that require having adventures--kardashian gin and deckcheats being the big two

Rinn:
Resolves #343


horrible-little-slime:
For the future, we could consider `while`ing this

PR #433, opened 2021-09-01
Get God Lobster equipment then dish of clarified butter

Rinn:


horrible-little-slime:
Is it worth not gunning for butter if we're ascending & don't have enough equipment to get there?

Rinn:
> Is it worth not gunning for butter if we're ascending & don't have enough equipment to get there?

my thought was the stats are so nominal it wouldn't really matter

PR #434, opened 2021-09-01
uneffect dangerous effects for stasis

horrible-little-slime:
we currently get random effects from universal seasoning & glenn's golden dice, which is bad! 

Tied behind a user pref because we only need to call it once and it's a lot of effects.

Credit to aen for the list.

### DIFF
--- a/src/mood.ts
+++ b/src/mood.ts
@@ -117,6 +117,7 @@ export function meatMood(urKels = false, embezzlers = false): Mood {
   }
 
   potionSetup(embezzlers);
+  shrugPassiveDamage();
 
   return mood;
 }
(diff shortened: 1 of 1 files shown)


Rinn:
I think you don't want the pref though in case you happen to get a bad effect from a random source (if that ends up being added), and even if it's a lot tearing through an array of this size is going to be unnoticeable. Also don't you need this in the met mood as well since we rely on not doing extra damage when estimating the turns before the pointer ring crit.

horrible-little-slime:
I put it in freeFightMood because that's where golden dice gets used, but you're right if we're thinking about crazyhorse we should put it in both moods with no preference. changing now and then squooshing.

PR #438, opened 2021-09-04
Add initial support for Fire extinguisher

s-k-z:
After a cursory search these are the candidates that are worthwhile to grab with Fire Extinguisher:

- transdermal smoke patch
- tattered scraps of paper
- perfect icecube

Since free-running from Witchess is a complete waste, I decided to skip that one. Getting underwater-breathing/fishy for the ice hole is more complicated as well. Any additional items can be added to the [Item,[Location,Monster]] data structure.


src/fights.ts
```diff
@@ -1529,8 +1530,73 @@ const freeRunFightSources = [
...
+      ],
+    }
+  ),
 ];
 
+// Need to define this outside the freeRunFightSources to not duplicate a bunch of logic
+const fireExtinguishZones = new Map<Location, () => boolean>([
+  [$location`The Deep Dark Jungle`, () => get("_spookyAirportToday") || get("spookyAirportAlways")],
+  [$location`The Ice Hotel`, () => get("_coldAirportToday") || get("coldAirportAlways")],
+  [$location`The Haunted Library`, () => canAdv($location`The Haunted Library`, false)],
```
horrible-little-slime, line 1597: consider using have(library key)--canAdv isn't always as reliable as it could be

src/fights.ts
```diff
@@ -1529,8 +1530,73 @@ const freeRunFightSources = [
...
+  new FreeRunFight(
+    () =>
+      // eslint-disable-next-line libram/verify-constants
+      have($item`industrial fire extinguisher`) &&
+      get("_fireExtinguisherCharge") >= 10 &&
+      have($skill`Comprehensive Cartography`) &&
+      get("_monstersMapped") < 3 &&
+      Array.from(fireExtinguishZones.values()).some((fez) => fez()),
+    (runSource: FreeRun) => {
+      const targets = new Map<Item, { location: Location; monster: Monster }>();
```
horrible-little-slime, line 1542: Merge this map with the location -> boolean one

src/fights.ts
```diff
@@ -1529,8 +1530,73 @@ const freeRunFightSources = [
...
     }
   ),
+  new FreeRunFight(
+    () =>
+      // eslint-disable-next-line libram/verify-constants
+      have($item`industrial fire extinguisher`) &&
+      get("_fireExtinguisherCharge") >= 10 &&
+      have($skill`Comprehensive Cartography`) &&
+      get("_monstersMapped") < 3 &&
+      Array.from(fireExtinguishZones.values()).some((fez) => fez()),
```
horrible-little-slime, line 1540: I think the use of `fez`makes this harder to read than perhaps is optimal

horrible-little-slime:
We run regular freeFights after freeRunFights--is it worth adding a () => Requirement and a custom macro to the witchess fights, lest we somehow find ourselves unable to do any of the free run versions?

src/fights.ts
```diff
@@ -1529,8 +1530,61 @@ const freeRunFightSources = [
...
+    monster: $monster`ice bartender`,
+    item: $item`perfect ice cube`,
+    open: () => get("_coldAirportToday") || get("coldAirportAlways"),
+  },
+
+  {
+    location: $location`The Haunted Library`,
+    monster: $monster`bookbat`,
+    item: $item`tattered scrap of paper`,
+    open: () => () => have($item`[7302]Spookyraven library key`),
```
horrible-little-slime, line 1584: I don't think you need the double () => or the [7302]

s-k-z: The [7302] is necessary, there are two keys of the same name.

s-k-z:
> 
> 
> We run regular freeFights after freeRunFights--is it worth adding a () => Requirement and a custom macro to the witchess fights, lest we somehow find ourselves unable to do any of the free run versions?

Not seeing any examples where requirements are optional. Is that supported?

horrible-little-slime:
https://github.com/Loathing-Associates-Scripting-Society/garbage-collector/blob/5a10b402e2a64e574f1d72e9f458ae3371f24296/src/fights.ts#L717-L721

FreeFightOptions includes an optional () => Requirement[] entry, so you make it check whether we have the fire extinguisher and have foam before pushing it as a requirement

s-k-z:
> 
> 
> https://github.com/Loathing-Associates-Scripting-Society/garbage-collector/blob/5a10b402e2a64e574f1d72e9f458ae3371f24296/src/fights.ts#L717-L721
> 
> FreeFightOptions includes an optional () => Requirement[] entry, so you make it check whether we have the fire extinguisher and have foam before pushing it as a requirement

Oh, I meant conditional, not optional.

s-k-z:
Okay, Witchess monsters aren't pickpocketable, so I reverted those commits.

s-k-z:
Vortexing he bookbat without killing it requires maximizing ML. Doing so across the board makes the smoke monster harder to fight since that one can scale up to 10000 attack. There doesn't seem to be any elegant solution to both that I see.

s-k-z:
This doesn't really work if there's a Vykea furniture active. They do way too much damage.

horrible-little-slime:
Rebase and move the vykea stuff and I'll merge

PR #449, opened 2021-09-07
forbid dance police

horrible-little-slime:


src/index.ts
```diff
@@ -201,7 +204,12 @@ export function canContinue(): boolean {
 }
 
 export function main(argString = ""): void {
-  sinceKolmafiaRevision(20815);
+  sinceKolmafiaRevision(20901);
+  const forbiddenStores = property.getString("forbiddenStores").split(",");
+  if (!forbiddenStores.includes(getPlayerId("Dance Police"))) {
```
phulin, line 209: getPlayerId does a /whois request which is unreliable - can we use a fixed player id instead?

PR #455, opened 2021-09-08
Use libram requirement, add nobarf option

burningbman:


src/combat.ts
```diff
@@ -30,7 +37,83 @@ import {
   Macro as LibramMacro,
   SourceTerminal,
 } from "libram";
-import { maxPassiveDamage, monsterManuelAvailable } from "./lib";
```
burningbman, line 33: Brought these methods into this file to remove a circular dependency (since lib uses embezzlers file)

src/fights.ts
```diff
@@ -116,56 +108,16 @@ import {
   waterBreathingEquipment,
 } from "./outfit";
 import { bathroomFinance } from "./potions";
-import { estimatedTurns, globalOptions, log } from "./globalvars";
-
-function checkFax(): boolean {
```
burningbman, line 121: These methods were moved to the new embezzlers file

src/fights.ts
```diff
@@ -208,222 +160,6 @@ const secondChainMacro = () =>
     ).meatKill()
   ).abort();
 
-const embezzlerMacro = () =>
```
burningbman, line 211: These lines were moved to the new embezzlers file

src/fights.ts
```diff
@@ -1828,10 +1564,22 @@ function deliverThesis(): void {
     outfit("checkpoint");
   }
   cliExecute("gain 1800 muscle");
+
+  let thesisLocation = $location`Uncle Gator's Country Fun-Time Liquid Waste Sluice`;
```
burningbman, line 1568: Since the user might not have access to barf now, have to find another place to deliver thesis

src/index.ts
```diff
@@ -343,30 +340,32 @@ export function main(argString = ""): void {
         // 0. diet stuff.
         runDiet();
 
-        // 1. get a ticket
-        ensureBarfAccess();
```
burningbman, line 347: This is now done later after checking nobarf option

src/lib.ts
```diff
@@ -319,45 +302,6 @@ function testZoneForWanderers(location: Location): boolean {
   return !wandererBlacklist.includes(location) && location.wanderers;
 }
 
-export class Requirement {
```
burningbman, line 322: Now use libram Requirement!

src/lib.ts
```diff
@@ -615,82 +559,6 @@ export function fairyMultiplier(familiar: Familiar): number {
   return Math.pow(Math.sqrt(itemBonus + 55 / 4 + 3) - Math.sqrt(55) / 2, 2);
 }
 
-let monsterManuelCached: boolean | undefined = undefined;
```
burningbman, line 618: Moved to combat.ts to remove circular dependency (since lib needs to use embezzler)

src/fights.ts
```diff
@@ -1828,10 +1564,22 @@ function deliverThesis(): void {
...
+  if (thesisInNEP) {
+    thesisLocation = $location`The Neverending Party`;
+  }
+  // if running nobarf, might not have access to Uncle Gator's. Space is cheaper.
+  else if (!(get("stenchAirportAlways") || get("_stenchAirportToday"))) {
```
horrible-little-slime, line 1573: `thesisInNEP` keys off of having NEP fights left--I think it's worth checking the NEP before buying a transponder, because spending a non-free turn in the NEP is still better than spending however much on a transponder (I know they're cheap, but still)

src/globalvars.ts
```diff
@@ -1,22 +1,15 @@
...
+    ascending: false,
+    noBarf: false,
+  };
 
-export function estimatedTurns(): number {
-  return globalOptions.stopTurncount
-    ? globalOptions.stopTurncount - myTurncount()
-    : (myAdventures() + (globalOptions.ascending && myInebriety() <= inebrietyLimit() ? 60 : 0)) *
-        (have($item`mafia thumb ring`) ? 1.04 : 1);
+export function clamp(n: number, min: number, max: number): number {
```
horrible-little-slime, line 13: Libram should export its utility functions now, so we can get rid of this export. That can be a job for another day though.

horrible-little-slime (changes requested):
Update the thesis code to check the NEP before buying a transponder, and stop exporting clamp if we can get it from libram. After that I'll squoosh immediately!

PR #461, opened 2021-09-09
Don't buy internet items we already have

burningbman:


phulin:
I don't think we should merge this - you can easily sell these in the mall.

PR #464, opened 2021-09-10
deal with lov chocolates in a way that is both way longer and way better

horrible-little-slime:


src/diet.ts
```diff
@@ -277,12 +279,17 @@ export function runDiet(): void {
...
+  while (get("_loveChocolatesUsed") < 3) {
+    const price = have($item`LOV Extraterrestrial Chocolate`) ? 15000 : 20000;
+    const value = clamp(3 - get("_loveChocolatesUsed"), 0, 3) * get("valueOfAdventure");
+    if (value < price) break;
+    if (!have($item`LOV Extraterrestrial Chocolate`)) {
+      Kmail.send("sellbot", `${$item`LOV Extraterrestrial Chocolate`.name} (1)`, undefined, 20000);
+      wait(11);
+      cliExecute("refresh inventory");
+    }
+    use($item`LOV Extraterrestrial Chocolate`);
```
Rinn, line 291: should put a break here if the kmail request doesn't return in time and you still don't have the chocolate

Rinn (approved):

PR #466, opened 2021-09-13
Fix some free runs when running nobarf without access to dinseyland

gausie:


horrible-little-slime (changes requested):
Won't this try to adventure in $location`none` when you don't have access to any?

Can we add having a stench location as a condition to trying to do jelly?

PR #467, opened 2021-09-13
sanity price check on glarks

horrible-little-slime:


src/fights.ts
```diff
@@ -886,9 +887,11 @@ const freeFightSources = [
   ),
 
   new FreeFight(
-    () => (get("questL11Ron") === "finished" ? 5 - get("_glarkCableUses") : 0),
+    () =>
+      get("questL11Ron") === "finished"
+        ? 5 - get("_glarkCableUses")
+        : 0 && have($item`glark cable`),
```
burningbman, line 893: Doesn't this always come out to 0?

burningbman (approved):

src/fights.ts
horrible-little-slime: oh whoops

horrible-little-slime: How does it look now

burningbman (approved):

src/fights.ts
burningbman: Looks good. Approved.

PR #469, opened 2021-09-14
buy chocolate at reasonable price

horrible-little-slime:


src/diet.ts
```diff
@@ -318,7 +318,7 @@ export function runDiet(): void {
       };
     });
     const best = chocoVals.sort((a, b) => b.value - a.value)[0];
-    acquire(1, best.choco, best.value, false);
+    acquire(1, best.choco, best.value + get("valueOfAdventure"), false);
```
burningbman, line 321: Shouldn't it add back `mallPrice(best.choco)`, since that is what's subtracted when calculating the best choco?

horrible-little-slime: I think in retrospect what this means is that we shouldn't use acquire at all

horrible-little-slime: yeah i'm very dumb

PR #470, opened 2021-09-15
Use ugly checkboxes in relay

burningbman:
They're ugly right now, but at least you click instead of typing true/false


horrible-little-slime:
If I haven't merged this by Friday, remind me to; I want to test it, and I am extremely forgetful, and cannot do it right now.

PR #475, opened 2021-09-16
list potential embezzler sources for diagnostic and other reasons

horrible-little-slime:


src/fights.ts
```diff
@@ -261,29 +261,34 @@ function getEmbezzlerFight(): EmbezzlerFight | null {
...
-        visitUrl(
-          "choice.php?pwd&whichchoice=1267&option=1&wish=to fight a Knob Goblin Embezzler ",
-          true,
-          true
-        );
-        visitUrl("main.php", false);
-        runCombat();
-      }
-    );
+  if (potential > 0 && get("_genieFightsUsed") < 3) {
```
Rinn, line 264: I'd say just always print them before the genie check so they end up in the log if someone has already used their wishes.

Rinn (approved):

PR #476, opened 2021-09-16
Option to disable borrowing from clan stashes

Rinn:
By popular demand.


horrible-little-slime:
I feel like this should just be part of the stashClan property, maybe set it to "null" or "None" or "No" or 0 or "fuck off" or something like that

PR #477, opened 2021-09-17
Check for turtle tamer adventures while mapping monsters

s-k-z:


horrible-little-slime:
I sometimes run into a similar issue while checking WLF? Is there a way to generalize this solution?

s-k-z:
You can get turtle adventures at the WLF bunker?

s-k-z:
I think we can make checking for turtle adventures more general, but it needs to be applied in any place where we do visitUrl with an adv.php location and for gingerbread city it seems to be necessary for the counters there which uses adv1.

horrible-little-slime:
Either that or wherever we go right before then.

PR #478, opened 2021-09-17
only prompt if expected copy value is more than the cost of a pocket wish

lewismd13:


src/fights.ts
```diff
@@ -261,14 +261,15 @@ function getEmbezzlerFight(): EmbezzlerFight | null {
...
-
+  const averageEmbezzlerNet = ((baseMeat + 750) * meatDropModifier()) / 100;
   if (potential > 0) {
     print(`You have the following embezzler-sources untapped right now:`, "blue");
     embezzlerSources
       .filter((source) => !source.available() && source.potential() > 0)
       .map((source) => `${source.potential()} from ${source.name}`)
       .forEach((text) => print(text, "blue"));
     if (
+      Math.floor((potential * averageEmbezzlerNet) / 50000) > 0 &&
```
horrible-little-slime, line 272: This feels like a profoundly weird way to do the math on this--is there a reason you've chosen to execute it this way?

src/fights.ts
```diff
@@ -261,14 +261,15 @@ function getEmbezzlerFight(): EmbezzlerFight | null {
     if (fight.available()) return fight;
   }
   const potential = embezzlerCount();
-
+  const averageEmbezzlerNet = ((baseMeat + 750) * meatDropModifier()) / 100;
```
horrible-little-slime, line 264: Per discussion in discord, this is weird but fine for now

src/fights.ts
lewismd13: I swear it made sense in my head at that the time. We can just do potential * averageEmbezzlerNet >50000 I suppose

horrible-little-slime: Make it so and I'll push it

src/fights.ts
```diff
@@ -261,14 +261,15 @@ function getEmbezzlerFight(): EmbezzlerFight | null {
...
   if (potential > 0) {
     print(`You have the following embezzler-sources untapped right now:`, "blue");
     embezzlerSources
       .filter((source) => !source.available() && source.potential() > 0)
       .map((source) => `${source.potential()} from ${source.name}`)
       .forEach((text) => print(text, "blue"));
     if (
+      potential * averageEmbezzlerNet > 50000 &&
```
horrible-little-slime, line 272: (potential + 1) * (averageEmbezzlerNet - get("valueOfAdventure"))

horrible-little-slime (changes requested):

PR #479, opened 2021-09-17
make enamorang more like digitize in startDigitize

horrible-little-slime:


src/fights.ts
```diff
@@ -288,11 +288,13 @@ function getEmbezzlerFight(): EmbezzlerFight | null {
   return null;
 }
 
-function startDigitize() {
+function startWandererCounters() {
   if (
```
Rinn, line 292: Could we also check if a guaranteed sausage goblin, free vote monster, or ghost bustin' happens to be ready and do those instead of the free run?

horrible-little-slime: None of those will start the counter. Which, oof, I gotta revise freecandy.

horrible-little-slime: Specifically, all 3 of those are above digitize in the encounter ontology. In order to initialize a counter like digitize, we must pass it in the ontology.

Rinn: okay makes sense. I would maybe change this to check if the monster is a crate and only run from that, so it doesn't accidently run away from a wanderer we want to fight, and maybe this should keep looping until the last monster encountered was a crate?

horrible-little-slime: I think checking for KGEs and not running from those seems smart, but mafia I believe is smart enough to know if the counter got initialized, so the worst case scenario is we run it a second time. I am a little nervous about looping until we hit a crate, that feels dangerous for reasons I can't articulate.

horrible-little-slime: Alright, it now runs until it hits a crate, but will at most do it thrice. Should I throw if we reach it thrice, or just keep grindin'?

PR #487, opened 2021-09-25
help menu

JamesDowney:
This isn't perfect, but given the issues I've had with html formatting and wanting to come up with a more perfect final solution, this is what I have in the meantime.

image


JamesDowney:
> This isn't perfect, but given the issues I've had with html formatting and wanting to come up with a more perfect final solution, this is what I have in the meantime.
> 
> image

There is a typo/some junk in the description of valueOfAdventure, but that's only in the screenshot, it has since been corrected.

PR #494, opened 2021-09-27
refresh inventory before retrieving item so this stops happening

horrible-little-slime:


JamesDowney (approved):

phulin:
I think this is only a problem with VSRD, and we should really just fix Mafia. This will refresh every turn even we're not about to hit the NC.

horrible-little-slime:
Closed per bean's remark

PR #499, opened 2021-10-02
saberforce for pills

horrible-little-slime:


horrible-little-slime:
@docrostov become a LASS so I can formally request your review

src/fights.ts
```diff
@@ -738,6 +739,54 @@ const freeFightSources = [
...
+  new FreeFight(
+    () => (wantPills() ? clamp(5 - get("_saberForceUses"), 0, 3 - get("_monstersMapped")) : 0),
+    () => {
+      ensureEffect($effect`Transpondent`);
+      if (have($familiar`Red-Nosed Snapper`)) cliExecute(`snapper ${$phylum`dude`}`);
+      setChoice(1387, 3);
+      if (
+        have($skill`Comprehensive Cartography`) &&
+        get("_monstersMapped") <
+          (getBestFireExtinguisherZone() && get("_fireExtinguisherCharge") >= 10 ? 2 : 3)
```
burningbman, line 751: Why does fire extinguisher charge count matter?

src/index.ts
```diff
@@ -167,20 +169,35 @@ function barfTurn() {
     // d. get dressed
     meatOutfit(embezzlerUp, [], underwater);
 
-    adventureMacroAuto(
-      location,
-      Macro.externalIf(
-        underwater,
+    if (
```
burningbman, line 172: Do we want to do this even if nobarf is true?

src/index.ts
```diff
@@ -167,20 +169,35 @@ function barfTurn() {
...
+      !embezzlerUp &&
+      myInebriety() > inebrietyLimit() &&
+      globalOptions.ascending &&
+      estimatedTurns() - digitizedMonstersRemaining() <=
+        availableAmount($item`Map to Safety Shelter Grimace Prime`)
+    ) {
+      const choiceToSet =
+        availableAmount($item`distention pill`) <
+        availableAmount($item`synthetic dog hair pill`) +
+          availableAmount($item`Map to Safety Shelter Grimace Prime`)
```
burningbman, line 182: Why add `availableAmount($item'Map to Safety Shelter Grimace Prime')`? Isn't just a less than comparison sufficient?

src/fights.ts
horrible-little-slime: We want to be able to map to use polar vortex later; i should perhaps add this in with comments, because it is entirely non-obvious

src/index.ts
horrible-little-slime: I was modelling the logic on something bean did, to avoid constantly toggling the property. Let me reflect and make sure taht this makes sense.

src/index.ts
horrible-little-slime: Would you ever run `nobarf` alongside `ascend` while also overdrunk?

PR #503, opened 2021-10-04
cast hobo songs when relevant & able

horrible-little-slime:


src/mood.ts
```diff
@@ -126,6 +128,19 @@ export function meatMood(urKels = false, embezzlers = false): Mood {
     }
   }
 
+  if (
+    myClass() === $class`Accordion Thief` &&
+    myLevel() >= 15 &&
+    !(
+      getWorkshed() === $item`warbear LP-ROM burner` ||
+      have($item`warbear LP-ROM burner` || get("questG04Nemesis") === "finished")
+    )
```
phulin, line 137: I would pull this out into a named variable

src/mood.ts
```diff
@@ -126,6 +128,19 @@ export function meatMood(urKels = false, embezzlers = false): Mood {
...
+  ) {
+    if (have($skill`The Ballad of Richie Thingfinder`))
+      useSkill($skill`The Ballad of Richie Thingfinder`);
```
phulin, line 140: We should enforce this through linting, but two-line if statements without braces are bad.

src/mood.ts
```diff
@@ -126,6 +128,19 @@ export function meatMood(urKels = false, embezzlers = false): Mood {
...
+    if (have($skill`Chorale of Companionship`)) useSkill($skill`Chorale of Companionship`);
```
phulin, line 141: Don't you want to use both of these for all remaining casts, rather than just one at a time? See the bird code for an example.

src/mood.ts
```diff
@@ -126,6 +128,19 @@ export function meatMood(urKels = false, embezzlers = false): Mood {
     }
   }
 
+  const canRecord =
+    getWorkshed() === $item`warbear LP-ROM burner` ||
+    have($item`warbear LP-ROM burner` || get("questG04Nemesis") === "finished");
+
+  if (myClass() === $class`Accordion Thief` && myLevel() >= 15 && !canRecord) {
```
Rinn, line 135: I feel like this is gonna hit the song limit all the time, if it doesn't have priority over the other songs.

horrible-little-slime: The way our song slots are set up should make that impossible

PR #506, opened 2021-10-10
cast ode before drinkin

spaghetti-squash:


horrible-little-slime:
Could not find that function when I was searching around for existing
support. Yep thats way better

PR #507, opened 2021-10-10
Fish heads are no longer worth it

s-k-z:
Fish heads have fallen in value significantly following the events of Vanduffel


phulin:
Looks good.

PR #514, opened 2021-10-16
Map monsters time prank

s-k-z:
Since time spinner pranks can show up here, we need a way to determine them and kill them.


horrible-little-slime:
This looks fine, but I'm starting to wonder if there's a more general solution to this.

I'd like to wait for someone to test it before I approve it,

PR #518, opened 2021-10-18
only get pasta eyeball if it's worth it

horrible-little-slime:


src/mood.ts
```diff
@@ -73,8 +79,23 @@ export function meatMood(urKels = false, embezzlers = false): Mood {
...
-
+  if (myClass() !== $class`Pastamancer`) {
+    const isWorthIt =
+      (baseMeat * getModifier("Meat Drop", $effect`Pasta Eyeball`)) /
+        100 /
+        mpCost($skill`Bind Lasagmbie`) >
+      Math.max(
+        (1.5 * myLevel() + 5) / 80,
+        (2.5 * myLevel()) / mallPrice($item`generic mana potion`)
+      );
```
phulin, line 90: can you break this into sub-computations with names? A little incomprehensible right now.

src/mood.ts
```diff
@@ -73,8 +79,23 @@ export function meatMood(urKels = false, embezzlers = false): Mood {
...
+    if (
+      isWorthIt ||
+      (have($item`porquoise-handled sixgun`) && monsterManuelAvailable()) ||
+      getFloristPlants()[$location`Barf Mountain`.toString()]?.includes("Pitcher Plant") ||
+      get("_bittycar") === "soulcar"
+    )
+      mood.skill($skill`Bind Lasagmbie`);
```
phulin, line 97: we should probably also apply to Spirit of Taking

src/mood.ts
horrible-little-slime: sure!

horrible-little-slime:
Closing this PR after MP changes; will keep the branch alive in case we change our minds.

PR #522, opened 2021-10-21
force garbo to only use Skill and Item inputs to macros

horrible-little-slime:


horrible-little-slime:
This will likely be obsoleted by the next libram update.

src/combat.ts
```diff
@@ -134,12 +134,49 @@ function shouldRedigitize() {
...
   submit(): string {
     print(this.components.join("\n"));
     return super.submit();
   }
 
+  skill(...skills: Skill[]): this {
+    return super.skill(...skills);
+  }
+
+  static skill<T extends LibramMacro>(this: Constructor<T>, ...skills: Skill[]): T {
```
phulin, line 149: should be `T extends Macro`, no?

horrible-little-slime:
I got red squigglies when I did that, but didn't get that for StrictMacro
in Libram. Maybe relates to reusing the name?

src/combat.ts
horrible-little-slime: It seems like, for whatever reason, VSCode hates that (red squigglies), but I can do it in my libram PR. I'd be willing to guess the issue is recycling the name 'Macro.' I'll close this PR and wait until we get StrictMacro in libram.

PR #527, opened 2021-11-02
account for using the pillkeeper semirare as a chain-starter

horrible-little-slime:


src/potions.ts
```diff
@@ -210,7 +211,12 @@ export function potionSetup(doEmbezzlers = false): void {
       // Estimate that the opportunity cost of free PK useage is 10k meat - approximately +1 embezzler.
       if (
         potion.doublingValue(embezzlers) >
-        (canAdv($location`Cobb's Knob Treasury`, false) ? 15000 : 0)
+        (canAdv($location`Cobb's Knob Treasury`, false)
```
burningbman, line 214: This if statement feels like the kind of thing that scares people from contributing to the project. Maybe add a method in lib (or a new iotm.ts?) called getPillKeeperOpportunityCost and have comments explaining the different conditions.

horrible-little-slime: I don't know if an iotm.ts feels right, and lib doesn't feel appropriate, but you're right that this should be sliced out. @docrostov is the king of clarissa explains garbo--where might we put this?

src/lib.ts
```diff
@@ -402,3 +405,12 @@ export function printHelpMenu(): void {
     |           Note:          | You can manually set these properties, but it's recommended that you use the relay interface. |
     +--------------------------+-----------------------------------------------------------------------------------------------+</pre>`);
 }
+
+export function pillkeeperOpportunityCost(): number {
```
burningbman, line 409: Need jsdocs and comments, bruv

horrible-little-slime: thanks king

burningbman (approved):

Issue #504, opened 2021-10-08
Synthesis ignores foldables


AmirhPooya:
Synthesis currently uses the cheapest possible pair of candies, which is usually bags of W&W. This ignores the much cheaper alternative that comes from the foldable, untradable sugar sheet derivatives which is sugar shorts + modulo 3 candy.

PR #528, opened 2021-11-03
Add synth code to use up untradeables.

horrible-little-slime:
This is untested, and hasn't been touched in a while; creating as draft.

@gausie is it worth adding this synth code to Libram?

### DIFF
--- a/src/synthesis.ts
+++ b/src/synthesis.ts
@@ -31,8 +31,7 @@ export default function synthesize(effect: Effect, casts: number): void {
         retrieveItem(itemA, castsToDo);
         retrieveItem(itemB, castsToDo);
       }
-      sweetSynthesis(castsToDo, itemA, itemB);
-      casts -= castsToDo;
+      if (sweetSynthesis(castsToDo, itemA, itemB)) casts -= castsToDo;
     }
   }
 
--- a/src/synthesis.ts
+++ b/src/synthesis.ts
@@ -1,7 +1,7 @@
 import {
   availableAmount,
-  itemAmount,
   mySpleenUse,
+  retrieveItem,
   spleenLimit,
   sweetSynthesis,
   sweetSynthesisResult,
@@ -21,11 +21,16 @@ export default function synthesize(effect: Effect, casts: number): void {
       if (sweetSynthesisResult(itemA, itemB) !== effect) continue;
       const possibleCasts =
         itemA === itemB
-          ? Math.floor((itemAmount(itemA) - 1) / 2)
-          : Math.min(itemAmount(itemA), itemAmount(itemB)) - 1;
+          ? Math.floor((availableAmount(itemA) - 1) / 2)
+          : Math.min(availableAmount(itemA), availableAmount(itemB)) - 1;
       const spleen = Math.max(spleenLimit() - mySpleenUse(), 0);
       const castsToDo = Math.min(possibleCasts, casts, spleen);
       if (castsToDo === 0) continue;
+      if (itemA === itemB) retrieveItem(itemA, castsToDo * 2);
+      else {
+        retrieveItem(itemA, castsToDo);
+        retrieveItem(itemB, castsToDo);
+      }
       sweetSynthesis(castsToDo, itemA, itemB);
       casts -= castsToDo;
     }
--- a/src/synthesis.ts
+++ b/src/synthesis.ts
@@ -1,0 +1,40 @@
+import {
+  availableAmount,
+  haveEffect,
+  itemAmount,
+  mySpleenUse,
+  spleenLimit,
+  sweetSynthesis,
+  sweetSynthesisResult,
+} from "kolmafia";
+import { $items, clamp } from "libram";
+import { shuffle } from "./lib";
+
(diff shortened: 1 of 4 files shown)


horrible-little-slime:
Resolves #504

horrible-little-slime:
Feedback from @lewismd13 suggests that:
1. this struggles when you leave all of your stuff in hagnks. We can add some retrieveItem calls, but ultimately the best solution is to not do that. I don't think manny has write access in LASS, so he may not be able to push that directly?
2. this fails silently when it fails, which is suboptimal. That being said, optimal behavior is not clear.

src/synthesis.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const whitelist = $items`sugar shotgun, sugar shillelagh, sugar shank, sugar chapeau, sugar shorts, sugar shield, sugar shirt, Fudgie Roll`;
+export default function synthesize(effect: Effect, casts: number): void {
+  const shuffledWhitelist = shuffle([...whitelist]);
+  for (const itemA of shuffledWhitelist) {
+    if (availableAmount(itemA) <= 1) continue;
+    if (casts === 0) return;
+    for (const itemB of shuffledWhitelist) {
+      const minimum = itemA === itemB ? 2 : 1;
+      if (availableAmount(itemB) <= minimum) continue;
```
lewismd13, line 20:
```suggestion
      if (itemAmount(itemB) <= minimum) continue;
```

src/synthesis.ts
```diff
@@ -0,0 +1,37 @@
...
+  sweetSynthesisResult,
+} from "kolmafia";
+import { $items, clamp } from "libram";
+import { shuffle } from "./lib";
+
+const whitelist = $items`sugar shotgun, sugar shillelagh, sugar shank, sugar chapeau, sugar shorts, sugar shield, sugar shirt, Fudgie Roll`;
+export default function synthesize(effect: Effect, casts: number): void {
+  const shuffledWhitelist = shuffle([...whitelist]);
+  for (const itemA of shuffledWhitelist) {
+    if (availableAmount(itemA) <= 1) continue;
```
lewismd13, line 16:
```suggestion
    if (itemAmount(itemA) <= 1) continue;
```

lewismd13 (changes requested):
I think this is cleaner than throwing in a bunch of retrieveItem()s in this section. The options are either throw in retrieveItem()s or do itemAmount for everything. I think this is better, and people like me can just add all the untradeables to their pull script. The issue that I ran into came from using availableAmount for one part and itemAmount for the other.

src/synthesis.ts
horrible-little-slime: My concern with these changes is that it won't pull candies from hagnks--wouldn't we rather use availableAmount and use retrieveItem, rather than using itemAmount to begin with?

src/synthesis.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (casts === 0) return;
+    for (const itemB of shuffledWhitelist) {
+      const minimum = itemA === itemB ? 2 : 1;
+      if (availableAmount(itemB) <= minimum) continue;
+      if (sweetSynthesisResult(itemA, itemB) !== effect) continue;
+      const possibleCasts =
+        itemA === itemB
+          ? Math.floor((itemAmount(itemA) - 1) / 2)
```
lewismd13, line 24:
```suggestion
          ? Math.floor((availableAmount(itemA) - 1) / 2)
```

src/synthesis.ts
```diff
@@ -0,0 +1,37 @@
...
+          : Math.min(itemAmount(itemA), itemAmount(itemB)) - 1;
```
lewismd13, line 25:
```suggestion
          : Math.min(availableAmount(itemA), availableAmount(itemB)) - 1;
```

src/synthesis.ts
```diff
@@ -0,0 +1,37 @@
...
+      const spleen = Math.max(spleenLimit() - mySpleenUse(), 0);
+      const castsToDo = Math.min(possibleCasts, casts, spleen);
+      if (castsToDo === 0) continue;
+      sweetSynthesis(castsToDo, itemA, itemB);
```
lewismd13, line 29:
```suggestion
      retrieveItem(itemA, castsToDo);
      retrieveItem(itemB, castsToDo);
      sweetSynthesis(castsToDo, itemA, itemB);
```

lewismd13 (dismissed):
Ok, I think this is the manny-friendly version of the fix.

src/synthesis.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  availableAmount,
+  itemAmount,
```
lewismd13, line 3:
```suggestion

```

lewismd13 (commented):
forgot one

src/synthesis.ts
horrible-little-slime: Is this safe for double-synthing a single thing?

lewismd13: that... is a good and annoying question.

lewismd13: ok using newfound powers, have pushed changes to the PR that retrieveItem as needed, and take into account double-synthing. only screwed it up once.

lewismd13 (approved):
This has not exploded in testing, so I think it's good to go.

PR #529, opened 2021-11-03
equip the blue plate if relevant

horrible-little-slime:


horrible-little-slime:
I don't think this modification justifies a jsdoc, and I feel like the omnibus daily functions are fine?

burningbman (approved):

PR #530, opened 2021-11-04
Setup KGEs after prompting user to use wish

burningbman:
Check embezzler sources before running the setup (buying tons of potions). So that the user can decide to wish for KGE before buffing up.
Also did some refactoring to take embezzler stuff out of fights.

### DIFF
--- a/src/embezzler.ts
+++ b/src/embezzler.ts
@@ -8,10 +8,16 @@ import {
   haveEquipped,
   inebrietyLimit,
   itemAmount,
+  meatDropModifier,
   myAdventures,
+  myHash,
   myInebriety,
   myTurncount,
+  print,
+  runCombat,
   use,
+  userConfirm,
+  visitUrl,
   wait,
 } from "kolmafia";
 import {
@@ -30,9 +36,10 @@ import {
   SourceTerminal,
   sum,
 } from "libram";
+import { acquire } from "./acquire";
 import { Macro } from "./combat";
+import { baseMeat, globalOptions, WISH_VALUE } from "./lib";
 import { determineDraggableZoneAndEnsureAccess, draggableFight } from "./wanderer";
-import { globalOptions } from "./lib";
 
 type EmbezzlerFightOptions = {
   location?: Location;
@@ -296,7 +303,10 @@ export const embezzlerSources = [
   new EmbezzlerFight(
     "Professor MeatChain",
     () => false,
-    () => (have($familiar`Pocket Professor`) && !get<boolean>("_garbo_meatChain", false) ? 10 : 0),
+    () =>
+      have($familiar`Pocket Professor`) && !get<boolean>("_garbo_meatChain", false)
+        ? Math.max(10 - get("_pocketProfessorLectures"), 0)
+        : 0,
     () => {
       return;
     }
@@ -304,7 +314,10 @@ export const embezzlerSources = [
   new EmbezzlerFight(
     "Professor WeightChain",
     () => false,
-    () => (have($familiar`Pocket Professor`) && !get<boolean>("_garbo_weightChain", false) ? 5 : 0),
+    () =>
+      have($familiar`Pocket Professor`) && !get<boolean>("_garbo_weightChain", false)
+        ? Math.min(15 - get("_pocketProfessorLectures"), 5)
+        : 0,
     () => {
       return;
     }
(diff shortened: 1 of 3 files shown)


src/embezzler.ts
```diff
@@ -30,8 +36,10 @@ import {
   SourceTerminal,
   sum,
 } from "libram";
+import { acquire } from "./acquire";
 import { Macro } from "./combat";
 import { globalOptions } from "./globalvars";
+import { baseMeat, WISH_VALUE } from "./lib";
```
horrible-little-slime, line 42: I'm not entirely sure I understand the utility of making WISH_VALUE a variable, but I guess it does come up a lot

burningbman: The advantage is it reduces one thing a person has to think about. Instead of seeing 50000 and having to remember it is the value of a wish, that is now explicit.

src/embezzler.ts
```diff
@@ -7,11 +7,17 @@ import {
   haveEquipped,
   inebrietyLimit,
   itemAmount,
+  meatDropModifier,
```
horrible-little-slime, line 10: Style question: To what extent is meatDropModifier different from numericModifier("Meat Drop")/getModifier("Meat Drop")? Is there a compelling reason for us to consistently pick one of these?

src/embezzler.ts
```diff
@@ -340,3 +353,57 @@ export function estimatedTurns(): number {
...
+      )
+    ) {
+      acquire(1, $item`pocket wish`, WISH_VALUE);
+      const id = "Pocket Wish";
+      const wishFight = new EmbezzlerFight(
+        id,
+        () => true, // available so that future getNextEmbezzlerFight requests don't queue up more pocket wish KGEs
+        () => 1,
+        () => {
+          visitUrl(`inv_use.php?pwd=${myHash()}&which=3&whichitem=9537`, false, true);
```
horrible-little-slime, line 387: can we not just use the &pwd here? Doesn't mafia typically expand pwd into pwd=[hash]?

src/embezzler.ts
```diff
@@ -340,3 +353,57 @@ export function estimatedTurns(): number {
...
+  const potential = embezzlerCount();
+  const averageEmbezzlerNet = ((baseMeat + 750) * meatDropModifier()) / 100;
+  if (potential > 0) {
+    print(`You have the following embezzler-sources untapped right now:`, "blue");
+    embezzlerSources
+      .filter((source) => !source.available() && source.potential() > 0)
+      .map((source) => `${source.potential()} from ${source.name}`)
+      .forEach((text) => print(text, "blue"));
+    if (
+      get("_genieFightsUsed") < 3 &&
```
horrible-little-slime, line 375: Good catch! I don't think we were checking this before!

src/embezzler.ts
```diff
@@ -340,3 +353,57 @@ export function estimatedTurns(): number {
...
+          runCombat();
+
+          // remove the pocket wish embezzler source
+          const index = embezzlerSources.findIndex((source) => source.name === id);
+          embezzlerSources.splice(index, 1);
+        }
+      );
+
+      // add the pocket wish as the first embezzler source
+      embezzlerSources.splice(0, 0, wishFight);
```
horrible-little-slime, line 403: I am not sure why we are adding & removing the wishFight from embezzlerSources. Is just returning the fight insufficient?

burningbman: The problem with just returning the fight is that there are cases where the user will get prompted multiple times. Any time we come to check the list of sources, the user would get prompted again. If you prefer to use a property or something or session variable, I'm fine doing that.

src/embezzler.ts
burningbman: Probably. This was just copy-pasted.

src/embezzler.ts
burningbman: No idea the difference, but we should definitely pick one to consistently use.

src/embezzler.ts
horrible-little-slime: I think a session variable is our best bet. You could make it part of globalOptions, or you could not make it part of globalOptions. Lots of options here!

My worry is that adding and removing it from the array is, well, a little arcane.

burningbman:
Closing due to better version.

PR #536, opened 2021-11-06
Be more careful about getting enough synth.

phulin:


src/embezzler.ts
```diff
@@ -316,13 +316,18 @@ export function embezzlerCount(): number {
...
+  const thumbRingMultiplier = have($item`mafia thumb ring`) ? 1.04 : 1;
   let turns;
   if (globalOptions.stopTurncount) turns = globalOptions.stopTurncount - myTurncount();
   else if (globalOptions.noBarf) turns = embezzlerCount();
-  else
-    turns =
-      (myAdventures() + (globalOptions.ascending && myInebriety() <= inebrietyLimit() ? 60 : 0)) *
-      (have($item`mafia thumb ring`) ? 1.04 : 1);
+  else {
+    turns = (myAdventures() + pantsgivingAdventures + nightcapAdventures) * thumbRingMultiplier;
```
horrible-little-slime, line 329: if we're rehauling this, we should account for magical sausages

src/embezzler.ts
```diff
@@ -316,13 +316,18 @@ export function embezzlerCount(): number {
 }
 
 export function estimatedTurns(): number {
+  // Assume roughly 2 fullness from pantsgiving and 8 adventures/fullness.
+  const pantsgivingAdventures = have($item`Pantsgiving`)
+    ? Math.max(0, 2 - get("_pantsgivingFullness")) * 8
+    : 0;
+  const nightcapAdventures = globalOptions.ascending && myInebriety() <= inebrietyLimit() ? 60 : 0;
+  const thumbRingMultiplier = have($item`mafia thumb ring`) ? 1.04 : 1;
```
horrible-little-slime, line 324: I feel the need to say it's actually 1/0.96 but I think we can call that negligible

phulin: yep

src/embezzler.ts
phulin: OK

horrible-little-slime (approved):

PR #538, opened 2021-11-07
use StrictMacro for garbo, rather than the classic Macro

horrible-little-slime:


horrible-little-slime:
Tested 11/8, no explosions

PR #540, opened 2021-11-07
start using libram's bjorn logic

horrible-little-slime:


horrible-little-slime:
Should bjorn.ts be merged into outfit.ts in light of this?

PR #543, opened 2021-11-09
Use feel superior if appropriate

CheeseyPickle:
Add feel superior to .kill(), so people can get pvp fights.
It costs 0 MP, so might as well.

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -417,7 +417,10 @@ export class Macro extends LibramMacro {
     return (
       this.externalIf(
         monsterManuelAvailable(),
-        Macro.if_("monsterhpbelow 20", Macro.trySkill($skill`Feel Superior`))
+        Macro.if_(
+          `monsterhpbelow ${Math.floor((100 + numericModifier("Monster Level")) / 5)}`,
+          Macro.trySkill($skill`Feel Superior`)
+        )
       )
         .externalIf(
           myClass() === $class`Sauceror` && have($skill`Curse of Weaksauce`),
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -416,9 +416,13 @@ export class Macro extends LibramMacro {
   kill(): Macro {
     return (
       this.externalIf(
-        myClass() === $class`Sauceror` && have($skill`Curse of Weaksauce`),
-        Macro.trySkill($skill`Curse of Weaksauce`)
+        monsterManuelAvailable(),
+        Macro.if_("monsterhpbelow 20", Macro.trySkill($skill`Feel Superior`))
       )
+        .externalIf(
+          myClass() === $class`Sauceror` && have($skill`Curse of Weaksauce`),
+          Macro.trySkill($skill`Curse of Weaksauce`)
+        )
         .externalIf(
           !(myClass() === $class`Sauceror` && have($skill`Curse of Weaksauce`)),
           Macro.while_("!pastround 20 && !hppercentbelow 25 && !missed 1", Macro.attack())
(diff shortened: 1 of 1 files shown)


horrible-little-slime (changes requested):
I have a few concerns about this that could very well be easily alleviated:

This does damage, and we want to use it to kill monsters. Does its position in the code account for the fact that, due to bjorn shenanigans, we often deliberately want to stasis? I'm pretty sure the answer here is "yes", because it appears to be just above the kill() section, but still.

monsterhpbelow talks about absolute hp, but Feel Superior's damage is done by percent hp. Is there a way for us to figure out what foe we expect to fight, calculate its hp adjusted for +ML, and figure out an appropriate damage threshold? There does not appear to be a monsterhppercentbelow predicate, otherwise I'd suggest that. I'm going to go ahead and shorten this question to "is there a better way to determine the hp cutoff"?

Feel Superior grants fights. You can't get fights if your stone is unbroken. Would we want to verify that a player's stone is broken before using Feel Superior? Obviously if you _never_ break your stone, you aren't wasting anything by blowing your Feel Superior uses, but I could imagine a player who runs garbo, then breaks their stone, then uses feel superior manually. It's a weird player, but still probably worth adding.

I'm marking this as 'request changes', but I'm very open to not doing any changes, so long as we address these things.

CheeseyPickle:
For your first concern, I'm pretty sure I put the code directly inside the .kill() Macro, so we should be good.

I put in fixes for the other 2.

horrible-little-slime (approved):

src/combat.ts
```diff
@@ -416,9 +417,16 @@ export class Macro extends LibramMacro {
   kill(): Macro {
     return (
       this.externalIf(
-        myClass() === $class`Sauceror` && have($skill`Curse of Weaksauce`),
-        Macro.trySkill($skill`Curse of Weaksauce`)
+        hippyStoneBroken() && monsterManuelAvailable(),
+        Macro.if_(
+          `monsterhpbelow ${Math.floor((100 + numericModifier("Monster Level")) / 5)}`,
```
s-k-z, line 422: I think it's important to add something like
```
`(monsterid 1758 || 1759 || monsterid 1760 ) && monsterhpbelow ${Math.floor((100 + numericModifier("Monster Level")) / 5)}`,
```
because the kill() function is used against more enemies than just barf mountain native monsters.

CheeseyPickle: Fair enough, added

horrible-little-slime:
I do not have appeared to have cast feel superior at all during this run of garbo, post-CS, during which time my stone has been broken the entire time.

CheeseyPickle:
> I do not have appeared to have cast feel superior at all during this run of garbo, post-CS, during which time my stone has been broken the entire time.

Hmm, it shouldn't trigger too often, only when you want to kill something directly. So it's never going to trigger if you do mafia pointer ring crits and the monster probably dies before kill() is called if you're stalling them. And even if you are killing them, we only do this when it has like <20 HP remaining, which is a hard range to hit.

This is really just for niche cases since we don't really value PVP fights, so meat always wins.

s-k-z:
If you value a pvp fight at the value of an adventure, then skipping a crit for 3 combats to obtain pvp fites should definitely be worth it.

CheeseyPickle:
Alright, I guess that's a reasonable assumption since the stone would have to be broken

s-k-z:
I think there needs to be some combat behavior to ensure pushing enemies down to <20% HP. Otherwise I don't see this ever being triggered unless there's a Vykea Couch or some other incidental damage sources active to help.

CheeseyPickle:
That's...mostly what I was counting on.
I don't really know of any good, accessible ways people can ping garbo monsters until their HP drops into that range, so I was hoping passive damage combined with MeatStasis would land them into it every so often, or at least 3 times a day.

horrible-little-slime:
I am happy with where this is; I wish we didn't have to override crit stuff, but I don't see a better way of handling that.

PR #550, opened 2021-11-12
get rid of smoke patch price cap

horrible-little-slime:


s-k-z:
If removing the price cap on transdermal smoke patches, why not all spleen adventure items like octolus oculus or antimatter wads?

s-k-z (approved):

PR #552, opened 2021-11-14
prevent all our various equips within the outfit functions

horrible-little-slime:
Right now, garbo doesn't actually pass any of our default preventEquips to the maximizer, because it passes a different array of preventEquips. This changes that.


horrible-little-slime:
Tested this last night, it does what I want it to. Libram will soon make this a nonissue, but it hasn't yet. I'm squooshin, baby.

PR #553, opened 2021-11-14
add preliminary support for the daylight shavings helmet

horrible-little-slime:
Mafia doesn't currently track the most recent buff, so we can't optimize based on that. But we can optimize not based on that!


horrible-little-slime:
Tested 11/18; I crashed a lot, but I believe that's due to computer issues, not this. Still, I'd like someone to test this _without_ their computer exploding before merging.

We're still waiting on mafia tracking for this, but I see no reason to not support it before then.

horrible-little-slime:
Did not explode; squooshing

PR #554, opened 2021-11-15
I guess we retrieve the pill keeper now

horrible-little-slime:


phulin:
This will unequip it if it’s equipped, also we don’t turn the closet auto satisfy on? Nor should we I think

horrible-little-slime:
This should probably not be equipped to fight an embezzler, but if we want
to add a check just for style reasons we can

horrible-little-slime:
I'm squashing this--we can roll it back once the mafia thing is patched, but I see no compelling reason not to per discussion in discord

PR #561, opened 2021-11-17
take advantage of libram's polymorphic if_

horrible-little-slime:


horrible-little-slime:
Tested 17 November, no apparent errors.

PR #567, opened 2021-11-22
Setup KGEs after prompting user to use wish (mostly stolen from BMan)

horrible-little-slime:
I had a slightly different approach I wanted to try for this, and thus didn't feel it was appropriate to directly piggyback atop #530. Still, it is fruit from the same tree, and a vast majority of this comes from bman's juicy brain.


burningbman:
Ran into an infinite loop due to the global option being set AFTER checking embezzler sources. Changed the meaning of the variable and shifted it around to get around this problem.

PR #568, opened 2021-11-23
warn about items we want to return

horrible-little-slime:


docrostov (commented):
this looks good to me; i like the change to make this print the items in order so it is easier to just copy/paste the list. i do not see any issue with squooshing this (but do not want to squoosh it myself because i live in eternal fear of messing up garbo)

PR #571, opened 2021-11-24
jsdocs for EmbezzlerFight

lewismd13:


src/embezzler.ts
```diff
@@ -46,7 +46,17 @@ export class EmbezzlerFight {
...
-
+  /**
+   * This is the class that creates all the different ways to fight embezzlers
+   * @classdesc Something goes here
+   * @prop {string} name The name of the fight so you know what you're doing
+   * @prop {boolean} available this should be a statement that checks the availability of the fight
+   * @prop {number} potential this should be a statement that checks the number of this type of embezzler fight available
+   * @prop {Location, Macro} run This runs the combat, using the macro and location provided, if any
+   * @prop {[]} requirements This is an array of requirements to do this fight
+   * @prop {boolean} [draggable=false] This tells the script if it should try to drag the embezzler to a different zone
```
horrible-little-slime, line 57:
```suggestion
   * @prop {boolean} [draggable=false] This tells the script if it is able the embezzler to a different zone
```
The able/should distinction is important; draggability is a property of the fight, not of the running itself. When we run a fight and provide a location input, we're telling it that it _should_ drag the embezzler to a different zone. Draggability tells us that a particular fight _can_ be taken to a different zone.

src/embezzler.ts
```diff
@@ -46,7 +46,17 @@ export class EmbezzlerFight {
   requirements: Requirement[];
   draggable: boolean;
   name: string;
-
+  /**
+   * This is the class that creates all the different ways to fight embezzlers
+   * @classdesc Something goes here
+   * @prop {string} name The name of the fight so you know what you're doing
```
horrible-little-slime, line 52:
```suggestion
   * @prop {string} name The name of the source of this fight, primarily used to identify special cases.
```

src/embezzler.ts
```diff
@@ -46,7 +46,17 @@ export class EmbezzlerFight {
...
+   * @prop {boolean} available this should be a statement that checks the availability of the fight
```
horrible-little-slime, line 53:
```suggestion
   * @prop {() => boolean} available Returns whether or not we can do this fight right now.
```
"checks the availability of" feels vague

src/embezzler.ts
```diff
@@ -46,7 +46,17 @@ export class EmbezzlerFight {
...
+   * @prop {number} potential this should be a statement that checks the number of this type of embezzler fight available
```
horrible-little-slime, line 54:
```suggestion
   * @prop {() => number} potential Returns the number of embezzlers we expect to be able to fight from this source.
```

We do not want to recycle the word "available", for Reasons

src/embezzler.ts
```diff
@@ -46,7 +46,17 @@ export class EmbezzlerFight {
...
+   * @prop {Location, Macro} run This runs the combat, using the macro and location provided, if any
```
horrible-little-slime, line 55:
```suggestion
   * @prop {(options: EmbezzlerFightOptions) => void} run This runs the combat, optionally using the provided location and macro. Location is used only by draggable fights.
```
Importantly, providing location and macro inputs does not actually force the fight to use those inputs.

src/embezzler.ts
```diff
@@ -46,7 +46,17 @@ export class EmbezzlerFight {
...
+   * @prop {[]} requirements This is an array of requirements to do this fight
```
horrible-little-slime, line 56:
```suggestion
   * @prop {Requirement[]} requirements This is an array of requirements to do this fight
```

PR #574, opened 2021-11-29
take advantage of autoattacks against drunks

horrible-little-slime:


src/fights.ts
```diff
@@ -573,17 +574,25 @@ class FreeRunFight extends FreeFight {
...
-  .abort();
+const pygmyMacro = () =>
+  Macro.if_($monster`pygmy bowler`, Macro.trySkill($skill`Snokebomb`).item($item`Louder Than Bomb`))
+    .if_(
+      $monster`pygmy orderlies`,
+      Macro.trySkill($skill`Feel Hatred`).item($item`divine champagne popper`)
+    )
+    .if_($monster`pygmy janitor`, Macro.item($item`tennis ball`))
+    .if_($monster`time-spinner prank`, Macro.basicCombat())
+    .externalIf(
```
burningbman, line 585: I don't understand the reasoning of this additional combat logic, but wanted to make sure it should be included in this change.

horrible-little-slime: I didn't touch the contexts of janitors, bowlers, or orderlies, or even time-pranks--i changed it to a macro-generating function and added the part referring to drunks specificlally

src/fights.ts
```diff
@@ -573,17 +574,25 @@ class FreeRunFight extends FreeFight {
...
+      SourceTerminal.have() || SongBoom.have(),
+      Macro.if_(
```
phulin, line 587: do you actually get multiple rounds of stuff to use?

horrible-little-slime: No, but the intent is to only add one of them to the macro. In retrospect I can make this much, much simpler. 1 sec.

PR #587, opened 2021-12-05
Extrovermectin Round 1

horrible-little-slime:


src/embezzler.ts
```diff
@@ -169,6 +180,127 @@ export const embezzlerSources = [
...
+        get("crystalBallMonster") !== $monster`Knob Goblin Embezzler` &&
+        !(get("_saberForceMonster") === $monster`crate` && get("_saberForceMonsterCount") > 0)
+      ) {
+        equip($slot`familiar`, $item`miniature crystal ball`);
+      }
+      const baseMacro = options.macro ?? embezzlerMacro();
+      const macro = Macro.if_($monster`crate`, Macro.skill($skill`Macrometeorite`)).step(baseMacro);
+      adventureMacro($location`Noob Cave`, macro);
+    }
+    //do we want to equip orb on these guys?
```
Thoth19, line 225: Theoretically, we can get an extra one by putting the orb on for the final greg fight. Whether or not that is EV positive depends on how many macros you have left when you have one greg left and how valueable your familiar equipment slot is. 
Might be worth to rule of thumb yes and then use other zones like barf to wipe the prediction if it is ever an embezzler as opposed to a crate. Until the end of our proposed chain when we do want one last Embez prediction

horrible-little-slime: I think we use orb on the final warren fight, not noob cave; are you saying we'd get another one?

I think I don't want to start interpolating barf turns, but I can add some sort of check to the available() function for the two replacement guys, that way we can do other embezzler fights to reset orb. The only thing is we need to find a zone other than noob cave for our wanderers to default to if we're going to do that--backups could be dangerous.

src/embezzler.ts
```diff
@@ -169,6 +180,127 @@ export const embezzlerSources = [
...
+    () =>
+      have($item`Time-Spinner`) &&
+      cliExecuteOutput("timespinner list monsters").includes(
+        $monster`Knob Goblin Embezzler`.name
+      ) &&
+      get("_timeSpinnerMinutesUsed") <= 7,
+    () =>
+      have($item`Time-Spinner`) &&
+      cliExecuteOutput("timespinner list monsters").includes($monster`Knob Goblin Embezzler`.name)
+        ? Math.min((10 - get("_timeSpinnerMinutesUsed")) / 3)
```
Thoth19, line 192: Two stupid questions:
1. In line 188, we split the arg of includes onto its own line, but we don't do that in 193. Is that the intended style? I didn't see a linter around here, but I didn't look for one either.
2. What are we doing with time spinner minutes? I understood that the first one is to check that we have enough minutes to do this. Is this to tell us how many times we are going to timespin the Embez?

src/embezzler.ts
```diff
@@ -169,6 +180,127 @@ export const embezzlerSources = [
...
+      saberCrateIfDesired();
+      if (
+        have($item`miniature crystal ball`) &&
+        get("crystalBallLocation") === $location`Noob Cave` &&
+        get("crystalBallMonster") === $monster`crate`
+      ) {
+        equip($slot`familiar`, $item`miniature crystal ball`);
+      }
+      const baseMacro = options.macro ?? embezzlerMacro();
+      const macro = Macro.if_($monster`crate`, Macro.skill($skill`CHEAT CODE: Replace Enemy`)).step(
```
Thoth19, line 244: Theoretically, do we want to extract etc. here before macro-ing or are we too afraid of winning the fight from random effects to risk the extra 10 meat?

src/extrovermectin.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export function doingExtrovermectin(): boolean {
+  return get("beGregariousCharges") > 0;
+}
+
+export function crateStrategy(): "Sniff" | "Saber" | "Orb" | null {
+  if (!doingExtrovermectin()) return null;
+  if (have($skill`Transcendent Olfaction`)) return "Sniff";
+  if (have($item`miniature crystal ball`)) return "Orb";
```
Thoth19, line 53: Do we also want a strategy that uses both olfaction and orb if possible, or is that too marginal to be useful? On round 1, the embez isn't in the queue yet, so it is 1/4 to hit your first greg there. It gets a lot less likely (1/13) with the orb and olfaction.
Unless I'm misunderstanding what this function does. The way I read this is that it is there to say how we want to ensure that we hit crates and not gregs so we can hit all 20 macro-ed Embezs

src/embezzler.ts
horrible-little-slime: 1. It's what the linter wants. Very weird choice, though! I assume it's the difference between the ternary vs continuing the condition
2. If you view it in VSCode you can see the documentation Manendra lovingly crafted for the EmbezzlerFight class; the first arrow function is whether the fight is available right now, the second tells us the number of fights we expect to be able to get from it.

src/embezzler.ts
horrible-little-slime: I am far too afraid to lose the fight from elemental-fairy-in-the-bjorn damage. We can do stuff with the existing stasis code but I'm declaring that out of scope of this PR.

src/extrovermectin.ts
horrible-little-slime: You're reading it right. Right now, our code uses orb regardless of whether we actually have olfaction. I hadn't planned out all of the code as I wrote this function.

horrible-little-slime: That being said, I'd like to keep it as is until I'm more certain of how much the rest of the code is going to change.

src/embezzler.ts
Thoth19: 1. OK.
2. Oh that's what is happening here. I thought we were  doing some pattern matxhing thing. The syntax reminds me of Haskel. So these are 0 argument lambdas that give us what we want. I assume this is the pattern that is generally used around here etc. Do we know that things like "cliExecuteOutput(timespinner list monsters)" is fast? Bc it looks like we are doing it twice and could theoretically cache the value. I haven't done any work on profiling this sort of thing, so it might turn out that this is actually more time efficient if storing takes time for some reason etc.

src/embezzler.ts
Thoth19: You are correct. I misunderstood the zone change. You can't get another extra because the Embez isn't a free fight.

I'm not sure I fully understand what you mean in the second section. But yes. The idea is that when we predict an Embez and we are not done doing macros, we don't want to hit the greg'd Embez. So we want to do some turn costing action to wipe the prediction. We would really like it to be another Embez bc we are already mega buffed. 
Is the problem with fighting wanderers in the noobcave that it might confuse us about which source we got the Embez from?

src/embezzler.ts
Thoth19: That makes complete sense and I totally support that (and agree that it is likely out of scope). I just wanted to make sure that this was something that was considered and rejected.
I thought that not beating a crate was easy, but based on some testing I did while overdrunk I was able to kill crates with 100ML too quickly with just regular attacks. And losing a turn here is p bad.

src/extrovermectin.ts
Thoth19: Oh I see. Since we are going to orb if possible anyway, Sniff means "sniff and if possible orb". Orb means "orb only" and if we have neither, we have to use saber.
That makes sense.

src/embezzler.ts
horrible-little-slime: I'm not sure if there's a specific thing about going to noob cave outside of this context that puts me on edge, but not defaulting to it would generally be safer for probably a lot of reasons.

src/embezzler.ts
horrible-little-slime: I have no idea if it's fast, and in fact it doesn't even work right now! Would love an alternative.
I'm concerned about caching because garbo is meant to be re-entrant and this section is trying to be as stateless as it can be. I want it to check every iteration whether it currently is able to spin an Embezzler.

src/fights.ts
```diff
@@ -260,10 +262,52 @@ function embezzlerSetup() {
...
+        forceEquip: $items`latte lovers member's mug`.filter((item) => have(item)),
+      })
+        .merge(run.requirement ? run.requirement : new Requirement([], {}))
+        .maximize();
+      useFamiliar(freeFightFamiliar());
+      if (run.prepare) run.prepare();
+      adventureMacro(
+        $location`Noob Cave`,
+        Macro.if_($monster`crate`, macro)
+          .if_($monster`time-spinner prank`, Macro.kill())
```
Rinn, line 291: Saw this time-spinner prank check few places, it would probably be better suited to be handed in the main() function of combat.ts similar to tentacles.

horrible-little-slime: Declaring that out of scope of this PR, but very good idea. That won't work for our autoattack macros--a silly fix for this would be to add it to the constructor function for our specialized Macro class, but also I'm _pretty_ sure we only use this sort of thing in adventureMacro rather than adventureMacroAuto anyways

src/diet.ts
```diff
@@ -501,6 +503,23 @@ export function runDiet(): void {
...
   if (embezzlers) {
+    if (
+      mySpleenUse() + 2 <= spleenLimit() &&
+      !get("beGregariousCharges") &&
+      !(
+        get("beGregariousFightsLeft") &&
+        get("beGregariousMonster") === $monster`Knob Goblin Embezzler`
+      )
+    ) {
+      const value = expectedGregs() * 25000;
```
Thoth19, line 514: Is this meant to be the expected meat from an Embez fight? Is this accounting for the diff between an Embez fight and a barf fight? I'm not sure I'm making 25k+7k from my Embezzlers. It is totally good back of the envelope math though. I'm just surprised we don't have an EmbezzlerMPA var if we have a regular mpa var

horrible-little-slime: 7k is pretty high an estimate for a barf-turn, but I guess it's start of the day. EmbezzlerMPA is very hard to estimate and varies a lot more over the course of a day than barf MPA. I consider this a decent temproary measure.

PR #588, opened 2021-12-06
Rework diet ordering so we do Refined Palate first.

phulin:


horrible-little-slime (approved):
My understanding of this could make it dangerous if we add salty mouth, but we'll burn that bridge when we come to it.

phulin:
Salty Mouth would be a helper, not a separate item

PR #591, opened 2021-12-08
Wanderer refactor

pstalcup:
rework wanderer.ts in support of future wanderer jobs (docbag, dinsey quests, fighting wart dinsey, daily currencies from sbb and glaciest, etc.)

Later we will add valuation to each wanderer type so it can prioritize appropriately

### DIFF
--- a/src/wanderer.ts
+++ b/src/wanderer.ts
@@ -1,5 +1,5 @@
 import { canAdv } from "canadv.ash";
-import { buy, craftType, myTurncount, retrieveItem, use } from "kolmafia";
+import { buy, craftType, myTurncount, print, retrieveItem, use } from "kolmafia";
 import {
   $effect,
   $item,
@@ -8,6 +8,7 @@ import {
   $skill,
   clamp,
   get,
+  getSaleValue,
   Guzzlr,
   have,
   questStep,
@@ -19,10 +20,6 @@ import { globalOptions, propertyManager } from "./lib";
 export enum draggableFight {
   BACKUP,
   WANDERER,
-}
-export enum wandererMode {
-  GUZZLR,
-  DEFAULT,
 }
 const WANDERER_PRICE_THRESHOLD = 10000;
 
@@ -53,9 +50,10 @@ export function digitizedMonstersRemaining(): number {
 }
 
 interface ZoneUnlocker {
-  zone: string;
+  zone: string | null;
+  location: Location | null;
   available: () => boolean;
-  unlocker: Item;
+  unlocker: Item | null;
   noInv?: boolean;
 }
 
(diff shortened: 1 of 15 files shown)


src/wanderer.ts
```diff
@@ -13,12 +14,17 @@ import {
...
 
 export enum draggableFight {
   BACKUP,
   WANDERER,
 }
+export enum wandererMode {
+  GUZZLR,
+  DEFAULT,
+}
+const WANDERER_PRICE_THRESHOLD = 10000;
```
horrible-little-slime, line 27: I would rename this to ZONE_UNLOCK_PRICE_THRESHOLD or something of that ilk.

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
 }
 
-interface ZonePotion {
+interface ZoneUnlocker {
   zone: string;
-  effect: Effect;
-  potion: Item;
+  available: () => boolean;
+  unlocker: Item;
+  noInv?: boolean;
```
horrible-little-slime, line 59: Should this be a property of the ZoneUnlocker itself?

The answer could very well be "maybe!"

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
+}
+
+function airportAvailable(element: "spooky" | "stench" | "hot" | "cold" | "sleaze"): boolean {
```
horrible-little-slime, line 62: I instinctively want this to use a $element, but that is probably wrong of me

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
-    ) {
-      Guzzlr.acceptPlatinum();
-    } else if (Guzzlr.canGold() && (!Guzzlr.haveFullGoldBonus() || Guzzlr.haveFullBronzeBonus())) {
-      Guzzlr.acceptGold();
-    } else {
-      Guzzlr.acceptBronze();
-    }
-  }
-}
+function testZoneAndUsePotionToAccess(zone: Location | null) {
```
horrible-little-slime, line 108: I get why you're doing this, but I feel like letting this take in a null input is a bad idea

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
-  }
-  if (!get("_hotAirportToday") && !get("hotAirportAlways")) {
-    forbiddenZones.push("That 70s Volcano");
-  }
-  if (!get("_coldAirportToday") && !get("coldAirportAlways")) {
-    forbiddenZones.push("The Glaciest");
-  }
-  if (!get("_sleazeAirportToday") && !get("sleazeAirportAlways")) {
-    forbiddenZones.push("Spring Break Beach");
+  // these zones have equip or effect requirements that we don't want to deal with
```
horrible-little-slime, line 111: This doesn't fully summarize it (caldera and oasis, for example), but that's probably fine

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
+  const skiplist = $locations`The Oasis, The Bubblin' Caldera, Barrrney's Barrr, The F'c'le, The Poop Deck, Belowdecks, 8-Bit Realm, Madness Bakery, The Secret Government Laboratory`;
+  if (skiplist.includes(zone)) {
```
horrible-little-slime, line 113: can we make this a 1-liner

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
+    return false;
   }
-
-  zonePotions.forEach((place) => {
-    if (guzzlZone.zone === place.zone && !have(place.effect)) {
-      if (!have(place.potion)) {
-        buy(1, place.potion, 10000);
+  zoneUnlockers.forEach((place) => {
```
horrible-little-slime, line 116: In retrospect, it feels very silly to have this be a for each. Make it a `for` loop and `break` if you find the parent zone. I know I made it this way, but that was past me.

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
-  const skiplist = $locations`The Oasis, The Bubblin' Caldera, Barrrney's Barrr, The F'c'le, The Poop Deck, Belowdecks, 8-Bit Realm, Madness Bakery, The Secret Government Laboratory`;
-  if (
-    forbiddenZones.includes(guzzlZone.zone) ||
-    skiplist.includes(guzzlZone) ||
-    guzzlZone.environment === "underwater" ||
-    !canAdv(guzzlZone, false) ||
-    (guzzlZone === $location`The Upper Chamber` && questStep("questL11Pyramid") === -1) // (hopefully) temporary fix for canadv bug that results in infinite loop
-  ) {
+  const canAdvHack = zone === $location`The Upper Chamber` && questStep("questL11Pyramid") === -1; // (hopefully) temporary fix for canadv bug that results in infinite loop
+  if (zone.environment === "underwater" || canAdvHack || !canAdv(zone, false)) {
```
horrible-little-slime, line 127: again, pretty sure I made this, but can't we just return `!(zone.environment === "underwater" || ...)`? Rather than `if (condition) return false; else return true;` couldn't we just do `return !condition`? is there a stylistic reason to choose this over that?

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
     return false;
   } else {
     return true;
   }
 }
 
-function testZoneForBackups(location: Location): boolean {
+function testZoneForBackups(location: Location | null): boolean {
+  if (!location) return false;
```
horrible-little-slime, line 135: As above, not sure if letting this take a null input is the solution we're looking for

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
+    } else if (Guzzlr.canGold() && (!Guzzlr.haveFullGoldBonus() || Guzzlr.haveFullBronzeBonus())) {
+      Guzzlr.acceptGold();
+    } else {
+      Guzzlr.acceptBronze();
+    }
   }
 
-  const choicesToSet = unsupportedChoices.get(guzzlZone);
-  if (choicesToSet) propertyManager.setChoices(choicesToSet);
+  const location = Guzzlr.getLocation()!;
```
horrible-little-slime, line 186:
```suggestion
  const location = Guzzlr.getLocation();
  if (!location) return false;
```

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
-      print("It's time to get buttery", "purple");
-      cliExecute("make buttery boy");
-    }
-  } else {
-    const guzzlrBooze = Guzzlr.getBooze();
-    if (!guzzlrBooze) {
-      return defaultLocation;
-    } else if (!have(guzzlrBooze)) {
-      print("just picking up some booze before we roll", "blue");
+  // error state - accepted a quest, but mafia says the quest location or booze is null
```
horrible-little-slime, line 192: Consider doing `refresh quests` if we end up in this state. Ultimately we'll need _some_ sort of catch, just for typescript's sake, but we might as well

src/wanderer.ts
```diff
@@ -46,165 +52,200 @@ export function digitizedMonstersRemaining(): number {
...
+  if (Guzzlr.have() && prepareGuzzlr() && wandererCheck(Guzzlr.getLocation())) {
+    return wandererMode.GUZZLR;
+  }
+  return wandererMode.DEFAULT;
+}
+
+function wandererLocation(type: draggableFight, mode: wandererMode): Location {
+  switch (mode) {
+    case wandererMode.GUZZLR:
+      return Guzzlr.getLocation()!;
```
horrible-little-slime, line 233: Typescript and github both hate it when you do this. I get it, but still!

horrible-little-slime (changes requested):
This is looking promising so far, and I'll look at it again in a few days on vscode rather than github.

src/wanderer.ts
pstalcup: this is not necessarily a function of hte zone unlocker, but rather the price of it. If the price exceeds ZONE_UNLOCK_PRICE_THRESHOLD by default, but a user has one in their inventory we do not want ot use it

src/wanderer.ts
pstalcup: $element includes lime and supercold, which are not valid airports. I can look and see if there is a way to exclude only specific elements, or modify libram to classify those elements different or provide a list to them. PrimaryElement maybe?

src/wanderer.ts
pstalcup: the "effect" requirement for caldera and oasis is covered in lava and ultrahydrated, respecitvely

src/wanderer.ts
horrible-little-slime: yeah that sucks then

horrible-little-slime (commented):
Approved pending testing by at least two people for at least two days.

PR #593, opened 2021-12-08
Do not use space pills if you don't have it and the skip is set

pstalcup:


phulin:
If this is still broken, the fix belongs in libram, not here

PR #604, opened 2021-12-13
don't do sausages if we're expecting a kramco

horrible-little-slime:


lewismd13 (approved):
So this just skips going for a kramco if we are in a wanderer window? With the caveat that I have not spent any time figuring out how counters work, it looks like it should work.

horrible-little-slime:
That is the goal, yes

PR #607, opened 2021-12-15
Horse

pstalcup:


src/embezzler.ts
```diff
@@ -261,14 +277,15 @@ export const embezzlerSources = [
...
+    [new Requirement([], { forceEquip: $items`Powerful Glove` })],
+    { swaps: true }
   ),
   new EmbezzlerFight(
     "Be Gregarious",
     () =>
       retrieveItem(1, $item`human musk`) &&
       get("beGregariousMonster") === $monster`Knob Goblin Embezzler` &&
-      get("beGregariousFightsLeft") > 0,
+      get("beGregariousFightsLeft") > (have($item`miniature crystal ball`) ? 1 : 0),
```
horrible-little-slime, line 288: I know what this is intended to do, but remove it; we need to rehaul orb code to match the new mafia system

src/extrovermectin.ts
```diff
@@ -44,7 +44,7 @@ export function expectedGregs(): number {
 }
 
 export function doingExtrovermectin(): boolean {
-  return get("beGregariousCharges") > 0;
+  return true; // get("beGregariousCharges") > 0;
```
horrible-little-slime, line 47: ...

src/lib.ts
```diff
@@ -68,13 +66,15 @@ export const globalOptions: {
...
   wishAnswer: boolean;
+  pillKeeperUses: number;
 } = {
   stopTurncount: null,
   ascending: false,
   saveTurns: 0,
   noBarf: false,
   askedAboutWish: false,
   wishAnswer: false,
+  pillKeeperUses: 0,
```
horrible-little-slime, line 77: I don't understand the utility of tracking this internally. Can you explain?

src/lib.ts
```diff
@@ -517,17 +517,12 @@ function coldMedicineCabinet(): void {
...
-    acquire(1, $item`jumping horseradish`, 5.5 * get("valueOfAdventure"));
-    if (!eat(1, $item`jumping horseradish`)) throw "Failed to eat safely";
+    fillPantsgiving();
   }
 }
 
-export function postCombatActions(): void {
-  horseradish();
+export function postCombatActions(skipDiet = false): void {
+  if (!skipDiet) horseradish();
```
horrible-little-slime, line 525: I'd rather handle this in the horseradish function. also I'm not sure why we have a separate `fillPantsgiving()` function from the `horseradish()` function--why not excise horseradish entirely and just use `fillPantsgiving`?

src/extrovermectin.ts
pstalcup: Some of the free runs compete with extro for resources, and so we have to do the free runs after we compute diet. We want to compute the diet after we do most free fights so that way we can safely compute what our Meat Drop modifier is. 
I marked this one as true simply because I was under the assumption that everyone testing this would always be doing extro, since that is a significantly better usage of resources

src/lib.ts
pstalcup: I added pillkeeper to the diet, so this is to mark how many "spleen" uses of pillkeeper the diet has budgeted for. Those pill uses will come later in the run, so we have to share this between diet and embezzler.ts somehow, and I felt this was the easiest way to accomplish that.

src/embezzler.ts
pstalcup: Yes, I did this as a short term fix because I'm on old mafia

src/lib.ts
pstalcup: this is a good suggestion - it should just call the diet's "pantsgiving" directly instead.

PR #608, opened 2021-12-15
Rework Potions

pstalcup:
-allow for overriding of effects and duration
-move useAsValuable outside of the class to keep that behavior private to the module
-create static convenience versions of the useful functions
-add in "pocket wish" potions for valuable meat drop effects (this will never be worth it without PYEC)

### DIFF
--- a/src/potions.ts
+++ b/src/potions.ts
@@ -210,8 +210,9 @@ export class Potion {
       : (quantity: number) => quantity;
 
     // compute the value of covering embezzlers
-    const embezzlerQuantity = this.usesToCover(startingTurns, false);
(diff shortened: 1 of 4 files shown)


src/potions.ts
```diff
@@ -85,137 +124,213 @@ export class Potion {
...
+
+  const total = amountsAcquired.reduce((total, amount) => total + amount);
+  if (total > 0) {
+    print(`Using ${total} ${potion.potion.plural}`);
+    potion.use(total);
+  }
+  return total;
+}
+
+const farmingPotiosn = [
```
phulin, line 286:
```suggestion
const farmingPotions = [
```

horrible-little-slime (approved):

PR #609, opened 2021-12-15
Isolate postCombat to its own file to handle its dependecies

pstalcup:


src/post.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (mallPrice($item`fudge spork`) < 3 * get("valueOfAdventure") && !get("_fudgeSporkUsed"))
+      eat(1, $item`fudge spork`);
+    MayoClinic.setMayoMinder(MayoClinic.Mayo.zapine, 1);
+    acquire(1, $item`Special Seasoning`, get("valueOfAdventure"));
+    acquire(1, $item`jumping horseradish`, 5.5 * get("valueOfAdventure"));
+    if (!eat(1, $item`jumping horseradish`)) throw "Failed to eat safely";
+  }
+}
+
+export function postCombatActions(skipDiet = false): void {
```
horrible-little-slime, line 54: may be worth exporting this as a default, but i'll leave that to future us

PR #612, opened 2021-12-18
use leftover sabers on crates regardless of olfaction status

horrible-little-slime:


s-k-z:
Won't this always use all the force uses up on crates and never leave any for pills?

horrible-little-slime:
We do pills first

s-k-z (approved):

PR #613, opened 2021-12-18
don't react as poorly to mistimed digitizes

horrible-little-slime:
kevinzana (#591276) had an issue where garbo banished a gregged embezzler while trying to initialize the digitize counter, despite having already started it. Adding some wiggle room to the counters (extending into -3), as well as a macro that won't banish the embezzler.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -322,7 +322,7 @@ function startWandererCounter() {
         Macro.if_($monster`Knob Goblin Embezzler`, embezzlerMacro()).step(run.macro)
       );
     } while (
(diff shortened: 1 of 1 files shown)


s-k-z (commented):
Is this window large enough inside a chain of force friends?

pstalcup (approved):

s-k-z:
Can putting on the meat outfit preclude you from being able to free run?

horrible-little-slime:
I don't think so; we pass the requirement of the run

s-k-z (approved):

PR #614, opened 2021-12-18
doingExtrovermectin() should check greg fights remaining and skill charge

s-k-z:


horrible-little-slime (changes requested):
I think we need to change the logic in `embezzlerSetup` to be `get("beGregariousCharges") > 0 && get("beGregariousFightsLeft") === 0` and then this should be ready 2 roll

s-k-z:
> I think we need to change the logic in `embezzlerSetup` to be `get("beGregariousCharges") > 0 && get("beGregariousFightsLeft") === 0` and then this should be ready 2 roll

If you do that then when re-entering code you will use up resources devoted to greg fights.

horrible-little-slime (approved):

PR #615, opened 2021-12-18
Only fight sandworms when it makes sense.

phulin:
Also fixes the bug where we don't equip champagne.


src/outfit.ts
```diff
@@ -85,11 +87,8 @@ export function freeFightOutfit(requirements: Requirement[] = []): void {
   );
 
   const finalRequirement = new Requirement(parameters, {
-    forceEquip: forceEquip,
-    preventEquip: [
-      ...preventEquip,
-      ...$items`broken champagne bottle, Spooky Putty snake, Spooky Putty mitre, Spooky Putty leotard, Spooky Putty ball, papier-mitre, smoke ball`,
-    ],
+    forceEquip,
+    ...(preventEquip.length > 0 ? { preventEquip } : {}),
```
horrible-little-slime, line 91: We use preventEquip for the bjorn/crown much of the time--wouldn't this perpetually override our default settings?

phulin: oh okay, I'll have to scrub the code for that

src/fights.ts
```diff
@@ -348,10 +347,11 @@ export function dailyFights(): void {
...
         const startLectures = get("_pocketProfessorLectures");
         useFamiliar($familiar`Pocket Professor`);
-        meatOutfit(true, [
-          ...fightSource.requirements,
-          new Requirement([], { forceEquip: $items`Pocket Professor memory chip` }),
-        ]);
+        const requirement = Requirement.merge(fightSource.requirements);
+        const forceEquip = requirement.maximizeOptions.forceEquip ?? [];
+        forceEquip.push($item`Pocket Professor memory chip`);
+        requirement.maximizeOptions.forceEquip = forceEquip;
```
horrible-little-slime, line 353: Does this work? `maximizeOptions` is a getter function, not the property itself.

src/outfit.ts
```diff
@@ -58,15 +57,12 @@ const bestAdventuresFromPants =
...
-  const compiledRequirements = requirement ?? new Requirement([], {});
-  const compiledOptions = compiledRequirements.maximizeOptions;
-  const compiledParameters = compiledRequirements.maximizeParameters;
 
-  const forceEquip = compiledOptions.forceEquip ?? [];
-  const bonusEquip = compiledOptions.bonusEquip ?? new Map<Item, number>();
-  const preventEquip = compiledOptions.preventEquip ?? [];
-  const preventSlot = compiledOptions.preventSlot ?? [];
-  const parameters = compiledParameters;
+  const parameters = requirement?.maximizeParameters ?? [];
```
horrible-little-slime, line 61: We should make this a function that we add to libram; something that essentially explodes a Requirement object into a bunch of consts.

horrible-little-slime (commented):
It feels like the solution here might be a rehaul of the Requirement class in libram, or an expansion of it.

src/fights.ts
```diff
@@ -1474,6 +1479,13 @@ const freeRunFightSources = [
   ),
 ];
 
+function sandwormRequirement() {
+  return new Requirement(
+    ["100 Item Drop"],
+    get("garbageChampagneCharge") > 0 ? { forceEquip: $items`broken champagne bottle` } : {}
```
horrible-little-slime, line 1485: does anything verify that you own the champagne bottle?

phulin: done

src/outfit.ts
phulin: after more thought i think maybe we should do something like your requirement builder thing

src/fights.ts
phulin: yeah

phulin:
I don't think this is the cleanest solution at this point but it's an improvement. I suggest we merge and continue to think about improving `Requirement`. Thoughts?

PR #617, opened 2021-12-19
Bonus drop items update

Rinn:
- mafia thumb ring based on valueOfAdventure
- lucky gold ring based on what currencies could drop (currently over estimated if the volcoino hasn't dropped yet). Uses day passes, karma shawarma, & good dread consumables for values. Hobo nickel is set at 100, but garbo already closets these so w/e
- Update Mr. Cheeng's spectacles to 400 based on discussion in discord. We speculate the drop items are any pvpable potion that will never be banned by standard restrictions, which seems to check out. Could be dynamically generated later if we can get the list of standard items for the current year
- Don't add bonuses to maximizer string if the equipment is unavailable

### DIFF
--- a/src/outfit.ts
+++ b/src/outfit.ts
@@ -355,7 +355,7 @@ function mafiaThumbRing(equipMode: BonusEquipMode) {
 
 function luckyGoldRing(equipMode: BonusEquipMode) {
   // Ignore for DMT, assuming mafia might get confused about the volcoino drop by the weird combats
-  if (!have($item`lucky gold ring`) || ["dmt"].some((mode) => mode === equipMode)) {
+  if (!have($item`lucky gold ring`) || equipMode === "dmt") {
     return new Map<Item, number>([]);
   }
 
@@ -398,7 +398,7 @@ function luckyGoldRing(equipMode: BonusEquipMode) {
   return new Map<Item, number>([
     [
       $item`lucky gold ring`,
-      dropValues.reduce((previous, current) => previous + current) / dropValues.length / 10,
+      sumNumbers(dropValues) / dropValues.length / 10,
     ],
   ]);
 }
--- a/src/outfit.ts
+++ b/src/outfit.ts
@@ -354,7 +354,8 @@ function mafiaThumbRing(equipMode: BonusEquipMode) {
 }
 
 function luckyGoldRing(equipMode: BonusEquipMode) {
-  if (!have($item`lucky gold ring`)) {
+  // Ignore for DMT, assuming mafia might get confused about the volcoino drop by the weird combats
+  if (!have($item`lucky gold ring`) || ["dmt"].some((mode) => mode === equipMode)) {
     return new Map<Item, number>([]);
   }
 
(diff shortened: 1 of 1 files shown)


src/outfit.ts
```diff
@@ -340,6 +342,108 @@ function cheeses(embezzlerUp: boolean) {
...
+  }
+
+  return new Map<Item, number>([
+    [$item`mafia thumb ring`, (1 / 0.96 - 1) * get("valueOfAdventure")],
+  ]);
+}
+
+function luckyGoldRing(equipMode: BonusEquipMode) {
+  // Ignore for DMT, assuming mafia might get confused about the volcoino drop by the weird combats
+  if (!have($item`lucky gold ring`) || ["dmt"].some((mode) => mode === equipMode)) {
```
horrible-little-slime, line 358:
```suggestion
  if (!have($item`lucky gold ring`) || equipMode === "dmt") {
```

I'm assuming this is a remnant from `isFree`, but it's still confusing to read even if it's paralleling something else

src/outfit.ts
```diff
@@ -340,6 +342,108 @@ function cheeses(embezzlerUp: boolean) {
...
+        : 0,
+      get("frAlways") || get("_frToday") ? getSaleValue($item`FantasyRealm guest pass`) / 350 : 0,
+    ].filter((value) => value > 0),
+  ];
+
+  // Items drop every ~10 turns
+  return new Map<Item, number>([
+    [
+      $item`lucky gold ring`,
+      dropValues.reduce((previous, current) => previous + current) / dropValues.length / 10,
```
horrible-little-slime, line 401:
```suggestion
      sumNumbers(dropValues) / dropValues.length / 10,
```

importing `sumNumbers` from Libram

src/outfit.ts
```diff
@@ -340,6 +342,108 @@ function cheeses(embezzlerUp: boolean) {
...
+      get("spookyAirportAlways") || get("_spookyAirportToday")
+        ? Math.max(
+            getSaleValue($item`one-day ticket to Conspiracy Island`) / 100,
+            getSaleValue($item`karma shawarma`) / 7
+          )
+        : 0,
+      get("stenchAirportAlways") || get("_stenchAirportToday")
+        ? getSaleValue($item`one-day ticket to Dinseylandfill`) / 20
+        : 0,
+      (get("hotAirportAlways") || get("_hotAirportToday")) && !get("_luckyGoldRingVolcoino")
```
horrible-little-slime, line 387: not important but once the wanderer revamp hits, we should replace all these with `airportAccessible` or whatever the name of that function is

src/outfit.ts
```diff
@@ -340,6 +342,108 @@ function cheeses(embezzlerUp: boolean) {
...
+  // Overestimating until it drops is probably fine, don't @ me
+  const dropValues = [
+    100, // 80 - 120 meat
+    ...[
+      itemAmount($item`hobo nickel`) > 0 ? 100 : 0, // This should be closeted
+      itemAmount($item`sand dollar`) > 0 ? getSaleValue($item`sand dollar`) : 0, // This should be closeted
+      itemAmount($item`Freddy Kruegerand`) > 0
+        ? Math.max(
+            getSaleValue($item`bottle of Bloodweiser`) / 200,
+            getSaleValue($item`electric Kool-Aid`) / 200
```
phulin, line 372: keys here too, i think?

phulin: also, sometimes bloodweiser/kool-aid mallPrice gets a little off but not sure if we can do anything about that

PR #618, opened 2021-12-20
move protonpack and cloake to bonusequips

horrible-little-slime:


horrible-little-slime:
Closed; this fucks with bjornalike handling and we need a better paradigm but don't yet have one

PR #619, opened 2021-12-20
Rework Diet

pstalcup:
This reworks the diet to use knapsack for spleen


pstalcup:
This PR needs a libram update to work correctly - it will fail builds until then

horrible-little-slime:
> This PR needs a libram update to work correctly - it will fail builds until then

Sounds like this should be a Draft then

pstalcup:
I am not good with computers how did this get here

src/diet.ts
```diff
@@ -477,76 +501,14 @@ function itemPriority(menuItems: MenuItem[]) {
...
-  // Sort descending by item priority.
-  diet.sort(([x], [y]) => -(itemPriority(x) - itemPriority(y)));
-
+export function consumeDiet(diet: [MenuItem[], number][]): void {
   print();
   print("===== PLANNED DIET =====");
   printDiet(diet);
   print();
 
+  diet.sort(([x], [y]) => -(itemPriority(x) - itemPriority(y)));
```
horrible-little-slime, line 510:
```suggestion
  diet.sort(([a], [b]) => itemPriority(b) - itemPriority(a));
```

This is more consistent with other writing we see in garbo

src/diet.ts
```diff
@@ -386,80 +290,200 @@ export function computeDiet(): [MenuItem[], number][] {
...
+    ...copiers(),
+
+    // FOOD POTIONS
+    ...potion($item`jumping horseradish`),
+    ...potion($item`jumping horseradish`),
+    ...potion($item`tempura cauliflower`),
+    ...potion($item`sea truffle`),
+    ...potion($item`tempura broccoli`),
+    ...limitedPotion(
+      $item`Dinsey food-cone`,
```
phulin, line 382: For some reason my garbo isn't getting the food-cones automatically.

phulin: Oh, looks like it's because it hasn't opened the zone yet.

horrible-little-slime: This means that we need to account for nobarf when we do that, too

src/potions.ts
```diff
@@ -85,137 +124,217 @@ export class Potion {
...
+          quantity: limitFunction(1),
+          value: this.gross(0, ascendingTurns),
+        });
       }
     }
 
-    if (quantityToUse > 0) {
-      if (doubleDuration) quantityToUse = 1;
-      use(quantityToUse, this.potion);
+    print(`POTION: Breakpoints for ${this.potion} (starting turns ${startingTurns}):`);
```
phulin, line 255: Can you reduce the quantity of debug messages? Fine for debugging but I think we want less text - and more human-readable text - in the final version.

src/diet.ts
```diff
@@ -386,80 +290,200 @@ export function computeDiet(): [MenuItem[], number][] {
...
+    return limitedPotion(potion, undefined, options);
+  }
+
+  return [
+    ...baseMenu,
+    ...copiers(),
+
+    // FOOD POTIONS
+    ...potion($item`jumping horseradish`),
+    ...potion($item`jumping horseradish`),
```
phulin, line 377: I don't think this should be included twice.

phulin:
If you fix the linter errors this is good to go, right?

phulin:
Also needs a libram update now

pstalcup:
I'll push my latest version through tonight, we can do a libram release and it will be good to go

phulin:
libram release is already done :)

pstalcup:
Closing to move to a new PR to make things easier to reason about

PR #627, opened 2021-12-21
update thumb ring logic

horrible-little-slime:


src/dropsgear.ts
```diff
@@ -0,0 +1,376 @@
...
+
+const bestAdventuresFromPants =
+  Item.all()
+    .filter(
+      (item) =>
+        toSlot(item) === $slot`pants` && have(item) && numericModifier(item, "Adventures") > 0
+    )
+    .map((pants) => numericModifier(pants, "Adventures"))
+    .sort((a, b) => b - a)[0] || 0;
+const haveSomeCheese = getFoldGroup($item`stinky cheese diaper`).some((item) => have(item));
```
pstalcup, line 138: This should probably not be only computed once per garbo run, but I think it is fine to do statically for now

src/dropsgear.ts
```diff
@@ -0,0 +1,376 @@
...
+  return new Map<Item, number>([[$item`Daylight Shavings Helmet`, bonusValue]]);
+}
+
+/**
+ * Calculates whether we expect to be wearing the thumb ring for most of the farming day.
+ * This is used in functions that leverage projected turns; for instance, calculating the
+ * number of turns of sweet synthesis required in our diet calcs or potion costs.
+ * @returns boolean of whether we expect to be wearing the thumb ring for much of the day
+ */
+export function usingThumbRing(): boolean {
```
pstalcup, line 350: this function should reflect if we are in noBarf mode or not

pstalcup (approved):

src/dropsgear.ts
horrible-little-slime: in nobarf, we don't care about long-term thumb ring use because our expectedTurns() is just embezzlerCount()

src/dropsgear.ts
horrible-little-slime: We don't take stinky cheese from the stash. unless you mean valuing this, in which case my answer is 'shrug'. We want to cache values because otherwise we get a bunch of tiny microadjustments from turn-to-turn which ruins our equipment caching.

PR #629, opened 2021-12-21
Do not feed robort a bloody nora if uneeded

Thoth19:
From conversation here:
https://discord.com/channels/466605739838930955/864070665615179786/922699965392900176
If garbo for some reason doesn't want to go to the Sea, the robort will
end up with a wasted drink. If we can predict that we won't use it, then
we should skip it.
Note that this logic will get hairy if we want to go the Sea for more
complex things in the future.

Note that I haven't compiled/tested this bc I do have a fishy pipe.

### DIFF
--- a/src/dailies.ts
+++ b/src/dailies.ts
@@ -201,7 +201,15 @@ function configureGear(): void {
 
 function prepFamiliars(): void {
   if (have($familiar`Robortender`)) {
-    for (const drink of $items`Newark, drive-by shooting, Feliz Navidad, single entendre, Bloody Nora`) {
+    // Omit the nora if we aren't going to go to the Sea. If we have another reason to go to
+    // the Sea, this logic will get a bit more hairy.
+    if (((itemAmount($item`fishy pipe`) === 0) || (get(_fishyPipeUsed) != false)) && 
+      (!have($effect`Fishy`))) {
+      robortDrinksList = `Newark, drive-by shooting, Feliz Navidad, single entendre`
+    } else {
+      robortDrinksList = `Newark, drive-by shooting, Feliz Navidad, single entendre, Bloody Nora`
+    }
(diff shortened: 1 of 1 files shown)


src/dailies.ts
```diff
@@ -201,7 +201,15 @@ function configureGear(): void {
 
 function prepFamiliars(): void {
   if (have($familiar`Robortender`)) {
-    for (const drink of $items`Newark, drive-by shooting, Feliz Navidad, single entendre, Bloody Nora`) {
+    // Omit the nora if we aren't going to go to the Sea. If we have another reason to go to
+    // the Sea, this logic will get a bit more hairy.
+    if (((itemAmount($item`fishy pipe`) === 0) || (get(_fishyPipeUsed) != false)) &&
```
horrible-little-slime, line 206:
```suggestion
    if ((!have($item`fishy pipe` || !get(_fishyPipeUsed)) && 
```

src/dailies.ts
```diff
@@ -201,7 +201,15 @@ function configureGear(): void {
 
 function prepFamiliars(): void {
   if (have($familiar`Robortender`)) {
-    for (const drink of $items`Newark, drive-by shooting, Feliz Navidad, single entendre, Bloody Nora`) {
+    // Omit the nora if we aren't going to go to the Sea. If we have another reason to go to
+    // the Sea, this logic will get a bit more hairy.
+    if (((itemAmount($item`fishy pipe`) === 0) || (get(_fishyPipeUsed) != false)) && 
+      (!have($effect`Fishy`))) {
```
horrible-little-slime, line 207:
```suggestion
      !have($effect`Fishy`)) {
```

horrible-little-slime:
The logic of this codeblock is going to be revisited and rehauled significantly.

PR #631, opened 2021-12-21
Refactor itemAmount to have where appropriate

Thoth19:
https://discord.com/channels/466605739838930955/864070665615179786/922713019690131476

There are other uses of itemAmount, but they are for actual numbers rather than checking for 0 or non0

### DIFF
--- a/src/dailies.ts
+++ b/src/dailies.ts
@@ -204,7 +204,7 @@ function prepFamiliars(): void {
     for (const drink of $items`Newark, drive-by shooting, Feliz Navidad, single entendre, Bloody Nora`) {
       if (get("_roboDrinks").includes(drink.name)) continue;
       useFamiliar($familiar`Robortender`);
-      if (itemAmount(drink) === 0) retrieveItem(1, drink);
+      if (!have(drink)) retrieveItem(1, drink);
       print(`Feeding robortender ${drink}.`, "blue");
       visitUrl(`inventory.php?action=robooze&which=1&whichitem=${toInt(drink)}`);
     }
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -658,7 +658,7 @@ const freeFightSources = [
   new FreeFight(
     () => get("questL02Larva") !== "unstarted" && !get("_eldritchTentacleFought"),
     () => {
-      const haveEldritchEssence = have($item`eldritch essence`);
+      const haveEldritchEssence = itemAmount($item`eldritch essence`) !== 0;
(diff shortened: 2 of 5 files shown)


src/fights.ts
```diff
@@ -225,7 +225,7 @@ function embezzlerSetup() {
     });
   }
 
-  if (!get("_enamorangs") && !itemAmount($item`LOV Enamorang`) && averageEmbezzlerNet > 20000) {
+  if (!get("_enamorangs") && !have($item`LOV Enamorang`) && averageEmbezzlerNet > 20000) {
```
horrible-little-slime, line 228: we actually want to use itemAmount here; itemAmount only checks the inventory, while have will sometimes check the closet

src/fights.ts
```diff
@@ -658,7 +658,7 @@ const freeFightSources = [
   new FreeFight(
     () => get("questL02Larva") !== "unstarted" && !get("_eldritchTentacleFought"),
     () => {
-      const haveEldritchEssence = itemAmount($item`eldritch essence`) !== 0;
+      const haveEldritchEssence = have($item`eldritch essence`);
```
horrible-little-slime, line 661: We only care about inventory for this one, so we want to keep itemAmount

src/outfit.ts
```diff
@@ -365,9 +364,9 @@ function luckyGoldRing(equipMode: BonusEquipMode) {
   const dropValues = [
     100, // 80 - 120 meat
     ...[
-      itemAmount($item`hobo nickel`) > 0 ? 100 : 0, // This should be closeted
-      itemAmount($item`sand dollar`) > 0 ? getSaleValue($item`sand dollar`) : 0, // This should be closeted
-      itemAmount($item`Freddy Kruegerand`) > 0
+      have($item`hobo nickel`) ? 100 : 0, // This should be closeted
```
horrible-little-slime, line 367: all three of these should be itemAmount; it's about what's in the inventory

src/dailies.ts
```diff
@@ -204,7 +204,7 @@ function prepFamiliars(): void {
     for (const drink of $items`Newark, drive-by shooting, Feliz Navidad, single entendre, Bloody Nora`) {
       if (get("_roboDrinks").includes(drink.name)) continue;
       useFamiliar($familiar`Robortender`);
-      if (itemAmount(drink) === 0) retrieveItem(1, drink);
+      if (!have(drink)) retrieveItem(1, drink);
```
horrible-little-slime, line 207: We should actually just remove the check here; retrieveItem does nothing if you have it in inventory, and we _do_ want it in inventory

PR #633, opened 2021-12-21
Add carnivorous potted plant for gingerbread city and the menagerie

s-k-z:
Free kills from the carnivorous potted plant are even better than free runs here so we get extra value out of equipping it in these locations.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -1335,6 +1335,14 @@ const freeRunFightSources = [
       ) {
         set("_gingerbreadCityTurns", 1 + get("_gingerbreadCityTurns"));
       }
+    },
+    {
+      requirements: () => [
+        new Requirement([], {
+          // eslint-disable-next-line libram/verify-constants
+          forceEquip: $items`carnivorous potted plant`.filter((item) => have(item)),
+        }),
+      ],
     }
   ),
   new FreeFight(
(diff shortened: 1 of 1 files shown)


src/fights.ts
```diff
@@ -1335,6 +1335,14 @@ const freeRunFightSources = [
       ) {
         set("_gingerbreadCityTurns", 1 + get("_gingerbreadCityTurns"));
       }
+    },
+    {
+      requirements: () => [
+        new Requirement([], {
+          // eslint-disable-next-line libram/verify-constants
+          forceEquip: $items`carnivorous potted plant`.filter((item) => have(item)),
```
horrible-little-slime, line 1343: It looks like the plugin just upgraded; you may not need this

horrible-little-slime (changes requested):
we need to update our macros to take advantage of this or else it's not actually reaping any benefits.

src/fights.ts
s-k-z: Sadly not, upgraded around 11 hours ago and KoLMafia only added the data around 2 hours ago.

s-k-z:
> we need to update our macros to take advantage of this or else it's not actually reaping any benefits.

I don't think so. It should end combat immediately and just advance delay and grant items/meat.

PR #634, opened 2021-12-21
Add disintegrating spiky collar to list of polar vortex options

s-k-z:


src/fights.ts
```diff
@@ -1775,41 +1778,70 @@ const fireExtinguishZones = [
...
-  )[0];
+  const targets = fireExtinguishZones.filter((zone) => zone.isOpen() && !isBanished(zone.monster));
+  const vorticesAvail = Math.floor(get("_fireExtinguisherCharge") / 10);
+  bestFireExtinguisherZoneCached = targets.sort((a, b) => {
+    const value = (zone: fireExtinguisherZone): number => {
+      return zone.dropRate * getSaleValue(zone.item) * vorticesAvail - zone.openCost();
+    };
+    return value(b) - value(a);
+  })[0];
   return bestFireExtinguisherZoneCached;
```
horrible-little-slime, line 1845: make sure the value is greater than 0 if we want to use it, as you pointed out in the discord.

src/fights.ts
```diff
@@ -1392,20 +1392,26 @@ const freeRunFightSources = [
       have($skill`Comprehensive Cartography`) &&
       get("_monstersMapped") < 3 &&
       get("_VYKEACompanionLevel") === 0 && // don't attempt this in case you re-run garbo after making a vykea furniture
-      getBestFireExtinguisherZone() !== undefined,
+      getBestFireExtinguisherZone() !== undefined &&
```
horrible-little-slime, line 1395: I really don't like letting this have two nullish return values; is there a reason to have these be separate returns?

s-k-z: It can't be anything else unless both values are not used anymore. Undefined means not yet set, and null means deliberately set to nothing.

PR #640, opened 2021-12-22
comb rare tiles

horrible-little-slime:


src/beach.ts
```diff
@@ -0,0 +1,730 @@
+import { cliExecute, myId } from "kolmafia";
+import { get, set } from "libram";
+
+type BeachTile = { minute: number; row: number; column: number };
+
+const rareTiles: BeachTile[] = [
+  { minute: 41, row: 4, column: 6 },
```
Thoth19, line 7: Would it be better to store this data in a kolmafia data file? Or is that more work? I figure that adding new data to this list will be more painful if the data is in a code file.

src/beach.ts
```diff
@@ -0,0 +1,730 @@
...
+  { minute: 9966, row: 8, column: 4 },
+  { minute: 9971, row: 8, column: 2 },
+  { minute: 9971, row: 10, column: 3 },
+  { minute: 9976, row: 8, column: 7 },
+  { minute: 9992, row: 9, column: 2 },
+];
+
+function _comb(tile: BeachTile): void {
+  const { minute, row, column } = tile;
+  cliExecute(`beach wander ${minute}; beach comb ${row} ${column}`);
```
Thoth19, line 708: What do we want to do if the tide is up and we can't hit a given tile? One option is to just omit all of those tiles from the pool.

src/beach.ts
```diff
@@ -0,0 +1,730 @@
...
+      .map((_value, index) => index + 1)
+      .filter((value) => value % 17 !== 0 && value % 41 !== 0);
+    const index = id % possibleStepValues.length;
+    stepLength = possibleStepValues[index];
+  }
+  return stepLength;
+}
+
+export default function comb(): void {
+  const index = getStepLength() + get("garbo_lastTileCombed", 0);
```
Thoth19, line 726: Oh this is cute. We are going to choose a random length of a step via playerId and then walk through the list of rare tiles in units of that step length. I like this move.

src/beach.ts
horrible-little-slime: Very good point! The short answer is "I don't know how the beach comb works well enough to have planned for this."

Would love to know the long answer. If we skip them, then the subsequent tiles have a higher load.

src/beach.ts
horrible-little-slime: That is probably a better idea. I should read up on how to do that.

src/dailies.ts
```diff
@@ -591,3 +597,11 @@ function pantogram(): void {
   visitUrl("inv_use.php?pwd&whichitem=9573");
   visitUrl(`choice.php?whichchoice=1270&pwd&option=1&m=${m}&e=5&s1=5789,1&s2=706,1&s3=24,1`);
 }
+
+function combBeach(): void {
```
Thoth19, line 601: Do we want to do combing as overdrunk turns if the ascend flag is turned on? Barf mpa is probably still higher right?

horrible-little-slime: I think for now we should only use people's free combs; people run garbo to run garbo, not to comb the beach.

src/beach.ts
docrostov: the kol beach, as with any normal beach, has tides. at low tide, the entire 10x10 grid is accessible. each day in the beach cycle, the tide changes -- it goes up one row for 4 days and down one row for 4 days. so the bottom 4 rows are not always accessible, especially the bottom-most one, which is only available once every eight days. the tides are reset at rollover, while the rest of the beach is not, which makes it even weirder in terms of accessibility.

src/beach.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export default function comb(): void {
+  const tileList = getShuffledArray();
+  const index = (get("garbo_lastTileCombed", 0) + 1) % tileList.length;
+  const tile = tileList[index];
+  const dayOfMonth = 1 + (gamedayToInt() % 8);
+  const rowsHidden = 4 - Math.abs(4 - dayOfMonth);
+  const shouldComb = tile.row > rowsHidden;
+  if (shouldComb) _comb(tile);
+  set("garbo_lastTileCombed", index);
```
Thoth19, line 45: What I'm reading here is that if the tide is high, we will skip tiles and act as if we had hit them and keep moving. This is likely fine for a first pass. On a second pass, we might want to prioritize tidal tiles in some way bc you can't hit them every day and we don't want them to go to "waste". (I really like this problem bc it has so many constraints)

src/beach.ts
```diff
@@ -0,0 +1,47 @@
...
+    shuffledArray = deterministicShuffle(rareTiles);
+  }
+  return shuffledArray;
+}
+
+export default function comb(): void {
+  const tileList = getShuffledArray();
+  const index = (get("garbo_lastTileCombed", 0) + 1) % tileList.length;
+  const tile = tileList[index];
+  const dayOfMonth = 1 + (gamedayToInt() % 8);
```
Thoth19, line 41: I'm also going to assume that gamedayToInt works the way you intend it to. I assume you've tested that the beach data this gives matches what we are observing from logs etc.

src/beach.ts
Thoth19: Maybe we should comment some of this information into the code?

Thoth19 (commented):
Seems legit.

src/beach.ts
horrible-little-slime: I think I will probably port at least some of this into libram. Definitely the functions for tidal prediction.

src/beach.ts
horrible-little-slime: That's a very good point--I'm not sure how to do that in this paradigm, but I think you're also right that we should just push this without addressing it, and address it later.

src/beach.ts
horrible-little-slime: This is code stolen from SSBBHax, who thought that linking tide patterns to the game month was a hacky solution instead of a great one. gamedayToInt does; today is day 83, and it's Boozember 4. We want to get 4 out of it, so this code should do that.

horrible-little-slime:
#646

PR #645, opened 2021-12-23
Pick the best cargo pocket

Rinn:


src/dailies.ts
```diff
@@ -591,3 +600,45 @@ function pantogram(): void {
...
+    }
+    if (pocket in meats) {
+      value += Object.values(pocketMeat(pocket))
+        .map((x) => parseInt(x))
+        .reduce((prev, cur) => prev + cur, 0);
+    }
+    if (pocket in scraps) {
+      value += 200;
+    }
+    return value;
```
horrible-little-slime, line 630: I don't particularly understand why you've set up this function this way. Are there pockets that fit more than one category?

horrible-little-slime (approved):
Structure seems a little odd, but generally looks good

PR #647, opened 2021-12-23
Reset the orb by visiting the shore

pstalcup:
Also more accurately count the number of embezzlers you will fight from being gregarious


horrible-little-slime:
Per discussion with @docrostov, I think the behavior we need is that we visit the shore AND run a different EmbezzlerFight

PR #665, opened 2021-12-30
only use a shotglass if you have a shotglass

horrible-little-slime:


docrostov (approved):
this looks right to me. i checked ash and the $item is correctly named. no issues.

PR #672, opened 2021-12-30
More mana burning

Rinn:
- Burn mp again after using express card or license to chill
- Don't maximize mp unless pyec or license will be used
- Burn mp after restoring hp or mp, leaving a threshold of double the target restore, to use any extra mp after a large restore (like  from eating a sausage)
- Enable allowNonMoodBurning only if not ascending, as the extra turns would be wasted
- Enable summon burning if not set, but don't burn on libram summons as that is handled manually based on sale price

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -23,7 +23,6 @@ import {
   myInebriety,
   myLevel,
   myMaxhp,
-  myMp,
   myPathId,
   numericModifier,
   outfit,
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -23,6 +23,7 @@ import {
   myInebriety,
   myLevel,
   myMaxhp,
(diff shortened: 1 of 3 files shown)


src/fights.ts
```diff
@@ -176,6 +176,7 @@ function embezzlerSetup() {
     if (have($item`Platinum Yendorian Express Card`) && !get("expressCardUsed")) {
       burnLibrams();
       use($item`Platinum Yendorian Express Card`);
+      burnLibrams();
```
horrible-little-slime, line 179: Can this (and line 188) burn down to 200, rather than to 0?

horrible-little-slime (changes requested):

horrible-little-slime (approved):

PR #682, opened 2022-01-01
Comb-head-during-dailybuffs

horrible-little-slime:


src/dailies.ts
```diff
@@ -274,6 +275,8 @@ function dailyBuffs(): void {
     ensureEffect($effect`Down the Rabbit Hole`);
     cliExecute("hatter 22");
   }
+
+  BeachComb.tryHead($effect`Do I Know You From Somewhere?`);
```
Rinn, line 279: Will this get the effect if the free combs are used and it would cost an adventure?

horrible-little-slime: yes it would. We probably want to do that no matter what though, right?

Rinn: If that's the case it should be done first so we don't burn a turn of all the other daily buffs acquired before it.

Rinn (approved):

PR #685, opened 2022-01-01
olfact garbage tourists

horrible-little-slime:


lewismd13 (approved):

src/combat.ts
```diff
@@ -247,7 +247,7 @@ export class Macro extends StrictMacro {
         Macro.if_($monster`Knob Goblin Embezzler`, Macro.trySkill($skill`Meteor Shower`))
       )
       .externalIf(
-        !have($effect`On the Trail`) && have($skill`Transcendent Olfaction`),
+        have($skill`Transcendent Olfaction`),
```
Rinn, line 250: I would still check olfactedMonster here, so we can skip adding unneeded logic to the macro if on the trail is already active

PR #686, opened 2022-01-01
Lucky! as an embezzler source & wish alternative

Rinn:
Also stubbed out using 11-leaf clovers for unlimited embezzlers, but not currently enabled

### DIFF
--- a/src/embezzler.ts
+++ b/src/embezzler.ts
@@ -530,19 +530,6 @@ export const embezzlerSources = [
       adventureMacro($location`Cobb's Knob Treasury`, embezzlerMacro());
     }
   ),
-  /*new EmbezzlerFight(
-    "11-leaf clover",
-    () =>
-      canAdv($location`Cobb's Knob Treasury`, true) &&
-      have($item`11-leaf clover`)
-    () => (canAdv($location`Cobb's Knob Treasury`, true) ? itemAmount($item`11-leaf clover`) : 0),
-    () => {
-      use($item`11-leaf clover`);
-      if (have($effect`Lucky!`)) {
-        adventureMacro($location`Cobb's Knob Treasury`, embezzlerMacro());
-      }
-    }
-  ),*/
   //These are very deliberately the last embezzler fights.
   new EmbezzlerFight(
     "11-leaf clover (untapped potential)",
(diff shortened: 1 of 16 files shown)


src/lib.ts
```diff
@@ -91,6 +92,18 @@ export const baseMeat =
...
+export function averageEmbezzlerNet(): number {
+  return ((baseMeat + 750) * meatDropModifier()) / 100;
+}
+
+export function averageTouristNet(): number {
+  return (baseMeat * meatDropModifier()) / 100;
+}
+
+export function averageEmbezzlerProfit(): number {
+  return averageEmbezzlerNet() - averageTouristNet();
```
horrible-little-slime, line 104: My concern with this is that it's the equivalent of just `750 * meatDropModifier() / 100`--drops gear is an important part of your average tourist net that often gets reduced in your average embezzler net.

horrible-little-slime: Should we call it `expected` instead of `average`?

src/embezzler.ts
```diff
@@ -522,25 +522,80 @@ export const embezzlerSources = [
...
   ),*/
+  new EmbezzlerFight(
+    "Lucky!",
+    () => canAdv($location`Cobb's Knob Treasury`, true) && have($effect`Lucky!`),
+    () => (canAdv($location`Cobb's Knob Treasury`, true) && have($effect`Lucky!`) ? 1 : 0),
+    () => {
+      adventureMacro($location`Cobb's Knob Treasury`, embezzlerMacro());
+    }
+  ),
+  /*new EmbezzlerFight(
```
horrible-little-slime, line 533: I know that we currently have commented code in here because of the pillkeeper, but I think we should refrain from actively adding commented code to garbo--going to ask @phulin if he agrees that these are meaningfully different, though

horrible-little-slime (commented):
The surface-level implementation is very solid, but I have a few minor concerns/questions/comments about some of the framework you're adding for this.

src/embezzler.ts
phulin: yeah, please do not add new commented code

src/lib.ts
Rinn: this is just moving the old logic from local variables to a function. if we want to to be more robust we can (and probably should) take into account drop items and item drops etc later

PR #688, opened 2022-01-02
Update eslint libram and Carnivorous potted plant rule skip is no longer needed

s-k-z:


s-k-z:
Okay actually this is still necessary for now, but one day... hopefully...

PR #698, opened 2022-01-03
Persist total embezzlers fought today in a property

pstalcup:


Tokoeka (approved):

s-k-z:
I recommend using `gamedayToInt()` or `gamedayToString()` instead of `todayToString()` because the latter is based on user time not server time.

PR #702, opened 2022-01-04
Clean up embezzler.ts and fights.ts to be more straight forward

pstalcup:
Primary goal of this PR is to
a) make fights.ts more streamlined and to refer to specific behaviors of an EmbezzlerFight rather than referring to the fight by name
b) describe the behavior of an embezzler fight with a set of options with reasonable defaults to make it clear what behavior is nonstandard for a given fight
c) reduce clutter and duplication of code so future changes will be less invasive

### DIFF
--- a/src/embezzler.ts
+++ b/src/embezzler.ts
@@ -51,18 +51,45 @@ import { crateStrategy, equipOrbIfDesired } from "./extrovermectin";
 import { averageEmbezzlerNet, globalOptions, ltbRun, WISH_VALUE } from "./lib";
 import { determineDraggableZoneAndEnsureAccess, draggableFight } from "./wanderer";
 
-type EmbezzlerFightOptions = {
+/**
+ * Configure the behavior of the fights in use in different parts of the fight engine
+ * @interface EmbezzlerFightConfigOptions
+ * @member {Requirement[]?} requirements maximizer requirements to use for this fight (defaults to empty)
+ * @member {draggableFight?} draggable if this fight can be pulled into another zone and what kind of draggable it is (defaults to undefined)
+ * @member {boolean?} canStartWanderers if this fight can be used to initialize wanderers (defaults to true)
+ * @member {boolean?} gregariousReplace if this is a "monster replacement" fight - pulls another monster from the CSV (defautls to false)
+ * @member {boolean?} wrongEncounterName if mafia does not update the lastEncounter properly when doing this fight (defaults to value of gregariousReplace)
+ */
+interface EmbezzlerFightConfigOptions {
+  requirements?: Requirement[];
+  draggable?: draggableFight;
+  canStartWanderers?: boolean;
+  wrongEncounterName?: boolean;
+  gregariousReplace?: boolean;
+}
+
+/**
+ * Options to override the default behavior of a fight
+ * @interface EmbezzlerFightRunOptions
+ * @member {Location?} location where to run this fight
+ * @member {Macro?} macro the macro to use when running this fight
+ */
+interface EmbezzlerFightRunOptions {
   location?: Location;
   macro?: Macro;
-};
+}
 
 export class EmbezzlerFight {
+  name: string;
   available: () => boolean;
   potential: () => number;
-  run: (options: EmbezzlerFightOptions) => void;
+  run: (options: EmbezzlerFightRunOptions) => void;
   requirements: Requirement[];
-  draggable: boolean;
-  name: string;
+  draggable?: draggableFight;
+  canStartWanderers: boolean;
+  wrongEncounterName: boolean;
+  gregariousReplace: boolean;
+
   /**
    * This is the class that creates all the different ways to fight embezzlers
    * @classdesc Something goes here
@@ -70,24 +97,25 @@ export class EmbezzlerFight {
    * @prop {() => boolean} available Returns whether or not we can do this fight right now.
    * @prop {() => number} potential Returns the number of embezzlers we expect to be able to fight from this source.
    * @prop {(options: EmbezzlerFightOptions) => void} run This runs the combat, optionally using the provided location and macro. Location is used only by draggable fights.
-   * @prop {Requirement[]} requirements This is an array of requirements to do this fight
-   * @prop {boolean} [draggable=false] This tells the script if it is able the embezzler to a different zone
+   * @prop {EmbezzlerFightConfigOptions} options configuration options for this fight
    *
    */
   constructor(
     name: string,
     available: () => boolean,
     potential: () => number,
-    run: (options: EmbezzlerFightOptions) => void,
-    requirements: Requirement[] = [],
-    draggable = false
+    run: (options: EmbezzlerFightRunOptions) => void,
+    options: EmbezzlerFightConfigOptions = {}
   ) {
     this.name = name;
     this.available = available;
     this.potential = potential;
     this.run = run;
-    this.requirements = requirements;
-    this.draggable = draggable;
+    this.requirements = options.requirements ?? [];
+    this.draggable = options.draggable;
+    this.canStartWanderers = options.canStartWanderers ?? true;
+    this.gregariousReplace = options.gregariousReplace ?? false;
+    this.wrongEncounterName = options.wrongEncounterName ?? this.gregariousReplace;
   }
 }
 
@@ -142,7 +170,7 @@ export const embezzlerSources = [
       get("_sourceTerminalDigitizeMonster") === $monster`Knob Goblin Embezzler` &&
       getCounters("Digitize Monster", 0, 0).trim() !== "",
     () => (SourceTerminal.have() && get("_sourceTerminalDigitizeUses") === 0 ? 1 : 0),
-    (options: EmbezzlerFightOptions) => {
+    (options: EmbezzlerFightRunOptions) => {
       adventureMacro(
         options.location ?? determineDraggableZoneAndEnsureAccess(draggableFight.WANDERER),
         Macro.externalIf(
@@ -156,8 +184,9 @@ export const embezzlerSources = [
         ).step(options.macro ?? embezzlerMacro())
       );
     },
-    [],
-    true
+    {
+      draggable: draggableFight.WANDERER,
+    }
   ),
   new EmbezzlerFight(
     "Guaranteed Romantic Monster",
@@ -168,7 +197,7 @@ export const embezzlerSources = [
       (getCounter("Romantic Monster window end") !== -1 ||
         getCounters("Romantic Monster window end", -1, -1).trim() !== ""),
     () => 0,
-    (options: EmbezzlerFightOptions) => {
+    (options: EmbezzlerFightRunOptions) => {
       const location =
         options.location ?? determineDraggableZoneAndEnsureAccess(draggableFight.WANDERER);
       const macro = Macro.externalIf(
@@ -182,8 +211,9 @@ export const embezzlerSources = [
       ).step(options.macro ?? embezzlerMacro());
       adventureMacro(location, macro);
     },
-    undefined,
-    true
+    {
+      draggable: draggableFight.WANDERER,
+    }
   ),
   new EmbezzlerFight(
     "Enamorang",
@@ -196,7 +226,7 @@ export const embezzlerSources = [
       (have($item`LOV Enamorang`) && !get("_enamorangs"))
         ? 1
         : 0,
-    (options: EmbezzlerFightOptions) => {
+    (options: EmbezzlerFightRunOptions) => {
       adventureMacro(
         options.location ?? determineDraggableZoneAndEnsureAccess(draggableFight.WANDERER),
         Macro.externalIf(
@@ -210,8 +240,9 @@ export const embezzlerSources = [
         ).step(embezzlerMacro())
       );
     },
-    [],
-    true
+    {
+      draggable: draggableFight.WANDERER,
+    }
   ),
   new EmbezzlerFight(
     "Orb Prediction",
@@ -220,17 +251,19 @@ export const embezzlerSources = [
       $monster`Knob Goblin Embezzler`,
     () =>
       (have($item`miniature crystal ball`) ? 1 : 0) *
-      get("beGregariousCharges") *
-      +(get("beGregariousFightsLeft") > 0 ||
-      CrystalBall.currentPredictions(false).get($location`The Dire Warren`) ===
-        $monster`Knob Goblin Embezzler`
-        ? 1
-        : 0),
-    (options: EmbezzlerFightOptions) => {
+      (get("beGregariousCharges") +
+        (get("beGregariousFightsLeft") > 0 ||
+        CrystalBall.currentPredictions(false).get($location`The Dire Warren`) ===
+          $monster`Knob Goblin Embezzler`
+          ? 1
+          : 0)),
+    (options: EmbezzlerFightRunOptions) => {
       const macro = options.macro ?? embezzlerMacro();
       adventureMacro($location`The Dire Warren`, macro);
     },
-    [new Requirement([], { forceEquip: $items`miniature crystal ball` })]
+    {
+      requirements: [new Requirement([], { forceEquip: $items`miniature crystal ball` })],
+    }
   ),
   new EmbezzlerFight(
     "Time-Spinner",
@@ -249,7 +282,7 @@ export const embezzlerSources = [
       )
         ? Math.floor((10 - get("_timeSpinnerMinutesUsed")) / 3)
         : 0,
-    (options: EmbezzlerFightOptions) => {
+    (options: EmbezzlerFightRunOptions) => {
       const macro = options.macro ?? embezzlerMacro();
       withMacro(macro, () => {
         visitUrl(`inv_use.php?whichitem=${toInt($item`Time-Spinner`)}`);
@@ -276,7 +309,7 @@ export const embezzlerSources = [
       have($skill`Meteor Lore`)
         ? 10 - get("_macrometeoriteUses")
         : 0,
-    (options: EmbezzlerFightOptions) => {
+    (options: EmbezzlerFightRunOptions) => {
       equipOrbIfDesired();
       const baseMacro = options.macro ?? embezzlerMacro();
       const macro = Macro.if_(
@@ -287,8 +320,10 @@ export const embezzlerSources = [
         ).skill($skill`Macrometeorite`)
       ).step(baseMacro);
       adventureMacro($location`Noob Cave`, macro);
+    },
+    {
+      gregariousReplace: true,
     }
-    // do we want to equip orb on these guys?
   ),
   new EmbezzlerFight(
     "Powerful Glove",
@@ -305,7 +340,7 @@ export const embezzlerSources = [
       have($item`Powerful Glove`)
         ? Math.min((100 - get("_powerfulGloveBatteryPowerUsed")) / 10)
         : 0,
-    (options: EmbezzlerFightOptions) => {
+    (options: EmbezzlerFightRunOptions) => {
       equipOrbIfDesired();
       const baseMacro = options.macro ?? embezzlerMacro();
       const macro = Macro.if_(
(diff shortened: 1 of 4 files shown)


src/familiar.ts
```diff
@@ -181,3 +181,7 @@ export function freeFightFamiliar(): Familiar {
 
   return argmax(familiarValue);
 }
+
+export function pocketProfessorCopies(): number {
+  return 2 + Math.ceil(Math.sqrt(familiarWeight(myFamiliar()) + weightAdjustment()));
```
pstalcup, line 186: d'oh myFamiliar() here should be $familiar`pocket professor`

src/fights.ts
```diff
@@ -323,157 +319,137 @@ function bestWitchessPiece() {
...
-        ) {
-          if (["Macrometeorite", "Powerful Glove"].includes(fightSource.name)) {
+      // PROFESSOR COPIES
+      if (have($familiar`Pocket Professor`)) {
+        const profCopies = [
+          {
+            property: "_garbo_meatChain",
+            maximizeParameters: [], // implicitly maximize against meat
+            macro: firstChainMacro(),
+            goalMaximize: (requirements: Requirement) => meatOutfit(true, requirements),
```
horrible-little-slime, line 360: I don't like the name of this personally

src/fights.ts
```diff
@@ -323,157 +319,137 @@ function bestWitchessPiece() {
...
-        maximizeCached(requirements.maximizeParameters, requirements.maximizeOptions);
-        if (
-          get("_pocketProfessorLectures") <
-          2 + Math.ceil(Math.sqrt(familiarWeight(myFamiliar()) + weightAdjustment()))
-        ) {
-          if (["Macrometeorite", "Powerful Glove"].includes(fightSource.name)) {
-            saberCrateIfDesired();
-          }
-          withMacro(secondChainMacro(), () =>
+          if (get("_pocketProfessorLectures") < pocketProfessorCopies()) {
```
horrible-little-slime, line 391: Seeing it in action, I'd rename `pocketProfessorCopies` to `potentialPocketProfessorLectures` or `pocketProfectorLectureMax` or something

src/fights.ts
```diff
@@ -323,157 +319,137 @@ function bestWitchessPiece() {
...
       while (nextFight !== null) {
         const startTurns = totalTurnsPlayed();
         if (have($skill`Musk of the Moose`) && !have($effect`Musk of the Moose`)) {
           useSkill($skill`Musk of the Moose`);
         }
-        if (["Macrometeorite", "Powerful Glove"].includes(nextFight.name)) {
+        if (nextFight.gregariousReplace) {
           saberCrateIfDesired();
         }
         withMacro(embezzlerMacro(), () => {
```
horrible-little-slime, line 415: We don't need this wrapper anymore

src/fights.ts
```diff
@@ -323,157 +319,137 @@ function bestWitchessPiece() {
...
-              if (!have($effect`Fishy`)) use($item`fishy pipe`);
-              nextFight.run({ location: $location`The Briny Deeps` });
-            } else if (nextFight.draggable) {
-              const type =
-                nextFight.name === "Backup" ? draggableFight.BACKUP : draggableFight.WANDERER;
-              const location = determineDraggableZoneAndEnsureAccess(type);
-              setLocation(location);
-              meatOutfit(true, Requirement.merge(nextFight.requirements));
-              nextFight.run({ location });
+            const romanticFamiliar = $familiars`Obtuse Angel, Reanimated Reanimator`.find(have);
```
horrible-little-slime, line 417: any time i see `.find(have)` it sets off  kill bill sirens, but this is fine for familiars

src/fights.ts
```diff
@@ -323,157 +319,137 @@ function bestWitchessPiece() {
...
+
+            const location = nextFight.draggable
+              ? checkUnderwater()
+                ? $location`The Briny Deeps`
+                : determineDraggableZoneAndEnsureAccess(nextFight.draggable)
+              : $location`Noob Cave`;
+
+            setLocation(location);
+            meatOutfit(true, Requirement.merge(nextFight.requirements));
+            nextFight.run({ location });
```
horrible-little-slime, line 432: shouldn't this be `nextFight.run({ location: location })`, or do you not need to do that?

src/fights.ts
```diff
@@ -323,157 +319,137 @@ function bestWitchessPiece() {
...
+        // try to deliver the thesis
         const romanticMonsterPossible =
           (getCounter("Romantic Monster Window end") === -1 &&
             getCounters("Romantic Monster Window end", -1, -1).trim() === "") ||
           getCounter("Romantic Monster Window begin") > 0;
-        if (
-          !(nextFight && ["Backup", "Digitize", "Enamorang"].includes(nextFight.name)) &&
-          romanticMonsterPossible
-        ) {
+        if (!(nextFight && nextFight.draggable) && romanticMonsterPossible) {
```
horrible-little-slime, line 452: this logic is not correct. It should be 
```suggestion
        if (!romanticMonsterPossible && (!nextFight || !nextFight.draggable)) {
```
You can de morgan from there, which could bring it to something like
`!romanticMonsterPossible && !(nextFight && nextFight.draggable)`
which could then be
`!(romanticMonsterPossible || (nextFight && nextFight.draggable)`
But definitely it's wrong as is.

src/fights.ts
```diff
@@ -323,157 +319,137 @@ function bestWitchessPiece() {
...
-        postCombatActions();
       }
 
-      startWandererCounter();
-
       // REMAINING EMBEZZLER FIGHTS
       let nextFight = getNextEmbezzlerFight();
       while (nextFight !== null) {
         const startTurns = totalTurnsPlayed();
         if (have($skill`Musk of the Moose`) && !have($effect`Musk of the Moose`)) {
```
horrible-little-slime, line 409: We should consider moving this to be in the backup code

src/fights.ts
horrible-little-slime: to clarify, the withMacro() wrapper was for EmbezzlerFights that don't involve adventureMacro, like the photocopy or chateau painting. But because of the setup here, it wasn't updating embezzlerMacro at the right time. It's better & more elevant to include withMacro in the run() functions of those EmbezzlerFights, which we do now.

src/fights.ts
horrible-little-slime: update: you don't! neat!

src/embezzler.ts
```diff
@@ -1,13 +1,15 @@
 // import { canAdv } from "canadv.ash";
 import {
   abort,
+  booleanModifier,
```
horrible-little-slime, line 4: consider using `getModifier`

src/embezzler.ts
```diff
@@ -49,46 +51,131 @@ import { Macro, shouldRedigitize, withMacro } from "./combat";
...
+    (booleanModifier("Underwater Familiar") || familiarWaterBreathingEquipment.some(have)) &&
+    (have($effect`Fishy`) || (have($item`fishy pipe`) && !get("_fishyPipeUsed")))
+  ) {
+    // then check if the underwater copy makes sense
+    if (mallPrice($item`pulled green taffy`) < 10000 && retrieveItem($item`pulled green taffy`)) {
+      // unlock the sea
+      if (get("questS01OldGuy") === "unstarted") {
+        visitUrl("place.php?whichplace=sea_oldman&action=oldman_oldman");
+      }
+      if (!have($effect`Fishy`) && !get("_fishyPipeUsed")) use($item`fishy pipe`);
```
docrostov, line 172: overall the checkUnderwater function seems great, much better than the implicit way we were doing this previously. my one quibble is that i think we should probably call out in some sort of user-facing FAQ that if they do not have a fishy pipe they should think about getting one. maybe the answer is a small markdown file linked in the readme that just goes over some "unexpected" (ie non-directly-meat-related) resources that garbo can use for profit if the user has it, to let people know ways they can make garbo more efficient? fishy pipes are really pretty cheap, but not everyone ever dips into the sea.

src/embezzler.ts
```diff
@@ -49,46 +51,131 @@ import { Macro, shouldRedigitize, withMacro } from "./combat";
...
+ */
+interface EmbezzlerFightConfigOptions {
+  requirements?: Requirement[];
+  draggable?: DraggableFight;
+  canInitializeWandererCounters?: boolean;
+  wrongEncounterName?: boolean;
+  gregariousReplace?: boolean;
+}
+
+class EmbezzlerFightRunOptions {
```
horrible-little-slime, line 74: I'm assuming we have a wrapper class here per our discussion about frequently using lines like `const macro = options.macro ?? embezzlerFightMacro()`, but don't see how this resolves the issue. It feels like overkill to make this little class, although a docstring would go a long way to making this feel less unnecessary

src/embezzler.ts
```diff
@@ -49,46 +51,131 @@ import { Macro, shouldRedigitize, withMacro } from "./combat";
...
+  wrongEncounterName: boolean;
+  gregariousReplace: boolean;
+
   /**
    * This is the class that creates all the different ways to fight embezzlers
    * @classdesc Something goes here
    * @prop {string} name The name of the source of this fight, primarily used to identify special cases.
    * @prop {() => boolean} available Returns whether or not we can do this fight right now.
    * @prop {() => number} potential Returns the number of embezzlers we expect to be able to fight from this source.
    * @prop {(options: EmbezzlerFightOptions) => void} run This runs the combat, optionally using the provided location and macro. Location is used only by draggable fights.
```
horrible-little-slime, line 113:
```suggestion
   * @prop {(options: EmbezzlerFightOptions) => void} execute This runs the combat, optionally using the provided location and macro. Location is used only by draggable fights.
```

src/embezzler.ts
```diff
@@ -49,46 +51,131 @@ import { Macro, shouldRedigitize, withMacro } from "./combat";
...
+  run(options: { macro?: Macro; location?: Location } = {}): void {
+    const fightMacro = options.macro ?? embezzlerMacro();
+    if (this.draggable) {
+      this.execute(new EmbezzlerFightRunOptions(fightMacro, this.location(options.location)));
+    } else {
+      this.execute(new EmbezzlerFightRunOptions(fightMacro));
+    }
+  }
+
+  location(location?: Location): Location {
```
horrible-little-slime, line 144: I'm confused--doesn't this mean that `location()` for the final orb fight returns something other than `the dire warren`??

src/embezzler.ts
```diff
@@ -49,46 +51,131 @@ import { Macro, shouldRedigitize, withMacro } from "./combat";
...
   requirements: Requirement[];
-  draggable: boolean;
-  name: string;
+  draggable?: DraggableFight;
+  canInitializeWandererCounters: boolean;
+  wrongEncounterName: boolean;
+  gregariousReplace: boolean;
+
   /**
    * This is the class that creates all the different ways to fight embezzlers
```
docrostov, line 108: so, EmbezzlerFights uses ~400 lines of embezzler.ts, and is the core of what makes garbo Different from a person's own random script. given the gigantic weight and importance of this class, i think this docstring needs some work. things i think it could benefit from:

- a formatting guide of what a properly written EmbezzlerFight actually looks like
- a clearer statement in the overarching summary that this is a Class That Runs a Fight; IMO, from a documentation standpoint, it is important to differentiate between "accounting" functions/classes and classes that are actually running turns for you
- a quick note as to why noob cave is the "default" zone. yeah, i know -- we all know why it is. someone reading garbo to try and build their own script may not, and if they were to ever change how noob cave operated, it would be good to have a description of what garbo's proposed viable conditions for a default zone is so we could quickly swap a new zone in.
- for the potential fights, we should maybe note that it is used by embezzlerCount to go into potions.ts to help stave off future circular references, since we don't docstring embezzlerCount (and probably shouldn't)

src/embezzler.ts
```diff
@@ -366,19 +437,20 @@ export const embezzlerSources = [
...
+        Macro.if_($monster`fluffy bunny`, Macro.item($item`human musk`)).step(options.macro)
       );
     },
-    [
-      new Requirement([], {
-        forceEquip: $items`miniature crystal ball`.filter((item) => have(item)),
-      }),
-    ]
+    {
+      requirements: [
```
docrostov, line 447: at the risk of starting a debate that gets me thrown out of every garbo party ever run, i would recommend one of two things here. effectively, "Be Gregarious" and "Final Be Gregarious" are exactly the same EmbezzlerFight. the only difference is that if we have the crystal ball on the 3rd fight, we want to equip the crystal ball. given the lack of differentiation, i would recommend either:
- making them one EmbezzlerFight and adding a condition on the requirement to only equip the orb if there's just 1 greg left
- keeping them separate and adding a note to both that if any changes are made to one, they should be made to the other to try and stop errors from mismatched behavior on the two aligned embezzlerfights

src/embezzler.ts
pstalcup: Basically:

What i wanted people to be able to do is `options.location` within the EmbezzlerFight without having to check for null or undefined. In order to do that, location needs to be not null/undefined. In order for it to be not null or undefined, it needs to have value. I was not comfortable giving a nonsense default value for non wandering EmbezzlerFights - I would prefer those error out if they try to access the location. The easiest way to accomplish both of these things was wrapping it in a class with a getter that would raise if location was undefined.

src/embezzler.ts
pstalcup: No. This is only used for wanderers, as the jsdoc states.

src/embezzler.ts
horrible-little-slime: I think merging them is fine so long as we now have the tools to do so. Historically requirement was static. If it's a function now, it's very easy to add.

If it's _not_ a function, we can still merge them, but it's spicier. We'd need to equip orb in the run() function, which means that we're getting a suboptimal deal out of our maximizer string--familiar weight is nonlinear wrt meat drop, so that makes it bad.

Another issue is just that Orb is a disaster. I have no solution to this disaster, I just want to hang that albatross on someone else's neck as we discuss this.

src/embezzler.ts
```diff
@@ -49,44 +51,56 @@ import { Macro, shouldRedigitize, withMacro } from "./combat";
...
+import { familiarWaterBreathingEquipment, waterBreathingEquipment } from "./outfit";
+import { determineDraggableZoneAndEnsureAccess, DraggableFight } from "./wanderer";
 
 /**
  * Configure the behavior of the fights in use in different parts of the fight engine
  * @interface EmbezzlerFightConfigOptions
  * @member {Requirement[]?} requirements maximizer requirements to use for this fight (defaults to empty)
  * @member {draggableFight?} draggable if this fight can be pulled into another zone and what kind of draggable it is (defaults to undefined)
- * @member {boolean?} canStartWanderers if this fight can be used to initialize wanderers (defaults to true)
+ * @member {boolean?} canInitializeWandererCounters if this fight can be used to initialize wanderers (defaults to true)
```
horrible-little-slime, line 62: I still think this docstring isn't quite accurate. This doesn't tell us whether the fight itself initializes the counter--spooky putty, for example, doesn't initialize it--this tells us if it's safe to start our wanderer counter by adventuring in another location when this fight is up. I think.

horrible-little-slime:
My current opinion is that I have some qualms and some wants, but frankly, it's been a few weeks and merge conflicts are piling up. I vote that as long as it's working, we fix the conflicts, merge, and then put in another PR to fix any issues.

src/embezzler.ts
```diff
@@ -106,13 +106,30 @@ export class EmbezzlerFight {
...
-   * @prop {EmbezzlerFightConfigOptions} options configuration options for this fight
-   *
+   * @prop {() => boolean} available Returns whether or not we can do this fight right now (this may change later in the day).
+   * @prop {() => number} potential Returns the number of embezzlers we expect to be able to fight from this source given the current state of hte character
+   *  This is used when computing turns for buffs, so it should be as accurate as possible to the number of KGE we will fight
+   * @prop {(options: EmbezzlerFightRunOptions) => void} execute This runs the combat, optionally using the provided location and macro. Location is used only by draggable fights.
+   *  This is the meat of each fight. How do you initialize the fight? Are there any special considerations?
+   * @prop {EmbezzlerFightConfigOptions} options configuration options for this fight. see EmbezzlerFightConfigOptions for full details of all available options
+   * @example
+   * // suppose that we wanted to add a fight that will use print screens repeatedly, as long as we have them in our inventory
```
docrostov, line 118: so, I like the rework of this docstring quite a lot. I think it's a very good enumeration. my only small quibble is that this print screen example isn't currently in garbo, so if there are future changes we have to remember to update this example to match new changes (and an enterprising code-sniffer wouldn't be able to just compare the example with the actual embezzlerFight). this is not a huge issue though and i would put this more as a minor quibble/callout vs a real problem. i think this is a great example, it's just funny it isn't in embezzler.ts in general.

src/embezzler.ts
```diff
@@ -106,13 +106,30 @@ export class EmbezzlerFight {
...
+   * new EmbezzlerFight(
+   *  "Print Screen Monster",
+   *  () => have($item`screencapped monster`) && get('screencappedMonster') === $monster`Knob Goblin Embezzler`, // in order to start this fight, a KGE must already be screen capped
+   *  () => availableAmount($item`screencapped monster`) + availableAmount($item`print screen button`) // the total of potential of this fight is the number of already copied KGE + the number of potentially copiable KGE
+   *  () => (options: EmbezzlerFightRunOptions) => {
+   *    const macro = Macro
+   *      .externalIf(have($item`print screen button`), Macro.tryItem($item`print screen button`))
+   *      .step(options.macro); // you should always include the macro passed in via options, as it may have special considerations for this fight
+   *    withMacro(macro, () => useItem($item`screen capped monster`));
```
docrostov, line 127: do both $item`screen capped monster` & $item`screencapped monster` work? i would assume it's one or the other but if both work this is also ironically teaching me something about libram's handling of slightly-off names :p

src/embezzler.ts
```diff
@@ -106,13 +106,30 @@ export class EmbezzlerFight {
...
+   * @classdesc Embezzler Fight enc
    * @prop {string} name The name of the source of this fight, primarily used to identify special cases.
-   * @prop {() => boolean} available Returns whether or not we can do this fight right now.
-   * @prop {() => number} potential Returns the number of embezzlers we expect to be able to fight from this source.
-   * @prop {(options: EmbezzlerFightOptions) => void} run This runs the combat, optionally using the provided location and macro. Location is used only by draggable fights.
-   * @prop {EmbezzlerFightConfigOptions} options configuration options for this fight
-   *
+   * @prop {() => boolean} available Returns whether or not we can do this fight right now (this may change later in the day).
+   * @prop {() => number} potential Returns the number of embezzlers we expect to be able to fight from this source given the current state of hte character
+   *  This is used when computing turns for buffs, so it should be as accurate as possible to the number of KGE we will fight
```
docrostov, line 113: i think this is good for a docstring, no quibbles or changes needed. but from a general "garbo dev" standpoint we might want to think about how to handle the variance we're seeing on embezzler counts vs expectation. i realize that a lot of those are bad or good luck w/ extrovermectin, but the point here is super valid; garbo may be spending a lot more on buffs than we need it to if it gets this part wrong. i realize this is a pretty vague comment but i think your garbo-debug mode would be great, and if we find that there really is a certain standard deviation on embezzler count, we may want to include that as something output by "potential" that is used in potion calculations.

src/fights.ts
```diff
@@ -403,11 +403,15 @@ export function dailyFights(): void {
...
         postCombatActions();
 
+        print(`Finished ${nextFight.name}`);
         if (
           totalTurnsPlayed() - startTurns === 1 &&
           get("lastCopyableMonster") === $monster`Knob Goblin Embezzler` &&
           (nextFight.wrongEncounterName || get("lastEncounter") === "Knob Goblin Embezzler")
         ) {
+          if (nextFight.wrongEncounterName) {
+            print(`fight ${nextFight.name} reports wrong encounter!`);
```
docrostov, line 413: just commenting to say thank you for this, i like this change a lot

src/fights.ts
horrible-little-slime: @pstalcup

horrible-little-slime (commented):
oh god I never actually posted this comment

PR #706, opened 2022-01-05
Don't use Jokester's gun and champagne.

phulin:


burningbman (approved):
This is good. We should probably do something to indicate these fights are done with champagne bottle, but I don't have a good solution for that other than changing the variable name to something large and unwieldy.

PR #710, opened 2022-01-05
Add a warning if today is halloween

pstalcup:


gausie:
I'm dumb and need something like this for sure

horrible-little-slime:
Per discussion in discord, please add a userpref for disabling this. I don't think that qualifies as scope creep, given that it falls under the fairly rigid guideline of "things that make garbo terminate unceremoniously that are ultimately not that dangerous to just let it run for"

pstalcup:
I am goign to leave this PR open, but I would like to revisit it from a different perspective (since we do have time before halloween comes.) I created a new issue #716 where we can have some design discussion before committing to ideas.

horrible-little-slime:
Next halloween is 24 March 2022

horrible-little-slime:
:broom:

horrible-little-slime:
I'm closing this because of ctrlv and talks of the `ween` flag

PR #712, opened 2022-01-06
ban all papier mache equips

horrible-little-slime:


s-k-z:
If you have `autoSatisfyWithCloset` enabled this doesn't work. The maximizer will just take the item from closet and equip it. For any users relying on the closet to store items for PvP, it makes a lot of sense to turn `autoSatisfyWithCloset` on since so items that garbo uses are also pvpable.

s-k-z (approved):

PR #725, opened 2022-01-10
Use libram free run ActionSource

Rinn:
Free kills will come in another PR

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -579,7 +579,7 @@ class FreeRunFight extends FreeFight {
     if (!this.available()) return;
     if ((this.options.cost ? this.options.cost() : 0) > get("garbo_valueOfFreeFight", 2000)) return;
     while (this.available()) {
-      const runSource = tryFindFreeRun(this.options.familiar ? { noFamiliar: () => true } : {});
+      const runSource = tryFindFreeRun({ noFamiliar: () => this.options.familiar !== undefined });
       if (!runSource) break;
       useFamiliar(
         runSource.constraints.familiar?.() ?? this.options.familiar?.() ?? freeFightFamiliar()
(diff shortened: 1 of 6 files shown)


src/fights.ts
```diff
@@ -578,20 +579,18 @@ class FreeRunFight extends FreeFight {
     if (!this.available()) return;
     if ((this.options.cost ? this.options.cost() : 0) > get("garbo_valueOfFreeFight", 2000)) return;
     while (this.available()) {
-      const runSource = findRun(this.options.familiar ? false : true);
+      const runSource = tryFindFreeRun(this.options.familiar ? { noFamiliar: () => true } : {});
```
horrible-little-slime, line 582: is there a reason it's not `{ noFamiliar: () => this.options.familiar !== undefined }`?

src/lib.ts
```diff
@@ -212,199 +204,13 @@ export function tryFeast(familiar: Familiar): void {
...
-    (a, b) => mallPrice(a) - mallPrice(b)
-  )[0];
-  return new FreeRun(
-    "LTB",
-    () => retrieveItem(freeRunItem),
-    Macro.item(freeRunItem),
-    new Requirement([], {}),
-    () => retrieveItem(freeRunItem)
-  );
+export const ltbRun: () => ActionSource = () => {
```
horrible-little-slime, line 207: It feels like the new ActionSource system obsoletes this, but also I get that we probably don't want to do too much surgery on the grape all at once

horrible-little-slime (approved):
If this is tested, it looks good to me. Great work!

src/lib.ts
Rinn: yes that's why i left it, i'm gonna do another pass

src/fights.ts
Rinn: sure

PR #728, opened 2022-01-10
Use libram free kill ActionSource

Rinn:
currently testing


horrible-little-slime:
One thing I see right off the bat as you're putting this together is jokester handling--we actively don't want to use the jokester as we do sandworms themselves, because you can't equip it with the champagne glass; I don't know if there's a way to say "prioritize this for this fight", but that seems worth it.

Rinn:
> One thing I see right off the bat as you're putting this together is jokester handling--we actively don't want to use the jokester as we do sandworms themselves, because you can't equip it with the champagne glass; I don't know if there's a way to say "prioritize this for this fight", but that seems worth it.

Ah okay, that makes sense now as to why the jokester gun was preferred for the tot. I'll deal with that.

PR #731, opened 2022-01-11
Use love as our gun if you have it + a robort.

phulin:


src/outfit.ts
```diff
@@ -160,12 +161,17 @@ export function meatOutfit(embezzlerUp: boolean, requirement?: Requirement, sea?
         have($item`unwrapped knock-off retro superhero cape`) &&
         forceEquip.every((equipment) => toSlot(equipment) !== $slot`back`)
       ) {
-        if (!have($item`ice nine`)) {
+        const gun =
```
Rinn, line 164: my concern with this block is if you have neither but do have the retrocape this will fail to equip a weapon

horrible-little-slime: there's a retrieveItem call down below; is it an issue even with that?

horrible-little-slime (approved):

PR #732, opened 2022-01-12
Diet tweaks - limit potions, cheapest food, gregarious counts

pstalcup:
* limit potions to once per day
* only consider the cheapest of items with the same adventure yield
* properly count existing gregarious charge

### DIFF
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -58,7 +58,7 @@ import { acquire } from "./acquire";
 import { withVIPClan } from "./clan";
 import { embezzlerCount, estimatedTurns } from "./embezzler";
 import { expectedGregs } from "./extrovermectin";
-import { arrayEquals, globalOptions } from "./lib";
+import { argmax, arrayEquals, globalOptions } from "./lib";
 import { Potion, PotionTier } from "./potions";
 import synthesize from "./synthesis";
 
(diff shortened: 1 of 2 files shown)


src/diet.ts
```diff
@@ -362,21 +361,29 @@ function copiers(): MenuItem<Note>[] {
...
-  const alreadyGregarious =
-    get("beGregariousCharges") > 0 ||
-    (get("beGregariousFightsLeft") > 0 &&
-      get("beGregariousMonster") === $monster`Knob Goblin Embezzler`);
+  const currentlygregarious =
+    get("beGregariousFightsLeft") > 0 &&
+    get("beGregariousMonster") === $monster`Knob Goblin Embezzler`
+      ? 1
+      : 0;
+  const gregariousCharges = get("beGregariousCharges") ?? 0 + currentlygregarious;
```
horrible-little-slime, line 369:
```suggestion
  const gregariousCharges = (get("beGregariousCharges") ?? 0) + currentlygregarious;
```
also I don't think you should need ??, but maybe you do idk

src/diet.ts
```diff
@@ -362,21 +361,29 @@ function copiers(): MenuItem<Note>[] {
...
 
-  const gregCounts = expectedGregs().slice(alreadyGregarious ? 1 : 0);
+  // returns an array of expected counts for number of greg copies to fight per pill use
+  // the last value is how much you expect to fight per pill
+  const gregCounts = expectedGregs();
+  const defaultGreg = gregCounts[gregCounts.length - 1];
```
horrible-little-slime, line 374: I would call this a marginalGreg or a marginalExtrovermectinGregCount or something silly like that

src/potions.ts
```diff
@@ -211,7 +211,7 @@ export class Potion {
     }[] = [];
     const limitFunction = limit
       ? (quantity: number) =>
-          Math.min(limit - values.reduce((_total, tier) => tier.quantity, 0), quantity)
+          Math.min(limit - values.reduce((total, tier) => total + tier.quantity, 0), quantity)
```
horrible-little-slime, line 214: I don't off the top of my head get what this does, but I trust you

src/diet.ts
pstalcup: oh i could default get 0 instead

src/potions.ts
pstalcup: i can rewrite it to use sumNumbers

horrible-little-slime: oh my god now I see why that's an issue

yeah go ahead and use sumNumbers just because

that's hilarious.

src/diet.ts
horrible-little-slime: i guess what i'm saying is "does that return null sometimes" because if so, that's weird

horrible-little-slime: keep using the ?? if you need it, but I should put in a tiny mafia PR to make this default to zero

horrible-little-slime (approved):

PR #758, opened 2022-01-16
draggable/wanderer fixes

horrible-little-slime:


lewismd13 (approved):

src/fights.ts
```diff
@@ -420,11 +420,11 @@ export function dailyFights(): void {
         nextFight = getNextEmbezzlerFight();
 
         // try to deliver the thesis
-        const romanticMonsterPossible =
+        const romanticMonsterImpossible =
           (getCounter("Romantic Monster Window end") === -1 &&
             getCounters("Romantic Monster Window end", -1, -1).trim() === "") ||
```
Tokoeka, line 425: Doesnt `getCounter("Romantic Monster Window end") === -1 && getCounters("Romantic Monster Window end", -1, -1).trim() === "") ` imply we are past the romantic window?

horrible-little-slime: That implies that the window end counter does not exist, which means that we're probably done

Tokoeka: Ah, that tracks, and makes the logic make sense.

PR #763, opened 2022-01-18
Snapshotting of a garbo run

pstalcup:


src/lib.ts
```diff
@@ -384,10 +375,20 @@ export function checkGithubVersion(): void {
...
       print("Garbo is out of date. Please run 'svn update!", "red");
       print(`${process.env.GITHUB_REPOSITORY}/main is at ${mainSha}`);
     }
   }
 }
+
+export function realmAvailable(
+  element: "spooky" | "stench" | "hot" | "cold" | "sleaze" | "fr"
+): boolean {
+  return get(`_${element}AirportToday`) || get(`${element}AirportAlways`);
```
horrible-little-slime, line 389: does...this work for `fr`?

src/lib.ts
```diff
@@ -384,10 +375,20 @@ export function checkGithubVersion(): void {
...
+}
+
+export function fmt(num: number): string {
```
horrible-little-slime, line 392: i know what this does but have no idea what this name is meant to mean

horrible-little-slime: format?
either way, new name

src/snapshot.ts
```diff
@@ -0,0 +1,346 @@
...
+} from "kolmafia";
+import { $item, $items, getFoldGroup, getSaleValue, property, set, sumNumbers } from "libram";
+import { fmt, HIGHLIGHT, resetDailyPreference } from "./lib";
+
+/**
+ * Return a mapping of the session items, mapping foldable items to a single of their forms
+ * @returns the item session results, with foldables mapped to a single of their folding forms
+ */
+function mySessionItemsWrapper(): Map<Item, number> {
+  const foldableMapping = (item: Item): [Item, Item][] =>
```
horrible-little-slime, line 19: I feel like I'd rather this be a `[Item, Item[]]` or, better yet a `{ representative: Item, group: Item[] }`, but that's neither here nor there

src/snapshot.ts
```diff
@@ -0,0 +1,346 @@
...
+  }
+  return inventory;
+}
+
+/**
+ * Performa a binary element-wise operation on two inventories
+ * @param a The LHS inventory to perform the operation on
+ * @param b The RHS inventory to perform the operation on
+ * @param op a function to compute between the sets
+ * @param transitive if true use the value of b for any items not in a. if false, ignore values not in a
```
horrible-little-slime, line 60: This is both an incorrect and imprecise name for this variable. i think it's more analogous to the commutator, but I think there's a better name regardless.

src/snapshot.ts
```diff
@@ -0,0 +1,346 @@
...
+    }
+  }
+  return inventory;
+}
+
+/**
+ * Performa a binary element-wise operation on two inventories
+ * @param a The LHS inventory to perform the operation on
+ * @param b The RHS inventory to perform the operation on
+ * @param op a function to compute between the sets
```
horrible-little-slime, line 59: Use the word operation in the docstring so people see why it's called an op

src/snapshot.ts
```diff
@@ -0,0 +1,346 @@
...
+ * @param transitive if true use the value of b for any items not in a. if false, ignore values not in a
+ * @returns a new map representing the combined inventories
+ */
+function inventoryOperation(
```
horrible-little-slime, line 63: I feel like this function could benefit from generic typing, but this is based on vibes alone

src/snapshot.ts
```diff
@@ -0,0 +1,346 @@
...
+    };
+    bufferToFile(JSON.stringify(val), filename);
+  }
+
+  /**
+   * Import a snapshot from a file in the data/ directory. Conventionally the file should end in ".json"
+   * @param filename The file from which to import
+   * @returns the Snapshot represented by the file
+   */
+  static fromFile(filename: string): Snapshot {
```
horrible-little-slime, line 214: is there something we can do other than throw/abort if we put in an invalid file?

horrible-little-slime: I am made...nervous isn't the right word, but wiggly, by the idea of having a function that acts on _arbitrary_ data files.

src/snapshot.ts
```diff
@@ -0,0 +1,346 @@
...
+  print(` You lost meat on ${lowValue.length} items including:`);
+  printProfit(lowValue);
+}
+
+const garboValueCache = new Map<Item, number>();
+export function garboValue(item: Item): number {
+  const cachedValue = garboValueCache.get(item);
+  if (cachedValue === undefined) {
+    const specialValueCompute = specialValueLookup.get(item);
+    const value = specialValueCompute ? specialValueCompute() : getSaleValue(item);
```
horrible-little-slime, line 300: given that getSaleValue also caches, it feels a little redundant to do this. Should we entire bypass getSaleValue and recreate the code, caching it directly? We'll be building up _two_ maps for every item we calculate the price of.

src/snapshot.ts
```diff
@@ -0,0 +1,346 @@
...
+    garboValueCache.set(item, value);
+    return value;
+  }
+  return cachedValue;
+}
+export function garboAverageValue(...items: Item[]): number {
```
horrible-little-slime, line 306: I feel that, like with getSaleValue, we might as well just use the average value for everything, given that we have the spread operator. Sometimes my feelings are wrong though!

src/snapshot.ts
```diff
@@ -0,0 +1,346 @@
...
+  return sumNumbers(items.map(garboValue)) / items.length;
+}
+
+let snapshot: Snapshot | null = null;
+export function startSnapshot(): void {
+  snapshot = Snapshot.current();
```
horrible-little-slime, line 312: based on the name of this function, it feels like this should verify first that snapshot is null; we should either do that, or make the name not be `start`

src/fights.ts
```diff
@@ -1957,3 +1958,32 @@ function voidMonster(): void {
...
   postCombatActions();
 }
+
+export function printEmbezzlerLog(): void {
+  if (resetDailyPreference("garboEmbezzlerDate")) {
+    property.set("garboEmbezzlerCount", 0);
+    property.set("garboEmbezzlerSources", "");
+  }
+  const totalEmbezzlers =
+    property.getNumber("garboEmbezzlerCount", 0) +
```
horrible-little-slime, line 1968: I don't think you need to use `getNumber` if you assign a default, but I could be wrong about this.

src/fights.ts
```diff
@@ -1957,3 +1958,32 @@ function voidMonster(): void {
...
+    embezzlerLog.initialEmbezzlersFought +
+    embezzlerLog.digitizedEmbezzlersFought;
+
+  const allEmbezzlerSources = property
+    .getString("garboEmbezzlerSources")
+    .split(",")
+    .filter((source) => source);
```
horrible-little-slime, line 1975: I know what this line does but scotch would get a nosebleed and fall into an apoplectic rage upon seeing this, so it may be worth seeing if there's a slightly less funny way to filter out empty entries.

src/lib.ts
```diff
@@ -111,25 +113,14 @@ export function safeInterrupt(): void {
   }
 }
 
-export function persistEmbezzlerLog(): { total: number; sources: Array<string> } {
+export function resetDailyPreference(trackingPreference: string): boolean {
```
horrible-little-slime, line 116: based on the way  I see this getting used it might be worth doing
```suggestion
export function resetDailyPreference(trackingPreference: string, ...dependentPreferences: string[]): boolean {
```
It also might not!

src/post.ts
```diff
@@ -62,10 +62,15 @@ function numberology(): void {
   }
 }
 
+function updateMallPrices(): void {
```
horrible-little-slime, line 65: Possibly silly question: why do we want to update the snapshot every turn? why not just as we log out?

horrible-little-slime: or uhh stop running garbo. you know what i mean.

src/lib.ts
pstalcup: Fmt is pretty standard shortening for format. I can fix this but I prefer more concise names for things like this

src/snapshot.ts
pstalcup: It needs to be [Item, Item][] because it is beating a list of one to one , not one too many mappings

src/snapshot.ts
pstalcup: Can be fixed. This class will live in libram, but I put it here for review

src/snapshot.ts
pstalcup: Mafia only lets you read from the data/ directory, so it can only read files from there

src/snapshot.ts
pstalcup: I cached on this later because we bypass getSaleVslue for many items, and I was going to only cache those but it seemed silly to cache on just that later when I could just cache the full flow instead. I am not even a little worried about the memory implications of this

src/snapshot.ts
pstalcup: In my opinion start implies resetting the snapshot, but a doc would make this more explicit

src/fights.ts
pstalcup: I didn't write this lol I just moved it

src/lib.ts
pstalcup: I didn't do this because I thought that it would e cleaner to have it return a Boolean so the user could set their own defaults instead of forcing them to use empty string

src/post.ts
pstalcup: This is to spread out the searches on mall prices for the items we collect. If we didn't do this and Garbo errors, the error would be hidden by all the messages of searching through the mall

src/lib.ts
pstalcup: It does not so I will fix it

src/lib.ts
horrible-little-slime: if it's standard, that seems fine. my brain just keeps trying to read it as `famweight` despite the lack of w lol

src/snapshot.ts
horrible-little-slime: yeah, but I guess I mean what happens if somebody uses this function with a filename that points at something else? you'll get a big ol' error. Are we fine getting a big ol' error, or do we want to return null, or is there a mechanism where we can makethis only get pointed at files that we write?

src/snapshot.ts
horrible-little-slime: doc is a fine choice

horrible-little-slime:
Just a quick state of the union:
 - Needs update as per new libram
 - Needs a few PR comments addressed (these may just not be pushed yet):
 - - realmAvailable
 - - consider making garboValue take spread-out-arrays by default
 - - actually I think that's it
 - Needs merge conflicts resolved
 
 After that, she's good to go!

PR #772, opened 2022-01-24
add smithsness to diet

horrible-little-slime:


src/diet.ts
```diff
@@ -307,6 +307,8 @@ function menu(): MenuItem<Note>[] {
     new MenuItem($item`extra-greasy slider`),
     new MenuItem(mallMin(lasagnas)),
     new MenuItem(mallMin(smallEpics)),
+    new MenuItem($item`Miserable Pie`),
```
pstalcup, line 310: These items need to be in the potionMenu or it won't consider their extra value from the buffs

src/potions.ts
```diff
@@ -18,18 +17,25 @@ import {
...
   sumNumbers,
 } from "libram";
 import { acquire } from "./acquire";
 import { baseMeat, globalOptions, pillkeeperOpportunityCost } from "./lib";
 import { embezzlerCount, estimatedTurns } from "./embezzler";
+import { canAdv } from "canadv.ash";
 
 export type PotionTier = "embezzler" | "overlap" | "barf" | "ascending";
 const banned = $items`Uncle Greenspan's Bathroom Finance Guide`;
+const usingPurse =
```
pstalcup, line 35: I am a little hesitant about two things here:

1. This is baking in an assumption about equipment that could change in the future (what if a new IotM is released that is an off hand but it is slightly worse than the latte?)
2. Doing a canAdv on the static level is probably fine, but it makes me a little nervous

horrible-little-slime: I think it's not irresponsible to say that 1. is an issue for future-us. I don't want to iterate over Item.all() or anything, because the latte and the purse are both things that don't always give their maximal returns, and I have a feeling we'd do a lot of special casing. I'll make you less nervous, though.

PR #777, opened 2022-01-25
Diet tweaks

pstalcup:
Support lots of new potions and muffin

### DIFF
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -1,5 +1,6 @@
 import {
   availableAmount,
+  buy,
   chew,
   cliExecute,
   drink,
@@ -56,7 +57,7 @@ import { acquire } from "./acquire";
 import { withVIPClan } from "./clan";
 import { embezzlerCount, estimatedTurns } from "./embezzler";
 import { expectedGregs } from "./extrovermectin";
-import { argmax, arrayEquals, globalOptions, HIGHLIGHT } from "./lib";
+import { argmax, arrayEquals, globalOptions, HIGHLIGHT, realmAvailable } from "./lib";
 import { Potion, PotionTier } from "./potions";
 import { garboValue } from "./session";
 import synthesize from "./synthesis";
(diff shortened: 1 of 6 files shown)


src/diet.ts
```diff
@@ -715,6 +757,12 @@ export function consumeDiet(diet: Diet<Note>, name: DietName): void {
           synthesize(countToConsume, $effect`Synthesis: Greed`);
         } else if (getClanLounge()[`${menuItem.item}`] && itemType(menuItem.item) === "booze") {
           cliExecute(`drink ${menuItem.item}`);
+        } else if (menuItem.item === $item`campfire hot dog`) {
```
horrible-little-slime, line 760: this is getting to be a big enough statement that we might want to move to switch or something

horrible-little-slime: maybe a map?

src/diet.ts
```diff
@@ -464,6 +465,32 @@ export function potionMenu(
   const speakeasy = $item`Clan speakeasy`;
   const hasSpeakeasy = getClanLounge()[`${speakeasy}`];
 
+  const twiceHauntedPrice =
```
horrible-little-slime, line 468: just double checking here--do we have to think about crafting turns?

src/diet.ts
```diff
@@ -464,6 +465,32 @@ export function potionMenu(
...
+    Math.min(
+      garboValue($item`haunted orange`),
```
horrible-little-slime, line 470: I should have caught this elsewhere--we don't want to use `getSaleValue` and, by extension, `garboValue` for things we buy. `getSaleValue` multiplies the mallPrice by 0.9 when relevant, which is great for its stated purpose--the mallPrice is the price items _aren't_ selling at--but when we're comparing this against other things to buy, and when we're doing math to determine what things to eat, we decidedly don't want to be doing that.

pstalcup: In this particular instance I actually want to value items below their mall value if they see bottle of vodka or orange, but for the others I want mall value. I can write a version of this logic specifically for diet.

I don't think there is much of anything that uses garboValue to purchase except these recent changes

src/diet.ts
```diff
@@ -738,32 +754,71 @@ export function consumeDiet(diet: Diet<Note>, name: DietName): void {
...
+          $item`Rethinking Candy`,
+          (countToConsume: number, menuItem: MenuItem<Note>) =>
+            synthesize(countToConsume, menuItem.effect ?? $effect`Synthesis: Greed`),
+        ],
+        ...mayoActions,
+        ...speakeasyDrinks,
+      ]);
+
+      for (const menuItem of menuItems) {
+        const itemAction = itemActions.get(menuItem.item);
```
horrible-little-slime, line 816: I think you can just do `itemActions.get(menuItem.item)?.()` and have it be `| null` rather than `| "skip"`.

pstalcup: No, because i use falsey to see if there is no action, right? 
I guess I can rework the code to use null instead, but it requires an additional conditional or that ? you have

PR #781, opened 2022-01-28
Print out high/low value items at the end of the day

pstalcup:


horrible-little-slime (commented):
For short sessions, winners and losers could overlap--are we concerned about that?

PR #791, opened 2022-01-31
Holiday wanderers support

horrible-little-slime:


src/combat.ts
```diff
@@ -558,6 +559,18 @@ export function main(): void {
     Macro.if_($monster`Eldritch Tentacle`, Macro.basicCombat())
       .step(Macro.load())
       .submit();
+  } else if (getTodaysHolidayWanderers().length !== 0) {
```
Rinn, line 562: It's possible to have both Eldritch Attunement and for holiday wanderers to be available.

Rinn (approved):

PR #796, opened 2022-02-01
meat stasis rework

ktpolanski:
Summary of changes:
 - permit longer stasis if waiting for bittycar, hobo monkey or bjorn - I've seen them sometimes happen quite late in fights when I went to look at my pre-garbo logs
 - use a dictionary over a seal tooth for stasis if possible
 - throw OPS ahead of stasis to avoid double round awkwardness
 - redo the stasis code to have a single macro, with variables controlling the round count and whether to care about the monster HP
 - move the porquoise gun ahead of the stasis loop due to HP requirement awkwardness

Known issue in need of addressing:
 - `.kill()` would require a mild adjustment to deal with the longer stasis

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -332,7 +332,7 @@ export class Macro extends StrictMacro {
     const hpCheck = checkPassive
       ? `!hppercentbelow 25 && monsterhpabove ${passiveDamage}`
       : "!hppercentbelow 25";
-    // Same story but for the gun
+    // Same story but for the sixgun shot, which wants 40 more HP if possible
     const hpCheckSixgun = checkPassive
       ? `!hppercentbelow 25 && monsterhpabove ${passiveDamage + 40}`
       : "!hppercentbelow 25";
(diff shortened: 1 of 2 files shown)


src/combat.ts
```diff
@@ -307,76 +309,72 @@ export class Macro extends StrictMacro {
...
+    } else if (retrieveItem($item`dictionary`)) {
+      stasisItem = $item`dictionary`;
+    };
+
+    // Construct the monster HP component of the stasis condition
+    // Start by evaluating the passive damage
+    const passiveDamage =
+      maxPassiveDamage() + 5;
+    // If we don't care about killing the monster don't bother checking passive damage
+    let monsterHpCheck = '';
```
horrible-little-slime, line 327: make this a const, and use ternary notation to make it sometimes an empty string. Alternately, use ternary to sometimes not make it part of the BALLS condition it gets used in later.

src/combat.ts
```diff
@@ -563,4 +561,4 @@ export function main(): void {
   }
   while (inMultiFight()) runCombat();
   if (choiceFollowsFight()) visitUrl("choice.php");
-}
+}
```
horrible-little-slime, line 564: You need a newline at the end of the file for linting reasons.

PR #803, opened 2022-02-04
fix macro logic

horrible-little-slime:


src/combat.ts
```diff
@@ -170,51 +176,47 @@ export class Macro extends StrictMacro {
...
+            $items`Spooky Putty sheet, Spooky Putty monster`.some((item) => have(item)),
           Macro.tryItem(itemOrSkill)
         );
       case $item`Rain-Doh black box`:
         return this.externalIf(
-          get("_raindohCopiesMade") + Math.max(1, get("spookyPuttyCopiesMade")) < 6,
-          Macro.tryItem(itemOrSkill)
+          get("_raindohCopiesMade") + Math.max(1, get("spookyPuttyCopiesMade")) < 6 &&
+            $items`Rain-Doh black box, Rain-Doh box full of monster`.some((item) => have(item)),
+          Macro.tryHaveItem(itemOrSkill)
```
Rinn, line 187: this one needs to be tryItem, because the macro is generated before the combat starts and you may be just about to use a box full of monster and won't have the black box in your inventory yet

src/combat.ts
```diff
@@ -320,11 +322,14 @@ export class Macro extends StrictMacro {
     // Determine stasis item to use
     // Garbo already gets a seal tooth at the start of the day, so that's around always
     let stasisItem = $item`seal tooth`;
-    if (retrieveItem($item`facsimile dictionary`)) {
-      // The dictionaries are nicer though as they don't hurt the monster
-      stasisItem = $item`facsimile dictionary`;
-    } else if (retrieveItem($item`dictionary`)) {
-      stasisItem = $item`dictionary`;
+    if (!globalOptions.checkedForDictionaries) {
```
Rinn, line 325: stasisItem will only ever be set to a dictionary the first time this block is entered now

Rinn (changes requested):

src/combat.ts
horrible-little-slime: good catch; i had reverted putty but not doh

src/combat.ts
```diff
@@ -317,15 +319,12 @@ export class Macro extends StrictMacro {
...
-      // The dictionaries are nicer though as they don't hurt the monster
-      stasisItem = $item`facsimile dictionary`;
-    } else if (retrieveItem($item`dictionary`)) {
-      stasisItem = $item`dictionary`;
+    if (!globalOptions.stasisItem) {
+      globalOptions.stasisItem =
+        $items`facsimile dictionary, dictionary`.find((item) => retrieveItem(item)) ??
+        $item`seal tooth`;
     }
+    const stasisItem = globalOptions.stasisItem;
```
s-k-z, line 327: what about no global options, just something like
```ts
const checkGet = (i: Item) => have(i) && (itemAmount(i) > 0 || retrieveItem(i));
const stasisItem = $items`facsimile dictionary, dictionary, seal tooth`.find(checkGet);
if (!stasisItem) throw `Acquire a seal tooth and run garbo again.`;
```

horrible-little-slime: oh interesting

src/embezzler.ts
```diff
@@ -231,8 +231,11 @@ export const embezzlerMacro = (): Macro =>
   Macro.if_(
     $monster`Knob Goblin Embezzler`,
     Macro.if_($location`The Briny Deeps`, Macro.tryCopier($item`pulled green taffy`))
-      .trySkill($skill`Wink at`)
-      .trySkill($skill`Fire a badly romantic arrow`)
+      .externalIf(myFamiliar() === $familiar`Obtuse Angel`, Macro.trySkill($skill`Wink at`))
+      .externalIf(
+        myFamiliar() === $familiar`Reanimated Reanimator`,
+        Macro.trySkill($skill`Fire a badly romantic arrow`)
+      )
```
s-k-z, line 238: These are backwards. Obtuse angel lets you fire arrows. Reanimator lets you wink.

s-k-z (changes requested):

s-k-z (approved):

PR #810, opened 2022-02-09
Polar Vortex/XO-Skeleton Improvements

s-k-z:
Only olfact if we have more hugs and kisses
Save 2 olfactions for extrovermectin and barf mountain
Don't count fire extinguisher charges if you don't own the iotm

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -1914,8 +1914,9 @@ function getBestItemStealZone(): ItemStealZone | null {
         get("olfactedMonster") === zone.monster ||
         get("_gallapagosMonster") === zone.monster)
   );
-  const canExtinguish = have($item`industrial fire extinguisher`);
-  const vorticesAvail = canExtinguish ? Math.floor(get("_fireExtinguisherCharge") / 10) : 0;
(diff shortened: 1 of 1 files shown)


src/fights.ts
```diff
@@ -1916,7 +1914,8 @@ function getBestItemStealZone(): ItemStealZone | null {
         get("olfactedMonster") === zone.monster ||
         get("_gallapagosMonster") === zone.monster)
   );
-  const vorticesAvail = Math.floor(get("_fireExtinguisherCharge") / 10);
+  const canExtinguish = have($item`industrial fire extinguisher`);
+  const vorticesAvail = canExtinguish ? Math.floor(get("_fireExtinguisherCharge") / 10) : 0;
```
horrible-little-slime, line 1918:
```suggestion
  const vorticesAvail = have($item`industrial fire extinguisher`) ? Math.floor(get("_fireExtinguisherCharge") / 10) : 0;
```
Supes optional:
Deleting the canExtinguish line. Feels more stylistically similar to hugsAvail. Alternately, you can make a canHug variable.

horrible-little-slime (approved):

PR #813, opened 2022-02-12
Add more info to readme and items

ktpolanski:
- Add some relevant aftercore items to items.md
- Link items.md from the readme
- Take a stab at some low-shiny account support, by including bean's original sheet from the discord and trying to remove the SongBoom influence from the numbers; this feels like it could be done better than it currently is
- Add emphasis on Clubber = Good
- Slightly reword clan setting section, deemphasising the stash clan and mentioning they're optional (I remember struggling a bit with this when starting off)

Thoughts:
- Might be worth further deemphasising the stash clan and hiding it in the FAQ?
- Might be worth further emphasising the items.md link and bringing it up to its own little subsection?
- VOA stuff would be nice to bring up to current standard, but I'm not sure what this current standard is


ktpolanski:
Handled the various open ends I flagged, just the VOA left.

docrostov:
i support everything done here so far. wording seems to fit pretty well with everything around it. honestly, if you want, i have no real issue with just squashing it now and dealing with more VOA stuff later?

also re: the stashclan stuff, i agree it probably is a good idea to think about a different way. i am not sure what that way is, since it's useful to at least call out what a "useful" stash might have, but i am a bit concerned/annoyed by the (perhaps expected) number of people posting in #trading-discussion etc about how they need a stashclan now lol

PR #815, opened 2022-02-13
try maximizing again before throwing

horrible-little-slime:


horrible-little-slime:
@phulin I'm drafting this because it doesn't work as-is due to outfit caching. I think we should update libram to allow us to force a refresh of a cached outfit, I'm just not sure how to do that.

PR #819, opened 2022-02-15
Get Eldritch Attunement During Dailies (On GSH)

Tokoeka:


Tokoeka:
Currently just equips water wings, as I don't think we care about the stat buff

Tokoeka:
Possibly needs some form of "only get if you have free fights to use it on" preq, not sure if we have a "free fight sources" counter

horrible-little-slime:
Is this tested? I know we generally have Attunement support, but I want to make sure that's been verified to still work post-paste.

src/dailies.ts
```diff
@@ -666,3 +671,11 @@ function comb(): void {
   const combs = 11 - get("_freeBeachWalksUsed");
   cliExecute(`combo ${combs}`);
 }
+
+function getAttuned(): void {
+  if (holiday() === "Generic Summer Holiday" && !haveEffect($effect`Eldritch Attunement`)) {
```
horrible-little-slime, line 676: use have(), please

src/fights.ts
```diff
@@ -529,11 +530,18 @@ function bestFairy() {
 class FreeFight {
   available: () => number | boolean;
   run: () => void;
+  tentacle: () => boolean;
```
horrible-little-slime, line 533: I don't see the utility in making this a function rather than a boolean--it seems like a property that does not change.

horrible-little-slime (changes requested):

src/fights.ts
Tokoeka: Yeah, I am still a bit hazy on when to use one or tother, so I was just following how available and run were done. Changing

PR #823, opened 2022-02-16
autosell casings

horrible-little-slime:


horrible-little-slime:
Going to fix the root of this via libram

PR #830, opened 2022-02-21
support for unsupported character states

horrible-little-slime:
bit of an omnibus PR:
 - Makes messaging for unprepared characters more dramatic
 - Fixes issue with talking to the Old Man when underleveled
 - Removes the use of `.some(have)` for items (because that always returns true; it's fine for familiars and skills)
 - Fixes non-destructive but recurring issue where garbo attempts to do the romance arrow in the sea by switching to meat familiar after completing profchain

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -173,8 +173,10 @@ function barfTurn() {
     useFamiliar(meatFamiliar());
     const location = embezzlerUp
       ? !get("_envyfishEggUsed") &&
-        (booleanModifier("Adventure Underwater") || waterBreathingEquipment.some(have)) &&
-        (booleanModifier("Underwater Familiar") || familiarWaterBreathingEquipment.some(have)) &&
+        (booleanModifier("Adventure Underwater") ||
+          waterBreathingEquipment.some((item) => have(item))) &&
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -173,8 +173,10 @@ function barfTurn() {
...
       ? !get("_envyfishEggUsed") &&
-        (booleanModifier("Adventure Underwater") || waterBreathingEquipment.some(have)) &&
-        (booleanModifier("Underwater Familiar") || familiarWaterBreathingEquipment.some(have)) &&
+        (booleanModifier("Adventure Underwater") ||
+          waterBreathingEquipment.some((item) => have(item))) &&
+        (booleanModifier("Underwater Familiar") ||
+          familiarWaterBreathingEquipment.some((item) => have(item))) &&
         (have($effect`Fishy`) || (have($item`fishy pipe`) && !get("_fishyPipeUsed"))) &&
         !have($item`envyfish egg`)
         ? $location`The Briny Deeps`
```
Tokoeka, line 182: Does this want the Level > 11 check here also?

horrible-little-slime:
Good catch liege

PR #831, opened 2022-02-21
Fix boolean values on relay chnages.

burningbman:


horrible-little-slime:
Here king u dropped this 👑

PR #832, opened 2022-02-22
fix weird integer coercion issue

horrible-little-slime:


horrible-little-slime:
This needs a new libram, due to clan issues

PR #851, opened 2022-03-20
Colorize the hover/click of "Interrupt Garbo"

libraryaddict:


horrible-little-slime:
This is a good change that I will merge eventually, but I am extremely lazy about relay things and want to make sure it's colourblind-friendly. I don't think it's not, but it costs us nothing to check.

PR #852, opened 2022-03-21
add auto confirmation handling

burningbman:
Idea is for the auto confirmation to handle user confirms in a way aiming to ensure garbo keeps running and chooses options anticipating the most meat possible. To me it's expected that the optimal decision might not be made and that's fine since someone had to explicitly set this pref.

### DIFF
--- a/src/embezzler.ts
+++ b/src/embezzler.ts
@@ -54,11 +54,11 @@ import { usingThumbRing } from "./dropsgear";
 import { crateStrategy, equipOrbIfDesired } from "./extrovermectin";
 import {
   averageEmbezzlerNet,
-  userConfirmDialog,
   globalOptions,
   HIGHLIGHT,
   ltbRun,
   setChoice,
+  userConfirmDialog,
   WISH_VALUE,
 } from "./lib";
 import { familiarWaterBreathingEquipment, waterBreathingEquipment } from "./outfit";
--- a/src/embezzler.ts
+++ b/src/embezzler.ts
@@ -24,7 +24,6 @@ import {
   toMonster,
   toUrl,
   use,
-  userConfirm,
   visitUrl,
   wait,
 } from "kolmafia";
@@ -55,6 +54,7 @@ import { usingThumbRing } from "./dropsgear";
 import { crateStrategy, equipOrbIfDesired } from "./extrovermectin";
 import {
   averageEmbezzlerNet,
+  customUserConfirm,
   globalOptions,
   HIGHLIGHT,
   ltbRun,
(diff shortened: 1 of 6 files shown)


src/lib.ts
```diff
@@ -414,3 +419,9 @@ export function getChoiceOption(partialText: string): number {
   }
   return -1;
 }
+
+export function customUserConfirm(msg: string, defaultValue: boolean, timeOut?: number): boolean {
+  if (get("garbo_autoUserConfirm")) return defaultValue;
+  else if (timeOut) return userConfirm(msg, timeOut, defaultValue);
```
horrible-little-slime, line 425: Remove the `else` or add an `else` to the final return

src/lib.ts
```diff
@@ -414,3 +419,9 @@ export function getChoiceOption(partialText: string): number {
   }
   return -1;
 }
+
+export function customUserConfirm(msg: string, defaultValue: boolean, timeOut?: number): boolean {
+  if (get("garbo_autoUserConfirm")) return defaultValue;
```
horrible-little-slime, line 424:
```suggestion
  if (get("garbo_autoUserConfirm", false)) return defaultValue;
```

src/lib.ts
horrible-little-slime: just for consistency

src/lib.ts
```diff
@@ -414,3 +419,9 @@ export function getChoiceOption(partialText: string): number {
   }
   return -1;
 }
+
+export function customUserConfirm(msg: string, defaultValue: boolean, timeOut?: number): boolean {
```
horrible-little-slime, line 423: This function definitely should get renamed, although I'm not entirely sure to what.

src/embezzler.ts
```diff
@@ -761,8 +762,9 @@ export const embezzlerSources = [
         .map((source) => `${source.potential()} from ${source.name}`)
         .forEach((text) => print(text, HIGHLIGHT));
       globalOptions.askedAboutWish = true;
-      globalOptions.wishAnswer = userConfirm(
-        `Garbo has detected you have ${potential} potential ways to copy an Embezzler, but no way to start a fight with one. Current embezzler net (before potions) is ${averageEmbezzlerNet()}, so we expect to earn ${profit} meat, after the cost of a wish. Should we wish for an Embezzler?`
+      globalOptions.wishAnswer = customUserConfirm(
```
horrible-little-slime, line 765: I would argue that it is less destructive for these to default to `false`, but I can see why you'd want the default value to be otherwise

horrible-little-slime (commented):
Do we want to add this to the relay interface, or is that too much encouragement to use this?

I kind of want it to print `garbo_autoUserConfirm is set to true` for relevant users at the top of the log, so I know who to newspaper when they file bug reports

PR #853, opened 2022-03-21
no more triple size normally

horrible-little-slime:


src/fights.ts
```diff
@@ -1902,7 +1902,8 @@ function deliverThesis(): void {
   if (
     have($item`Powerful Glove`) &&
     !have($effect`Triple-Sized`) &&
-    get("_powerfulGloveBatteryPowerUsed") <= 95
+    get("_powerfulGloveBatteryPowerUsed") <= 95 &&
+    get("_powerfulGloveBatteryPowerUsed") % 10 === 5
```
burningbman, line 1906: Might be worth adding a comment explaining why since it isn't immediately obvious.

PR #856, opened 2022-03-23
move stash returns to EOD instead of start of day

horrible-little-slime:


src/clan.ts
```diff
@@ -26,6 +26,46 @@ export const stashItems = get("garboStashItems", "")
...
+              !retrieved &&
+              have($item`Spooky Putty monster`)
+            ) {
+              continue;
+            }
+            print(`Returning ${item} to ${getClanName()} stash.`, HIGHLIGHT);
+            if (putStash(item, 1)) stashItems.splice(stashItems.indexOf(item), 1);
+          }
+        });
+      } else print("Uh oh! You don't have a garbo_stashClan set to return items to.", "red");
```
burningbman, line 62: Why did you change this from an error to a print?

burningbman (approved):

PR #873, opened 2022-04-05
estimate goose experience per turn better

horrible-little-slime:


horrible-little-slime:
Largely obsoleted by other goose fix

PR #878, opened 2022-04-07
use retrieveItem rather than buy() at the end of acquire()

horrible-little-slime:


horrible-little-slime:
Obsoleted by upcoming libram features

PR #889, opened 2022-04-19
new libram; use retrievePrice for diet

horrible-little-slime:


horrible-little-slime:
I am medium-concerned about the counter handling here for the romance stuff, and I for sure want to cut that out into some library functions

horrible-little-slime:
I'm splitting this into two different PRs because they're each getting big

horrible-little-slime:
Pre-empted by its children

PR #904, opened 2022-05-06
Crate orb shenanigans

horrible-little-slime:
For players with orb but without olfaction, we've been doing some pretty suboptimal things to ensure crates during the replace:enemy phase of embezzlers. This does a lot to change that:
 - Slice up embezzlerSources into sources of several kinds, so we don't do a regular Greg fight when resetting orb for the other thing
  - Pseudosniff a crate even if we can't fullsniff it
  - Resequence our embezzlerSources a little bit


src/lib.ts
```diff
@@ -448,6 +450,20 @@ export function userConfirmDialog(msg: string, defaultValue: boolean, timeOut?:
...
 
+export const latteActionSourceFinderConstraints = {
+  allowedAction: (action: ActionSource): boolean => {
+    if (!have($item`latte lovers member's mug`)) return true;
+    const forceEquipsOtherThanLatte = (
+      action?.constraints?.equipmentRequirements?.().maximizeOptions.forceEquip ?? []
+    ).filter((equipment) => equipment !== $item`latte lovers member's mug`);
+    return (
+      forceEquipsOtherThanLatte.every((equipment) => toSlot(equipment) !== $slot`off-hand`) &&
+      forceEquipsOtherThanLatte.filter((equipment) => toSlot(equipment) === $slot`weapon`).length <
```
libraryaddict, line 461: Possible that garbo does 2 or 3 handed items?

src/extrovermectin.ts
```diff
@@ -219,9 +229,22 @@ function initializeDireWarren(): void {
...
-export function intializeExtrovermectinZones(): void {
+export function initializeExtrovermectinZones(): void {
   if (get("beGregariousFightsLeft") === 0) {
     if (hasMonsterReplacers()) initializeCrates();
     initializeDireWarren();
   }
 }
+
+export function ponderPrediction(location: Location): Monster | null {
+  visitUrl("inventory.php?ponder=1", false);
```
libraryaddict, line 240: Perhaps cache this if no turns has passed?

libraryaddict (approved):
I know you only moved the code I made comments on

src/extrovermectin.ts
horrible-little-slime: I plan on moving this to libram's upcoming CrystalBall.ponder function soon, which deliberately doesnt' cache it because you could have a free encounter that doesn't take a turn. So that micro-optimization would last at most a few days before it gets rolled back, unfortunately.

src/lib.ts
horrible-little-slime: very good point. What I should do is something where I sum the weaponHands of all the forceEquipsOtherThanLatte

PR #906, opened 2022-05-08
ponder orb

horrible-little-slime:


horrible-little-slime:
This appears to mostly be working; I'd like an Orb Understander to rubberstamp it, and maybe I should make a libram PR for this eventually

PR #909, opened 2022-05-14
Improve meatify logic when ascending

Pantocyclus:
Meatify Matter will try to trigger on the last possible free fight wanderer if the ascend flag is true, or any time under 50 adventures left if no free fight wanderers are available

### DIFF
--- a/src/familiar.ts
+++ b/src/familiar.ts
@@ -225,31 +225,49 @@ export function pocketProfessorLectures(): number {
 }
 
 export function timeToMeatify(): boolean {
-  if (!have($familiar`Grey Goose`) || get("_meatifyMatterUsed") || myInebriety() > inebrietyLimit()) return false;
-  else if ($familiar`Grey Goose`.experience >= 400) return true;
+  if (
+    !have($familiar`Grey Goose`) ||
+    get("_meatifyMatterUsed") ||
+    myInebriety() > inebrietyLimit()
+  ) {
+    return false;
+  } else if ($familiar`Grey Goose`.experience >= 400) return true;
   else if (!globalOptions.ascending || myAdventures() > 50) return false;
 
-  // Check Wanderers  
+  // Check Wanderers
   const totalTurns = totalTurnsPlayed();
-  const baseMeat = (have($item`SongBoom™ BoomBox`)) ? 275 : 250;
-  const usingLatte = (have($item`latte lovers member's mug`) && get("latteModifier").split(",").includes("Meat Drop: 40")) ? true : false;
+  const baseMeat = have($item`SongBoom™ BoomBox`) ? 275 : 250;
+  const usingLatte =
+    have($item`latte lovers member's mug`) &&
+    get("latteModifier").split(",").includes("Meat Drop: 40");
 
-  const nextProtonicGhost = (have($item`protonic accelerator pack`)) ? Math.max(0, get("nextParanormalActivity") - totalTurns) : Infinity;
-  const nextVoteMonster = (have($item`"I Voted!" sticker`) && get("_voteFreeFights") < 3) ? Math.max(0, (totalTurns % 11 - 1) % 11) : Infinity;  
-  const nextVoidMonster = (have($item`cursed magnifying glass`) && get("_voidFreeFights") < 5 && get("valueOfFreeFight", 2000) / 13 > baseMeat * (usingLatte ? 0.75 : 0.6)) ? -get("cursedMagnifyingGlassCount") % 13 : Infinity;
+  const nextProtonicGhost = have($item`protonic accelerator pack`)
+    ? Math.max(0, get("nextParanormalActivity") - totalTurns)
+    : Infinity;
+  const nextVoteMonster =
+    have($item`"I Voted!" sticker`) && get("_voteFreeFights") < 3
+      ? Math.max(0, ((totalTurns % 11) - 1) % 11)
+      : Infinity;
+  const nextVoidMonster =
+    have($item`cursed magnifying glass`) &&
+    get("_voidFreeFights") < 5 &&
+    get("valueOfFreeFight", 2000) / 13 > baseMeat * (usingLatte ? 0.75 : 0.6)
+      ? -get("cursedMagnifyingGlassCount") % 13
+      : Infinity;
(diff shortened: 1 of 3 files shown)


src/familiar.ts
```diff
@@ -222,10 +224,34 @@ export function pocketProfessorLectures(): number {
...
+  else if ($familiar`Grey Goose`.experience >= 400) return true;
+  else if (!globalOptions.ascending || myAdventures() > 50) return false;
+  
+  //Check Wanderers
+  const F = get("_sausageFights");
+  const totalTurns = totalTurnsPlayed();
+
+  const nextProtonicGhost = (have($item`protonic accelerator pack`)) ? Math.max(0, get("nextParanormalActivity") - totalTurns) : Infinity;
+  const nextVoteMonster = (have($item`"I Voted!" sticker`) && get("_voteFreeFights") < 3) ? Math.max(0, (totalTurns % 11 - 1) % 11) : Infinity;
+  const nextSausageGoblin = (have($item`Kramco Sausage-o-Matic™`)) ? Math.max(0, 4 + 3 * F + Math.pow(Math.max(0, F-5), 3) + get("_lastSausageMonsterTurn") - totalTurns) : Infinity;
```
horrible-little-slime, line 238: We fight probablistic sausages at end of day, not deterministic ones, so this part is a no-go

src/familiar.ts
```diff
@@ -222,10 +224,34 @@ export function pocketProfessorLectures(): number {
...
+  const nextVoidMonster = (have($item`cursed magnifying glass`) && get("_voidFreeFights") < 5) ? -get("cursedMagnifyingGlassCount") % 13 : Infinity;
```
horrible-little-slime, line 239: You'll need to somehow determine if the player will actually be wielding their CMG--you could do something like
```
get("valueOfFreeFight", 2000) / 13 > baseMeat * (usingLatte() ? 0.75 : 0.6)
```
, assuming that 0.75 is a decent estimate for latte

src/familiar.ts
```diff
@@ -222,10 +224,34 @@ export function pocketProfessorLectures(): number {
...
+    (nextProtonicGhost === 0) ? 50 : nextProtonicGhost,
+    (nextVoteMonster === 0) ? ((get("_voteFreeFights") < 2) ? 11 : Infinity) : nextVoteMonster,
+    (nextSausageGoblin === 0) ? 7 + 3 * F + Math.pow(Math.max(0, F-4), 3) : nextSausageGoblin,
+    (nextVoidMonster === 0) ? 13 : nextVoidMonster,
+  ].reduce((a, b) => (a < b) ? a : b);	
+
+  if (delay < myAdventures()) return false; //We can wait for the next free fight
+  else if (freeFightNow || $familiar`Grey Goose`.experience >= 121) return true;
+  
+  return false;
```
horrible-little-slime, line 256: Dumb question--are we calling this to determine familiar selection already?

Pantocyclus: Oh right we only do it for freeFightFamiliar() but not meatFamiliar(), so that needs updating too

src/familiar.ts
```diff
@@ -232,17 +232,20 @@ export function timeToMeatify(): boolean {
...
   // const nextSausageGoblin = (have($item`Kramco Sausage-o-Matic™`)) ? Math.max(0, 4 + 3 * F + Math.pow(Math.max(0, F-5), 3) + get("_lastSausageMonsterTurn") - totalTurns) : Infinity;
-  const nextVoidMonster = (have($item`cursed magnifying glass`) && get("_voidFreeFights") < 5) ? -get("cursedMagnifyingGlassCount") % 13 : Infinity;
+  const nextVoidMonster = (have($item`cursed magnifying glass`) && get("_voidFreeFights") < 5 && get("valueOfFreeFight", 2000) / 13 > baseMeat * (usingLatte ? 0.75 : 0.6)) ? -get("cursedMagnifyingGlassCount") % 13 : Infinity;
 
   // If any of the above are 0, then
   // (1) We should be fighting a free fight
   // (2) We meatify if Grey Goose is sufficiently heavy and we don't have another free wanderer in our remaining turns
 
-  const freeFightNow = (nextProtonicGhost === 0 || nextVoteMonster === 0 || nextSausageGoblin === 0 || nextVoidMonster === 0);
+  // const freeFightNow = (nextProtonicGhost === 0 || nextVoteMonster === 0 || nextSausageGoblin === 0 || nextVoidMonster === 0);
```
horrible-little-slime, line 247: No commented code unless there's a very compelling reason to do so

src/familiar.ts
```diff
@@ -222,10 +224,37 @@ export function pocketProfessorLectures(): number {
...
+export function timeToMeatify(): boolean {
+  if (!have($familiar`Grey Goose`) || get("_meatifyMatterUsed") || myInebriety() > inebrietyLimit()) return false;
+  else if ($familiar`Grey Goose`.experience >= 400) return true;
+  else if (!globalOptions.ascending || myAdventures() > 50) return false;
+
+  // Check Wanderers
+  // const F = get("_sausageFights");
+  const totalTurns = totalTurnsPlayed();
+  const baseMeat = (have($item`SongBoom™ BoomBox`)) ? 275 : 250;
+  const usingLatte = (have($item`latte lovers member's mug`) && get("latteModifier").split(",").includes("Meat Drop\: 40")) ? true : false;
```
horrible-little-slime: You should basically never have to do `x ? true : false`. You might even be able to import usingLatte from Outfit.ts
My bad for not mentioning it.

src/index.ts
```diff
@@ -193,6 +194,8 @@ function barfTurn() {
       }
       retrieveItem($item`pulled green taffy`);
       if (!have($effect`Fishy`)) use($item`fishy pipe`);
+    } else if (timeToMeatify()) {
```
horrible-little-slime, line 197:
```suggestion
    } else if (!embezzlerUp && timeToMeatify()) {
```

src/familiar.ts
Pantocyclus: When you put it like that it does indeed look silly, but I intentionally plucked it out for readability because the conditional was extremely lengthy

PR #910, opened 2022-05-16
cap price of robo drinks

horrible-little-slime:
Status: requires re-review


src/dailies.ts
```diff
@@ -226,13 +230,67 @@ function configureGear(): void {
...
+        priceCap: get("_envyfishEggUsed")
+          ? (750 + baseMeat) * (0.5 + ((4 + Math.sqrt(110 / 100)) * 30) / 100)
+          : 0,
+        isMandatory: false,
+      },
+      "Single entendre": { priceCap: 25000, isMandatory: true },
+      "Drive-by shooting": { priceCap: 50000, isMandatory: true },
+    };
+    for (const [drinkName, { priceCap, isMandatory }] of Object.entries(roboDrinks)) {
+      if (get("_roboDrinks").includes(drinkName)) continue;
```
Pantocyclus, line 287: apparently `get("_roboDrinks").includes(drinkName)` is case-sensitive, and mafia stores `single entendre` and `drive-by shooting` so the PR would always attempt to get a new set of single entendre and drive-by shooting (e.g. when re-running `garbo ascend` after overdrinking)
```suggestion
      if (get("_roboDrinks").toLowerCase().includes(drinkName.toLowerCase())) continue;
```

src/dailies.ts
```diff
@@ -226,13 +230,67 @@ function configureGear(): void {
...
       useFamiliar($familiar`Robortender`);
-      if (itemAmount(drink) === 0) retrieveItem(1, drink);
-      print(`Feeding robortender ${drink}.`, HIGHLIGHT);
+      const drink = toItem(drinkName);
+      withProperty("autoBuyPriceLimit", priceCap, () => retrieveItem(1, drink));
```
Pantocyclus, line 290: This acquires the item if we do not have to purchase anything from the mall, which may not be ideal (especially for the bloody nora)
```suggestion
      if (mallPrice(drink) <= priceCap) withProperty("autoBuyPriceLimit", priceCap, () => retrieveItem(1, drink));      
```

src/dailies.ts
```diff
@@ -226,13 +230,67 @@ function configureGear(): void {
...
+      },
+      "Feliz Navidad": { priceCap: 0.25 * felizValue() * estimatedTurns(), isMandatory: false },
+      "Bloody Nora": {
+        priceCap: get("_envyfishEggUsed")
+          ? (750 + baseMeat) * (0.5 + ((4 + Math.sqrt(110 / 100)) * 30) / 100)
+          : 0,
+        isMandatory: false,
+      },
+      "Single entendre": { priceCap: 25000, isMandatory: true },
+      "Drive-by shooting": { priceCap: 50000, isMandatory: true },
```
Pantocyclus, line 284: priceCap values seem low, especially for Single Entendre which appears to be selling at above 40k
At EoD we can expect robort to be around ~40lbs (an 80lbs lep proves 118.9% meat more than a 40lbs lep), so Drive-by shooting should be worth at least 1.189 * baseMeat * estimatedTurns(). More if we factor in other weight buffs + embezzlies.
Similarly, a ~40lbs fairy provides 83.9% item drops. Assuming we aren't capping the 15% drop for the park garbage, this should be worth 0.15 * 0.839 * 200 * 3 * estimatedTurns(). Single entendres are also worth more MPA for embezzlers (0.3*0.839*100*4 = 100.68) than garbage tourists (0.15*0.839*200*3 = 75.51)
Assuming ~400 turns for a low shiny account, this puts the lower bound priceCap of single entendre at 30204, and drive-by shooting (assuming 250 base meat) at 118900.
I'd further argue that:
1) We are usually running at >40lbs most of the day, so that lower bound would be ~1.2x to 1.5x higher than the above estimate
2) If we are setting a priceCap, then we do not really need isMandatory
3) (2) is especially true because it appears that someone is manipulating the mallprices of robort drinks, with single entendres up to 88k at the moment of writing this

src/dailies.ts
```diff
@@ -226,13 +230,67 @@ function configureGear(): void {
...
+}
+
 function prepFamiliars(): void {
   if (have($familiar`Robortender`)) {
-    for (const drink of $items`Newark, drive-by shooting, Feliz Navidad, single entendre, Bloody Nora`) {
-      if (get("_roboDrinks").includes(drink.name)) continue;
+    const roboDrinks = {
+      Newark: {
+        priceCap: 0.25 * newarkValue() * estimatedTurns(),
+        isMandatory: false,
```
pstalcup, line 274: I do not think you should call this `isMandatory` but instead just `mandatory`

src/dailies.ts
horrible-little-slime: We do need isMandatory alongside a price cap; the price cap tells us to abort if we're seeing insane prices, but we also don't want garbo to proceed without a drive-by shooting in our robort.

horrible-little-slime: Although the entendre specifically we could make not be mandatory, which may be what you're saying

Pantocyclus: Entendre probably should not be mandatory. 
It's hard to accurately estimate the value of drive-by shooting pre-diet (i.e. without computing the actual number of embezzlies), but safe to say if in the extreme case where the prices were manipulated into the millions and we are running a low number of turns (e.g. post-casual), I would much rather have garbo chug along with a monkey than just outright abort and refuse to run because it cannot acquire a drive-by shooting.

horrible-little-slime: Entendre is unmandatoried; I think I'll keep drive-by shooting mandatory for now, but in the coming weeks I'll be working on some code that will let garbo farm ingredients to help stabilize the price

lewismd13 (approved):

PR #912, opened 2022-05-16
Preliminary/Experimental marginal MPA estimation

Pantocyclus:
Creates a checkpoint/snapshot at 75adv and 25adv left to estimate marginal MPA from raw meat and items over those 50 turns


horrible-little-slime:
Has this been tested for weird edge cases? e.g., running garbo with fewer than 75 or fewer than 25 adventures?

Shiverwarp:
Just commenting that I'd really appreciate this feature, it's become more difficult to find VOA after the marginal familiars update

src/session.ts
```diff
@@ -188,6 +199,135 @@ export function valueSession(): void {
...
+  if (!barfSession) {
+    barfSession = Session.current();
+    barfSessionStartTurns = totalTurnsPlayed();
+  }
+
+  // Start tracking items if at least one of these is true
+  // 1) We have run at least 100 barf turns
+  // 2) We have less than 200 adv to run
+  if (
+    !marginalSession &&
```
docrostov, line 222: i don't mind these two as conditions for item tracking, but if they are going to be our only major conditions, we need a way somewhere in here to cut marginal reporting entirely if the user is running <200 turns overall -- otherwise it'll contain free fights and it will be a bad overestimate.

src/session.ts
```diff
@@ -188,6 +199,135 @@ export function valueSession(): void {
...
+  ) {
+    marginalSession = Session.current();
+    numTrackedItemTurns = myAdventures() - globalOptions.saveTurns - 25;
+  }
+
+  // Start tracking meat if we have less than 75 turns left
+  // Also create a backup tracker for items
+
+  if (
+    (!get("_garboMarginalMeatCheckpoint") || !get("_garboMarginalMeatTurns")) &&
```
docrostov, line 235: similarly, this meat tracker needs some sort of "if the user is only running 75 turns, do not report marginals" as discussed above with the items to ensure it does not encase initial embezzlers & yachtzee. to be clear, it doesn't need to be at these if statements -- it's fine if it tabulates this all regardless, but there should be -something- that shuts down reporting in cases where the reporting will clearly be wrong

src/session.ts
```diff
@@ -188,6 +199,135 @@ export function valueSession(): void {
...
+      numTrackedItemTurns &&
+      numTrackedItemTurns >= Math.max(50, meatTurns)
+    ) {
+      const { items, itemDetails } = marginalSessionDiff.value(garboValue);
+      const barfItemDetails = Session.current().diff(barfSession).value(garboValue).itemDetails;
+      const outlierItemDetails = itemDetails
+        .filter(
+          (detail) =>
+            detail.quantity === 1 &&
+            detail.value >= 1000 &&
```
docrostov, line 290: is 1000 a reasonable limit in assessing outlier items? that seems... low, to me, if we are talking true outliers. for instance, most candies dropped by robort will be over that value. many things from cleaver. knob visors. (ok the last one is a joke but still.) this feels like it will drag way too much of the profit into the outliers but not truly communicate why. what do yall think?

src/session.ts
```diff
@@ -188,6 +199,135 @@ export function valueSession(): void {
...
+          Math.round(IPA * 100) / 100
+        )} [items] (${formatNumber(
+          Math.round(outlierIPA * 100) / 100
+        )} [outliers]) = ${formatNumber(Math.round(totalMPA * 100) / 100)} [total] (${formatNumber(
+          Math.round(totalOutlierMPA * 100) / 100
+        )} [w/ outliers])`,
+        HIGHLIGHT
+      );
+    } else if (get("_garboMarginalItemValue")) {
+      const items = get("_garboMarginalMeatValue", 0);
```
docrostov, line 318: ... are we setting this pref? this way to get items/IPA seems very bizarre to me. wouldn't this just straight up double count MPA?

docrostov (commented):
structurally i don't see major issues with this. my issues are more around three things:

- we need to properly enumerate situations where this metric cannot be trusted
- we need to re-examine outliers; feels like outlier cutoff is way too low
- i don't get how we're doing IPA when there aren't enough item turns

overall i like the structure. let me know when this is updated and i will look again.

src/session.ts
Pantocyclus: Ah this is just a stupid error on my part, it should `get("_garboMarginalItemValue")` instead

src/session.ts
Pantocyclus: These are the "outliers" on the breakfast leg that I got yesterday
> 1 fake hand worth 50000 total
> 1 Sourfinger worth 43650 total
> 1 Pok&eacute;-Gro fertilizer worth 13860 total
> 1 recording of Chorale of Companionship worth 9225 total
> 1 chocolate seal-clubbing club worth 7416 total
> 1 chocolate pasta spoon worth 7076 total
> 1 salt wages worth 6919 total
> 1 goblin water worth 6296 total
> 1 peppermint sprout worth 5992 total
> 1 chocolate disco ball worth 5396 total
> 1 chocolate stolen accordion worth 5265 total
> 1 Friendliness Beverage worth 4946 total
> 1 fudge-shaped hole in space-time worth 3600 total
> 1 Mer-kin fastjuice worth 2699 total
> 1 eagle feather worth 2520 total
> 1 Mer-kin strongjuice worth 2156 total
> 1 gummi salamander worth 2052 total
> 1 half-orchid worth 1566 total
> 1 Comet Drop worth 1526 total
> 1 sterno-flavored Hob-O worth 1522 total
> 1 red drunki-bear worth 1521 total
> 1 sugar shorts worth 1521 total
> 1 tiny dancer worth 1494 total
> 1 lead necklace worth 1492 total
> 1 wind-up Whatsian robot worth 1229 total
> 1 skin oil worth 1080 total
> 1 green drunki-bear worth 1063 total
> 1 8-bit banana worth 1053 total
> 1 mother's necklace worth 1000 total
> Marginal MPA: 5,090.3 [raw] + 1,704.26 [items] (2,314.05 [outliers]) = 6,794.56 [total] (7,404.35 [w/ outliers])
I suppose ut does grab a few things from cleaver, robort and screeges, and I'm comfortable raising that limit to 5000. But the 1000-5000 items (totalling 34040 meat) really do not skew the outlier MPA calculations as much as the heavy hitters (totalling 161095 meat for the > 5000 ones)

src/session.ts
Pantocyclus: This is only triggered during barfTurns() so it should never consider free fights

src/session.ts
Pantocyclus: So in `printMarginalSession()` we only report outliers when we have run sufficiently many turns. We still provide an estimate if we ran less than sufficient turns, but we clearly cannot estimate outliers without sufficient data. I could put in a warning when we do this alternative printout.

horrible-little-slime:
@docrostov I know your plate is full right now, I'm just going through PRs and I wanted to keep this on your radar

docrostov (approved):
i went through the subsequent changes and i am overall comfortable with this feature and adding it. i have not tested the feature yet but the logic is solid and sound. can we resolve issues w/ main and get it merged up in the next few days?

src/session.ts
```diff
@@ -265,6 +265,7 @@ export function trackBarfSessionStatistics(): void {
   // Also create a backup tracker for items
 
   if (marginalSession) marginalFamiliarsExcessTotal += marginalFamiliarsExcessValue;
+  marginalFamiliarsExcessValue = 0;
```
eegeeZA, line 268: This happens after the `if` statement. Checking if the intention is to reset back to `0` here.

Pantocyclus: Yes. The code only sets `marginalFamiliarsExcessValue` when the `barfFamiliar()` code is called. If we do not reset it after adding this value to the total, then when we run a non-barf turn we would potentially continually add the excess value to the total, overaccounting for our marginal familiars (I guess technically we only call `trackBarfSessionStatistics()` when running barf so this shouldn't happen in theory, but I did hit a bug with overaccounting the other day so I'm in the process of figuring out what went wrong).

There's no harm always resetting it to 0 regardless of whether we are currently tracking the marginal session.

eegeeZA: Okay cool, that makes sense.

Pantocyclus: I had a similar idea, but unfortunately that would run into some issues with circular imports unless we do a whole bunch of refactoring.

`session.ts` provides not only `garboValue()` and a few other functionalities which the other files, so I cannot have it import from `barfFamiliars.ts`; the values of the marginal familiars are computed within the `/familiars` code itself (which does use `garboValue` to evaluate the drops), so it's hard to extract the logic from within.

I agree it's not the cleanest solution for now, but I believe it would take a pretty big overhaul to clean stuff up, which is probably out of scope for now (I think).

docrostov:
> i went through the subsequent changes and i am overall comfortable with this feature and adding it. i have not tested the feature yet but the logic is solid and sound. can we resolve issues w/ main and get it merged up in the next few days?

i posted this about a month before i got married and now i am married and this still hasn't been merged yet 😭 

have people been using the testing branch and found it working decently well?

eegeeZA:
I'll keep editing this comment to reflect the output of this PR compared to pTrack.

## 2023-05-07
### This PR
> Marginal MPA: 3,577.66 [raw] + 980.93 [items] (1,116.55 [outliers]) = 4,558.59 [total] (4,694.21 [w/ outliers])
### pPtrack
> You've earned 440 513 in item differences
You've earned 1 904 259 liquid meat
You've spent 538 adventures for a total (meat+item) 4358 mpa
From 09:36:32 beforeGarbo to 10:14:24 afterGarbo took 00:37:51.
You've earned a total of 2 344 772 meat between 20230507 beforeGarbo and 20230507 afterGarbo.

Pantocyclus:
I'm just slightly surprised that your overall MPA is anywhere close to, or even lower than, your marginal MPA (for comparison, my overall MPA is probably twice my marginal MPA), but at least all the estimates are consistent at the lower MPA range.

I do believe it's good to go at this stage.

eegeeZA:
Must be the lack of IotMs. 🤷‍♂️

eegeeZA:
Are there still plans to merge this?

Pantocyclus:
Unlikely

docrostov (dismissed):
@Pantocyclus i approved this twice and am approving again because the logic all makes sense and i'm generally happy with it. the testing performed here is fine. can we deal with merge conflicts and add this to main?

also, as a small note, please please please ping me on discord if you want me to re-review PRs at speed, i would have re-approved this much earlier if i had known that testing had been performed and seen the positive testing results. i do not spend a ton of time in this repo, and due to being an admin, i get an email/github notification at every single iota of activity performed on any loathers branch whatsoever, from an issue comment to a pull request to a comment on a PR etc. 

this means things like this slip through the cracks at the best of times, and "the summer where i spent 3 weeks traveling for work and 3 weeks traveling on my honeymoon and got sick as hell for 3 other weeks and also tried to do the challenge path and worked 50-60 hour weeks almost every non-honeymoon week" is assuredly not the best of times lol.

docrostov (approved):
approved, good work panto

PR #913, opened 2022-05-16
Grab a UMD if conditions are right

Pantocyclus:
Conditions being
1) Must own SBB charter
2) Mall price of UMD > how much we would obtain from the 2nd choice
3) At least 7 days have elapsed since the last UMD was obtained

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -2256,8 +2256,7 @@ function yachtzee(): void {
           2000 * (1 + numericModifier("meat drop") / 100) &&
         (!lastUMDDate || today - Date.parse(lastUMDDate) >= 1000 * 60 * 60 * 24 * 7);
 
-      if (getUMD) setChoice(918, 1);
-      else setChoice(918, 2);
+     setChoice(918, getUMD ? 1 : 2);
 
       adventureMacroAuto($location`The Sunken Party Yacht`, Macro.abort());
       if (
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -7,6 +7,7 @@ import {
   Effect,
   equip,
   Familiar,
+  gametimeToInt,
   getAutoAttack,
   getCampground,
   handlingChoice,
@@ -42,7 +43,6 @@ import {
   setLocation,
   stashAmount,
   takeCloset,
-  todayToString,
   toInt,
   toItem,
   toSlot,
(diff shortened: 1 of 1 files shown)


src/fights.ts
```diff
@@ -2247,8 +2248,21 @@ function yachtzee(): void {
 
       if (!equippedOutfit || !success()) return;
 
-      setChoice(918, 2);
+      const lastUMDDate = property.getString("umdLastObtained");
+      const today = todayToString().slice(0,4) + "-" + todayToString().slice(4,6) + "-" + todayToString().slice(6);
```
horrible-little-slime, line 2252: there is almost certainly a better way to do this

Pantocyclus: I'm not familiar with JS so I don't really know what the preferred coding style is / relevant standard libraries are to do date conversions. Could you advise on a better implementation?

src/fights.ts
```diff
@@ -2247,8 +2248,24 @@ function yachtzee(): void {
...
+      const lastUMDDate = property.getString("umdLastObtained");
+      const today = Date.now() - gametimeToInt() - 1000 * 60 * 3.5; // Import today from ./lib once the PR is merged
+      const getUMD =
+        !get("_sleazeAirportToday") && // We cannot get the UMD with a one-day pass
+        garboValue($item`Ultimate Mind Destroyer`) >=
+          2000 * (1 + numericModifier("meat drop") / 100) &&
+        (!lastUMDDate || today - Date.parse(lastUMDDate) >= 1000 * 60 * 60 * 24 * 7);
+
+      if (getUMD) setChoice(918, 1);
+      else setChoice(918, 2);
```
horrible-little-slime, line 2260: @docrostov may set my house on fire and say the exact opposite is true, but 
```suggestion
     setChoice(918, getUMD ? 1 : 2);
```
might be a little cleaner

src/fights.ts
```diff
@@ -2247,8 +2248,24 @@ function yachtzee(): void {
 
       if (!equippedOutfit || !success()) return;
 
-      setChoice(918, 2);
+      const lastUMDDate = property.getString("umdLastObtained");
```
horrible-little-slime, line 2251: I generally prefer using `get` with a default value, but this is fine

horrible-little-slime: What the heck does this property normally look like?

Pantocyclus: `umdLastObtained=2022-05-14` is my current pref, but I assume it should otherwise be just `umdLastObtained=`. Tbh I wasn't particularly sure of the return type of `get()` on the latter, so I wanted to just force it to return as a string (this was relevant before I made the `!lastUMDDate` falsy change, not sure if it is as relevant now).

PR #915, opened 2022-05-19
Grab more stench jellies where possible

Pantocyclus:


src/combat.ts
```diff
@@ -286,6 +286,10 @@ export class Macro extends StrictMacro {
           Macro.trySkill($skill`Feel Nostalgic`)
         )
       )
+      .externalIf(
+        myFamiliar() === $familiar`Space Jellyfish`,
+        Macro.if_($monster`garbage tourist`, Macro.trySkill($skill`Extract Jelly`))
```
horrible-little-slime, line 291: All tourists in the zone are stench-aligned; I think libram allows monster-arrays now, so you can use $monsters or you can use `monstername * tourist` or something like that

src/index.ts
```diff
@@ -196,6 +196,20 @@ function barfTurn() {
...
+      !embezzlerUp &&
+      have($familiar`Space Jellyfish`) &&
+      get(`_spaceJellyfishDrops`) < 5 &&
+      clamp(
+        myAdventures() - digitizedMonstersRemaining(),
+        globalOptions.saveTurns,
+        myAdventures()
+      ) -
+        globalOptions.saveTurns <
+        20 &&
```
horrible-little-slime, line 209: I might be about to call the kettle black here, but this math feels really hard to follow--could we get some comments and/or slicing it out into a function?

horrible-little-slime:
Lookin good, just a few remarks to address

src/index.ts
Pantocyclus: Upon second thoughts the clamp() is not really needed, since I can accept negative values on the LHS just fine. That should simplify things greatly.

src/index.ts
```diff
@@ -200,13 +200,7 @@ function barfTurn() {
...
       have($familiar`Space Jellyfish`) &&
       get(`_spaceJellyfishDrops`) < 5 &&
-      clamp(
-        myAdventures() - digitizedMonstersRemaining(),
-        globalOptions.saveTurns,
-        myAdventures()
-      ) -
-        globalOptions.saveTurns <
-        20 &&
+      myAdventures() - digitizedMonstersRemaining() - globalOptions.saveTurns < 20 &&
```
horrible-little-slime, line 203: Way, way cleaner looking, 👨‍🍳 💋

PR #916, opened 2022-05-19
Add support for fishing with wanderers

pstalcup:


src/wanderer.ts
```diff
@@ -200,15 +215,66 @@ function guzzlrAbandonQuest() {
...
   }
 }
 
+// NOTE: This list is in Priority Order as well.
+const ALL_FISH = ["carp", "cod", "hatchetfish", "bass", "trout", "tuna"] as const;
+type FishTuple = typeof ALL_FISH;
+type Fish = FishTuple[number];
+
+function fishLocations(): { [fish in Fish]: Location | null } {
+  const locations: { [fish in Fish]: Location | null } = {
```
horrible-little-slime, line 233: Feels like we should cache this, if only to avoid repeated server hits

src/wanderer.ts
```diff
@@ -200,15 +215,66 @@ function guzzlrAbandonQuest() {
...
+  return locations;
+}
+
+function fishTargets(): WandererTarget[] {
+  if (have($item`Clan VIP Lounge key`)) {
+    return withVIPClan<WandererTarget[]>(() => {
+      const locations = fishLocations();
+      let value = 8;
+      return ALL_FISH.map((fish) => {
+        value -= 1;
```
horrible-little-slime, line 265: I don't like changing the value of `value` in the map function--could you use something more like
```suggestion
      return ALL_FISH.map((fish, index) => {
        const value = 7 - index;
```
?

src/wanderer.ts
```diff
@@ -200,15 +215,66 @@ function guzzlrAbandonQuest() {
...
+        return new WandererTarget(
```
horrible-little-slime, line 266: This structure generally feels more like a for loop than a .map(), but I think that's a lot more wishy washy a thing, and this is fine

src/wanderer.ts
```diff
@@ -277,14 +343,16 @@ const wandererTargets = [
...
-    }
+    },
+    true // allow guzzlr to use unlockers
   ),
   new WandererTarget(
     "Coinspiracy",
     () => realmAvailable("spooky") && get("lovebugsUnlocked"),
     () => $location`The Deep Dark Jungle`,
-    () => 2 // slightly higher value
+    () => 8 // slightly higher value
```
horrible-little-slime, line 353:
```suggestion
    () => 8 // higher than the value of any Fish target, lower than virtually any conceivable Guzzlr
```

horrible-little-slime:
This branch should probably be scuttled, but it should be easy to add this in the new wanderer paradigm

PR #922, opened 2022-05-23
Yachtzee Chaining

Pantocyclus:
Supports yachtzee chaining under the following conditions
1) Requires access to SBB
2) Requires enough organ cleansers/space to chew at least 30 stench jellies (+2 extros if we have replacers, +1 synth if we have the skill but don't have enough turns of the effect), and uses up to 41 spleen total (inclusive of beggin' cologne).


horrible-little-slime:
This is incredibly out of scope, but we should eventually let it pick hobo monkey/cactus bud as relevant

src/index.ts
```diff
@@ -580,7 +583,17 @@ export function main(argString = ""): void {
     withStash(stashItems, () => {
       withVIPClan(() => {
         // 0. diet stuff.
-        if (!globalOptions.noDiet) runDiet();
+        if (globalOptions.noDiet || property.getBoolean("_garboYachtzeeChainCompleted")) {
```
horrible-little-slime, line 586:
```suggestion
        if (globalOptions.noDiet || get("_garboYachtzeeChainCompleted", false)) {
```

src/index.ts
```diff
@@ -580,7 +583,17 @@ export function main(argString = ""): void {
...
+          print("We should not be yachtzee chaining", "red");
+          globalOptions.yachtzeeChain = false;
+        }
+
+        if (
+          !globalOptions.noDiet &&
+          (!globalOptions.yachtzeeChain || property.getBoolean("_garboYachtzeeChainCompleted"))
```
horrible-little-slime, line 593:
```suggestion
          (!globalOptions.yachtzeeChain ||get("_garboYachtzeeChainCompleted", false))
```

src/post.ts
```diff
@@ -48,7 +48,13 @@ function coldMedicineCabinet(): void {
 }
 
 function horseradish(): void {
-  if (getRemainingStomach() > 0 && !globalOptions.noDiet) {
+  if (
+    getRemainingStomach() > 0 &&
+    !(
+      globalOptions.noDiet ||
+      (globalOptions.yachtzeeChain && !get(`"_garboYachtzeeChainCompleted"`))
```
horrible-little-slime, line 55:
```suggestion
      (globalOptions.yachtzeeChain && !get(`"_garboYachtzeeChainCompleted", false`))
```

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+import { runDiet } from "./diet";
+import { EmbezzlerFight, embezzlerSources, estimatedTurns } from "./embezzler";
+import { hasMonsterReplacers } from "./extrovermectin";
+import { baseMeat, globalOptions, safeRestore, turnsToNC } from "./lib";
+import { meatMood } from "./mood";
+import { farmingPotions, mutuallyExclusive, Potion, potionSetup } from "./potions";
+import { garboValue } from "./session";
+import synthesize from "./synthesis";
+
+class DietEntry<T> {
```
horrible-little-slime, line 59: I am extremely curious why you're setting this up here instead of in `diet.ts` and exporting from there.

Is there a compelling reason to repeat ourselves like this?

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+      song !== $effect`Ode to Booze` &&
+      song !== $effect`Polka of Plenty` &&
+      song !== $effect`Chorale of Companionship` &&
+      song !== $effect`The Ballad of Richie Thingfinder`
+    ) {
+      cliExecute(`shrug ${song}`);
+    }
+  }
+  // Shrug default Mood songs
+  cliExecute(`shrug ur-kel`);
```
horrible-little-slime, line 190: My attitude would be that we should only shrug as many songs as necessary to cast ode; this seems like it could plausibly be going overboard on the number of shrugs.

Probably doesn't matter that much, though.

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+  cliExecute(`shrug phat loot`);
+
+  while (haveEffect($effect`Ode to Booze`) < turns) {
+    useSkill($skill`The Ode to Booze`);
+  }
+  return true;
+}
+
+function executeNextDietStep(stopBeforeJellies?: boolean): void {
+  if (property.getBoolean("stenchJellyUsed")) return;
```
horrible-little-slime, line 200:
```suggestion
  if (get("stenchJellyUsed", false)) return;
```

Is there a reason this isnt' a _ property?

We should maybe add native mafia tracking for this

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+  print("Executing next diet steps", "blue");
+  const dietUtil = new DietUtils();
+  dietUtil.resetDietPref();
+
+  const dietString = property.getString("_garboYachtzeeChainDiet").split(",");
```
horrible-little-slime, line 205: As you've probably noticed, in general, `get` with a default value is better than `property.getType`
This feels meaningfully different, though, for reasons.

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+        }
+      });
+    } else {
+      dietUtil.addToPref(1, name);
+    }
+  }
+  dietUtil.setDietPref();
+
+  if (!stenchJellyConsumed) {
+    throw "We completed our entire diet but failed to get a stench jelly charge";
```
horrible-little-slime, line 245:
```suggestion
    throw new Error("We completed our entire diet but failed to get a stench jelly charge");
```

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+  let fullness = myFullness();
+  let drunkenness = myInebriety();
+  let spleenUse = mySpleenUse();
+  for (const entry of dietSchedule) {
+    fullness += entry.fullness;
+    drunkenness += entry.drunkenness;
+    spleenUse += entry.spleen;
+    if (fullness > fullnessLimit()) throw "Error in diet schedule: Overeating";
+    else if (drunkenness > inebrietyLimit()) throw "Error in diet schedule: Overdrinking";
+    else if (spleenUse > spleenLimit()) throw "Error in diet schedule: Overuse of spleen";
```
horrible-little-slime, line 318:
```suggestion
    if (fullness > fullnessLimit()) throw new Error("Error in diet schedule: Overeating");
    else if (drunkenness > inebrietyLimit()) throw new Error("Error in diet schedule: Overdrinking");
    else if (spleenUse > spleenLimit()) throw new Error("Error in diet schedule: Overuse of spleen");
```

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+  }
+
+  return dietSchedule;
+}
+
+export function yachtzeeChainDiet(simOnly?: boolean): boolean {
+  if (property.getBoolean("_garboYachtzeeChainDietPlanned")) return true;
```
horrible-little-slime, line 325:
```suggestion
  if (get("_garboYachtzeeChainDietPlanned", false)) return true;
```

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+
+  // Plan for Yachtzee Chain
+  // 1) Fish Juice Box + Fishy Pipe for 30 turns of Fishy and Really Deep Breath (so we can ignore underwater gear)
+  // 2) PYEC to extend Fishy + Really Deep Breath if possible
+  // 3) Chew 30 (or 35 with PYEC) Stench Jellies (ensure that we have enough organ space)
+  // 4) Find meat and famwt buff that makes sense for a 2k base drop
+  // 5) Plant underwater friar's plant if possible
+
+  const havePYECCharge = have($item`Platinum Yendorian Express Card`) && !get(`expressCardUsed`);
```
horrible-little-slime, line 334: Most players will be getting a PYEC from their clan--is there a way for us to check that as well?

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+  );
+  const embezzlerValue = embezzlerTurns > 0 ? potion.gross(embezzlerTurns) : 0;
+  const yachtzeeValue =
+    (effectiveYachtzeeTurns * 2000 * (potion.meatDrop() + 2.5 * potion.familiarWeight())) / 100; // Every 1lbs of lep ~ 2.5% meat drop
+  const barfValue = (barfTurns * baseMeat * turnsToNC) / (turnsToNC + 1);
+
+  return yachtzeeValue + embezzlerValue + barfValue - potion.price(true);
+}
+
+function yachtzeePotionSetup(yachtzeeTurns: number, simOnly?: boolean): number {
```
horrible-little-slime, line 603: I feel like it might be easier to put this code in `potion.ts`, or at least to make some bigger changes to `potion.ts` and let this code be a little smaller.

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+  let totalProfits = 0;
+  const PYECOffset =
+    have($item`Platinum Yendorian Express Card`) && !get(`expressCardUsed`) ? 5 : 0;
+  const excludedEffects = new Set<Effect>();
+
+  if (have($item`Eight Days a Week Pill Keeper`) && !property.getBoolean("_freePillKeeperUsed")) {
```
horrible-little-slime, line 609:
```suggestion
  if (have($item`Eight Days a Week Pill Keeper`) && !get("_freePillKeeperUsed")) {
```

This one's already a native mafia property! use regular get!

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+      cliExecute("florist plant Crookweed");
+    }
+    plantCrookweed = false;
+  }
+  set("choiceAdventure918", "");
+}
+
+export function yachtzeeChain(): void {
+  if (!globalOptions.yachtzeeChain) return;
+  else if (property.getBoolean("_garboYachtzeeChainCompleted")) return;
```
horrible-little-slime, line 823:
```suggestion
if get("_garboYachtzeeChainCompleted", false)) return;
```

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+  else if (myLevel() <= 13 || !canInteract()) return;
+  // We definitely need to be able to eat sliders and drink pickle juice
+  else if (property.getBoolean("fishyPipeUsed") && !have($effect`Fishy`)) return;
+  // If we have used our fishy pipe and have no fishy turns left, we're probably done
+  else if (
+    !property.getBoolean("sleazeAirportAlways") &&
+    !property.getBoolean("_sleazeAirportToday")
+  ) {
+    return;
+  }
```
horrible-little-slime, line 769:
```suggestion
  if (!have($item`fishy pipe`)) return;
  // hard require fishy pipe to run this chain
  if (!have($familiar`Urchin Urchin`)) return;
  // also hard require urchin urchin for now
   if (myLevel() <= 13 || !canInteract()) return;
  // We definitely need to be able to eat sliders and drink pickle juice
  if (get("fishyPipeUsed") && !have($effect`Fishy`)) return;
  // If we have used our fishy pipe and have no fishy turns left, we're probably done
  if (!realmAvailable("sleaze")) return;
```

You'll need to import `realmAvailable`, but that's fine and valid

horrible-little-slime: I also feel like if you can't interact we should just categorically not work right now, but that's another bridge to cross

src/yachtzee.ts
```diff
@@ -0,0 +1,831 @@
...
+  let turncount = myTurncount();
+  yachtzeePotionSetup(Math.min(jellyTurns, fishyTurns));
+  cliExecute(`closet take ${myClosetMeat()} meat`);
+  safeRestore();
+
+  let plantCrookweed = true;
+  set("choiceAdventure918", 2);
+  while (Math.min(jellyTurns, fishyTurns) > 0) {
+    executeNextDietStep();
+    if (!property.getBoolean("stenchJellyUsed")) throw "We did not use stench jellies";
```
horrible-little-slime, line 800:
```suggestion
    if (!("stenchJellyUsed", false)) throw new Error("We did not use stench jellies");
```

horrible-little-slime (changes requested):
I'm going to need someone else to look at the diet code, but I think we're generally on the right track. I'm concerned by hardcoding the juice boxes and things like that, but that's fine on a first pass probably.

Stylistically there are a few things inconsistent with the rest of garbo--never using regular `get`, throwing strings as opposed to god-fearing, american errors (I'll make sure to convert any current strings to errors in a bit), but those shouldn't be hard to catch.

I have complicated feelings about the amount of code that's concentrated in yachtzee.ts, and I feel like some of the potion and diet code should move to their respective files for both legibility here and for making our lives easier in future updates. I'm not sure how much of an ask that is--if there's a big structural reason why you can't do that, we can re-evaluate.

src/yachtzee.ts
Pantocyclus: Oh huh so Clara's Bell probably lasts across RO, but stench jelly doesn't. Makes sense to make it a _ property then.
I do think it's probably helpful to have it as a Mafia property, but it sounds like quite a bit of work differentiating between NCs that use a jelly charge and those that don't. Also, jelly charges can be stacked, so it should probably be a numeric value instead of a boolean in that case.

src/index.ts
Pantocyclus: I was in fact using `get` prior to this, but it seemed to be incorrectly parsing _ prefs for some reason. Forcing it with `getBoolean` seemed to resolve the issue for me - I can't really figure out what the underlying cause of the parsing error was

src/yachtzee.ts
Pantocyclus: Practically this makes no difference, since we always end up running Chorale, Polka and Ballad, with 1 slot left over for Ode casts between stench jellies

src/yachtzee.ts
Pantocyclus: We can probably try to pull 1 from the stash at the start if we have none on hand

src/yachtzee.ts
Pantocyclus: The thing about `potion.ts` is that it is highly specific to embezzlers (as the function arguments suggest), and there seems to be no way to fix that without a major overhaul.
Likewise, this piece of code here is extremely specific to yachtzee chains (as the function arguments also suggest). It could definitely be placed in `potion.ts` but its lack of generality means that it would never be imported/used by any other methods

src/yachtzee.ts
Pantocyclus: I feel like there are some extremely large assumptions made here that are not generally applicable, the key one being that we can always solve the scheduling problem greedily - this is specific to chaining since we necessitate the excessive use of stench jellies which are of size 1.

I do acknowledge the notion that dieting functionalities should go in `diet.ts`, but I'm also of the conflicting opinion that yachtzee-specific functionalities (that aren't general and will not be used elsewhere) should go in `yachtzee.ts`.

Structurally, there's no difficulty porting the functions between files, so I'm ok with supporting either case.

horrible-little-slime: I guess my major concern here is that `DietEntry` is the name of a class in the libram diet support, and this is similar-but-not-identical to it. That feels potentially confusing. 

While I know a lot of the consumption here can be done greedily, it feels like combining this with our knapsack code might allow us to better balance potion-foods with additional yachtzees. That feels like a bit too much scope creep, then.

src/yachtzee.ts
```diff
@@ -331,13 +334,14 @@ export function yachtzeeChainDiet(simOnly?: boolean): boolean {
   // 4) Find meat and famwt buff that makes sense for a 2k base drop
   // 5) Plant underwater friar's plant if possible
 
-  const havePYECCharge = have($item`Platinum Yendorian Express Card`) && !get(`expressCardUsed`);
+  const havePYECCharge =
+    have($item`Platinum Yendorian Express Card`) && !get(`expressCardUsed`, false);
```
horrible-little-slime, line 338: You shouldn't have to specify a default value for `get` with native mafia preferences. I'm very curious what's causing your weird get issues.

src/yachtzee.ts
horrible-little-slime: Fair enough!

src/yachtzee.ts
Pantocyclus: Tying it in with the knapsack code is probably way out of scope imo. 
I haven't looked too deeply at the libram diet code yet, so I'm unsure whether it can currently handle balancing fishy turns with jelly charges, with consideration given to PYEC and fishy pipe as additional sources of fishy (both non-diet sources). The diet also needs to consider that using a fish juice box provides extra meat% for 30 turns since we do not need an alternative source of underwater breathing. 
That's a bunch of discrete optimisation to be conducted over multiple step functions, which all act as a soft limit (we can always get excess turns of fishy or really deep breath - they just contribute 0 profits, unlike organ fillers which are hard limits; I don't know if the knapsack code currently handles soft limits optimally).
Additionally, the diet uses marginal VOA in its optimisation, which is currently fixed as a constant by the user. However, there could be a 750MPA difference depending on how many jellies we huff (and thus how many turns of synth we can generate for the rest of embezzlies + barf). We would need to dynamically alter the user's VOA, which was originally an invariant (the user-defined VOA should have already accounted for marginal synths in a nominal garbo run, but this assumption no longer holds when we are enumerating over different amounts of forced non-synth spleen usage), to get accurate profit estimates.

Pantocyclus: I could probably change the name `DietEntry`, but I had specifically crafted it to allow for the ease of using `.action()` for both simulation (recording the scheduled diet in the prefs) and execution (consuming and also setting jelly usage which is not tracked by mafia) by simply overriding the function. For sure, I could
1) Separate them into `.simulation()` and `.consume()` methods, and;
2a) Add a PR to libram, but once again I'm not sure if such a functionality would be generally relevant, or;
2b) Extend the `DietEntry` class to a derived `YachtzeeDietEntry` in `yachtzee.ts`, which also seems weird since I'm not really inheriting all that much from the base class.

These are all stylistic concerns, not technical ones, so if there's a general concensus / strong inclination towards a given implementation it shouldn't be too big of an issue to port things over.

src/yachtzee.ts
Pantocyclus: Unlike garbo's nominal behaviour which is not dependent on having stash access, our chaining diet heavily depends on whether we have access to a PYEC or not. I see 2 ways of implementing this.
1. We wrap `_yachtzeeChain()` in the scope of `withStash()` if we do not own a PYEC, but that requires holding on to the PYEC until we finish chaining (which honestly isn't that long), or;
2. We check if we have access to the PYEC prior to running `_yachtzeeChain()`, and set a global/mafia pref. This means that we only grab the PYEC when we need it. I'm not sure how I should parse `clan_stash.php` at this moment, as I don't know the exact details of how karma and stash restrictions interact with stash pulls. The simple assumption is that if I can pull a PYEC once, I can pull it infinitely many times per day (which is probably a fair assumption if the clan places enough trust in a member to not abscond with the PYEC in the first pull). Thus, we attempt to pull and immediately return the PYEC, and setting a _pref stating if we have access to a PYEC.

I've already pushed a commit for the 2nd implementation.

src/yachtzee.ts
Pantocyclus: libramGetWeirdness
Managed to get a screengrab of the `get` issue with _customPrefs, just in case you have an inkling of what might be going on

src/yachtzee.ts
```diff
@@ -1117,6 +1117,9 @@ function _yachtzeeChain(): void {
   let turncount = myTurncount();
   yachtzeePotionSetup(Math.min(jellyTurns, fishyTurns));
   cliExecute(`closet take ${myClosetMeat()} meat`);
+  if (haveEffect($effect`Beaten Up`)) {
+    cliExecute("hottub");
```
horrible-little-slime, line 1121: if you aren't able to hottub, we should uneffect beaten up in some other way, using libram's uneffect()

Pantocyclus: Resolved in latest commit

horrible-little-slime (approved):
In the future, I would use a little more `git rebase --interactive` and `git commit --amend`

PR #930, opened 2022-05-26
new libram;use robort for reminisce monsters

horrible-little-slime:


src/dailies.ts
```diff
@@ -336,9 +336,7 @@ function prepFamiliars(): void {
         continue;
       }
       withProperty("autoBuyPriceLimit", priceCap, () => retrieveItem(1, drink));
-      if (have(drink)) {
-        visitUrl(`inventory.php?action=robooze&which=1&whichitem=${toInt(drink)}`);
-      }
+      if (have(drink)) Robortender.feed(drink);
```
docrostov, line 339: i am trying to track back through the code and follow the processing but i am unsure how, exactly, we attempt to feed the robort a drink again after we reminisce. this feels like a "maybe we solve it for your next run of garbo with a reminisce but not for the current run" solution. and it doesn't even -really- do that, because if the price of fish heads is above market cap, it won't ever retrieve item and convert them to drive-bys. i think you can solve this by just coercing garbo to always convert fish heads to drive-bys; it's slightly annoying behavior for some people i guess but i think it would at least solve this issue

docrostov: okay talked to you over DM, now i get the first part. one small thing we could do here is to call prep familiars twice -- once as part of dailies and once as your post-free-fight setup, in case you now have a fish head and can actually give it to the robort. it's not perfect but it's better than a "fix for tomorrow" solution i think.

src/fights.ts
```diff
@@ -1282,7 +1282,7 @@ const freeFightSources = [
       CombatLoversLocket.reminisce(monster);
     },
     true,
-    { canOverrideMacro: true }
+    { canOverrideMacro: true, familiar: () => $familiars`Robortender`.find(have) ?? null }
```
docrostov, line 1285: one small concern here -- does this mean that we're going to potentially ruin our 100% robort drop by forcing robort on a non-black crayon penguin monster? i don't think so but it is probably worth thinking about maybe.

src/fights.ts
```diff
@@ -2177,7 +2177,10 @@ export function printEmbezzlerLog(): void {
 
 const isFree = (monster: Monster) => monster.attributes.includes("FREE");
 const valueDrops = (monster: Monster) =>
-  sumNumbers(itemDropsArray(monster).map(({ drop, rate }) => garboValue(drop) * rate));
+  sumNumbers(itemDropsArray(monster).map(({ drop, rate }) => garboValue(drop) * rate)) +
+  (have($familiar`Robortender`)
+    ? Robortender.dropChance() * garboValue(Robortender.dropFrom(monster))
```
docrostov, line 2182: you were saying that when you tested it didn't fight a penguin. two followups for you via this code:

- we've fully validated that dropChance & garbo value are playing nicely in multiplication, right? could see one being an int and the other being a float and some stupid bullshit happening with rounding if it's not a 100% chance maybe.
- do you have a black crayon penguin in your locket? i assume yes. i just realized i actually don't, so that is on my list of things to do today too lol

might be worth printing this for testing when it tabulates it just so you can have a reference for what garbo thinks it's worth just in case there's a valuations issue here...

src/fights.ts
horrible-little-slime: It does mean this; I can do a more exhaustive fix later

src/fights.ts
horrible-little-slime: js doesn't have ints. I'll test it again on casual leg

docrostov: thank you for resolving this. i am proud of myself that i had identified something that was adjacent to the actual problem even though it extremely was not the actual problem. i needed a win today.

src/fights.ts
```diff
@@ -2174,10 +2180,83 @@ export function printEmbezzlerLog(): void {
...
+const freeKills: FreeKill[] = [
+  {
+    source: $item`The Jokester's gun`,
+    skill: $skill`Fire the Jokester's Gun`,
+    used: () => get("_firedJokestersGun"),
+  },
+  {
+    source: $item`Lil' Doctor™ bag`,
+    skill: $skill`Chest X-Ray`,
+    used: () => get("_chestXRayUsed") < 3,
```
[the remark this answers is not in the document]
docrostov: also this one is funny because i know for a fact neither of us have it yet but should we include batoomerang here :garf:

src/fights.ts
```diff
@@ -2174,10 +2180,83 @@ export function printEmbezzlerLog(): void {
     HIGHLIGHT
   );
 }
+type FreeKill = { source?: Item; skill: Skill; used: () => boolean };
+const freeKills: FreeKill[] = [
```
docrostov, line 2184: this is funny to me but i think we actually need to add batoomerang here. this is funny to me because neither of us actually have a batoomerang so i'm not even sure we can test it but if we're including jokester's gun we probably gotta include that as a free fight source for, i don't know, James Spader, returning to kol after a brief hiatus spent filming The Blacklist with their only free fight source being their batoomerang.

src/fights.ts
```diff
@@ -2174,10 +2180,83 @@ export function printEmbezzlerLog(): void {
...
+  while (freeKill && CombatLoversLocket.have() && CombatLoversLocket.reminiscesLeft()) {
+    const roboTarget = CombatLoversLocket.findMonster(
+      () => true,
+      (monster: Monster) =>
+        valueDrops(monster) + garboValue(Robortender.dropFrom(monster)) * Robortender.dropChance()
+    );
+
+    if (!roboTarget) break;
+    const regularTarget = CombatLoversLocket.findMonster(() => true, valueDrops);
+    if (regularTarget === roboTarget) {
```
docrostov, line 2241: wait shouldn't this be if regularTarget != roboTarget? if it equals a roboTarget wouldn't we want to use robort and not the free fight fam lol

src/fights.ts
```diff
@@ -2174,10 +2180,83 @@ export function printEmbezzlerLog(): void {
...
+    freeFightOutfit(toRequirement(freeKill));
+    withMacro(
+      Macro.skill(freeKill.skill),
+      () => {
+        mapMonster($location`The Copperhead Club`, $monster`Mob Penguin Capo`);
+        runCombat();
+      },
+      true
+    );
+    freeKill = findFreeKill();
```
docrostov, line 2229: wait why are we doing this at the end again? is it just like... retabulating within the while to ensure the while is still active? that makes sense i guess. for some reason i would've thought the while logic would retabulate when it hits the end naturally but i am probably wrong.

src/fights.ts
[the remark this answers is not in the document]
horrible-little-slime: I have bat-oomerang

src/fights.ts
horrible-little-slime: We initialize it before the first while, but update it after each subsequent while. We want to update it before the condition is called because we don't want to proceed if there is no available free fight now

lewismd13 (dismissed):

horrible-little-slime:
I plan to merge this tomorrow because I don't have time to support any errors today

PR #944, opened 2022-06-08
allow moods to execute using mp restore items

horrible-little-slime:
Right now, moods will only restore mp to execute using sausages or oscus's soda.

For people without kramco, this is less than ideal.


horrible-little-slime:
Funnier and more elegant fix incoming

PR #951, opened 2022-06-11
Unset recovery script when we are equipping June cleaver

Pantocyclus:
We reset the recovery script in safeRestore after handling potentially being beaten up


horrible-little-slime:
Gone, but not forgotten; #952 takes precedent

PR #962, opened 2022-06-19
tentative revamp of free fight familiar code

horrible-little-slime:
The purpose here is to make it so we can expose the EV of a free fight familiar, which will be useful for robort shenanigans and panto's plan moving forward

### DIFF
--- a/src/familiar.ts
+++ b/src/familiar.ts
@@ -241,6 +241,13 @@ const standardFamiliars: () => GeneralFamiliar[] = () => [
         (familiarWeight($familiar`Stocking Mimic`) + weightAdjustment()),
   },
   {
+    familiar: $familiar`Shorter-Order Cook`,
+    expectedValue:
+      garboAverageValue(
+        ...$items`short beer, short stack of pancakes, short stick of butter, short glass of water, short white`
+      ) / 11,
+  },
+  {
     familiar: $familiar`Robortender`,
     expectedValue: 200,
   },
--- a/src/familiar.ts
+++ b/src/familiar.ts
@@ -253,7 +253,7 @@ const standardFamiliars: () => GeneralFamiliar[] = () => [
   },
   ...$familiars`Hobo Monkey, Cat Burglar, Urchin Urchin, Leprechaun`.map((familiar) => ({
     familiar,
-    expectedValue: 1,
+    expectedValue: 0,
   })),
   {
     familiar: $familiar`none`,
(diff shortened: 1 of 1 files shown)


src/familiar.ts
```diff
@@ -55,177 +53,237 @@ export function meatFamiliar(): Familiar {
...
+};
+
+function valueStandardDropFamiliar({
+  familiar,
+  expected,
+  drop,
+  pref,
+  additionalValue,
+}: StandardDropFamiliar): GeneralFamiliar | null {
+  const expectedTurns = expected[get(pref)];
```
Pantocyclus, line 76: If we get more than `expected.length` drops this would overflow the memory
```suggestion
  const expectedTurns = expected[Math.min(expected.length - 1, get(pref))];
```

src/familiar.ts
```diff
@@ -55,177 +53,237 @@ export function meatFamiliar(): Familiar {
...
+  if (!have(familiar) || !expectedTurns) return null;
+  const expectedValue = garboValue(drop) / expectedTurns + (additionalValue ?? 0);
```
Pantocyclus, line 78: `expectedValue` here seems to be the expected drop value of a spleener, but a `GeneralFamiliar` only has the field `expectedValue` and not `additionalValue`.

This seems weird, since a general leprechaun should have non-zero expected value (at the very least, it should have a different value from a non-lep).

I prefer the idea of having the following fields instead:
`baseValue`: or synonymously, `lepValue`
`dropValue`: `garboValue(drop) / expectedTurns`
`additionalValue`: For meatify, drones, spit, red present and whatnots (even item% if that ever becomes relevant)
`expectedValue`: `baseValue + dropValue + additionalValue`

src/familiar.ts
horrible-little-slime: This is not a necessary change; line 77 checks to see if expectedTurns is defined. Overflowing an array returns undefined.

src/familiar.ts
horrible-little-slime: I think I should cut this down to 
```
familiar: Familiar,
leprechaunMultiplier: number,
additionalValue: number,
```
where additionalValue combines the EV of drops of all sorts. I don't particularly see the value in caching the outfit value as part of the general familiar (I would cache that elsewhere, keyed by leprechaunMultiplier), and `leprechaunMultiplier` should capture all of that lep information you're looking for

horrible-little-slime: that being said, the leprechaunMultiplier function lets us derive that from the Familiar itself. But I suppose it's still worth saving as a property of the familiar

PR #964, opened 2022-06-21
devalue june cleaver for embezzlies

horrible-little-slime:


Tokoeka (approved):
Looks Good. Just curious, Is there a specfic reason you chose 30 as the interval length to tell garbo for EV calcs when embezzlers are up?

horrible-little-slime:
> Looks Good. Just curious, Is there a specfic reason you chose 30 as the interval length to tell garbo for EV calcs when embezzlers are up?

That's the endgame interval

PR #965, opened 2022-06-21
Add failsafe to equipping underwater breathing gear for familiars

Pantocyclus:


Tokoeka (dismissed):

src/yachtzee.ts
```diff
@@ -1042,28 +1021,46 @@ export function bestYachtzeeFamiliar(): Familiar {
...
-        (right.underwater && have($item`amulet coin`) ? 50 : 0) -
-        leprechaunMeatBonus(
-          findLeprechaunMultiplier(left) *
-            (famWt +
-              (left.underwater ? (have($item`amulet coin`) ? 10 : 5) : famEquipWeightPenalty))
+        numericModifier(
+          right,
+          "meat drop",
+          famWt,
+          right.underwater || have($effect`Driving Waterproofly`) || have($effect`Wet Willied`)
```
horrible-little-slime, line 1028: Do not do this logic in the middle of your numericModifier call, I would make this either be the output of a more destructured function or at the very least name a const

src/yachtzee.ts
```diff
@@ -1042,28 +1021,46 @@ export function bestYachtzeeFamiliar(): Familiar {
...
-      `${fam} (${(
-        leprechaunMeatBonus(
-          findLeprechaunMultiplier(fam) *
-            (famWt + (fam.underwater ? (have($item`amulet coin`) ? 10 : 5) : famEquipWeightPenalty))
-        ) + (fam.underwater && have($item`amulet coin`) ? 50 : 0)
+      `${fam} (${numericModifier(
+        fam,
+        "meat drop",
+        famWt,
+        fam.underwater || have($effect`Driving Waterproofly`) || have($effect`Wet Willied`)
```
horrible-little-slime, line 1057: I am going to become the joker

src/yachtzee.ts
```diff
@@ -1042,28 +1021,46 @@ export function bestYachtzeeFamiliar(): Familiar {
...
-              (right.underwater ? (have($item`amulet coin`) ? 10 : 5) : famEquipWeightPenalty))
-        ) +
-        (right.underwater && have($item`amulet coin`) ? 50 : 0) -
-        leprechaunMeatBonus(
-          findLeprechaunMultiplier(left) *
-            (famWt +
-              (left.underwater ? (have($item`amulet coin`) ? 10 : 5) : famEquipWeightPenalty))
+        numericModifier(
+          right,
+          "meat drop",
```
horrible-little-slime, line 1026: capitalize Meat Drop, just for consistency with the Modifiers type

src/yachtzee.ts
```diff
@@ -1003,34 +1003,13 @@ function yachtzeePotionSetup(yachtzeeTurns: number, simOnly?: boolean): number {
...
 }
 
-function leprechaunMeatBonus(wt: number): number {
-  return 2 * wt + Math.sqrt(220 * wt) - 6;
-}
-
 export function bestYachtzeeFamiliar(): Familiar {
   const haveUnderwaterFamEquipment = familiarWaterBreathingEquipment.some((item) => have(item));
   const famWt =
     familiarWeight(myFamiliar()) +
```
horrible-little-slime, line 1009: is this a correct estimate? does it matter?

src/yachtzee.ts
```diff
@@ -1042,28 +1021,46 @@ export function bestYachtzeeFamiliar(): Familiar {
     )
     .sort(
```
horrible-little-slime, line 1022: is there a reason we aren't sorting by leprechaunMultiplier? it seems easier. I know that the cactus bud slightly messes with it, but frankly its effect on the sqrt term is small enough at the weights we're dealing with.
It's alright to use numericModifier, but you for sure want to slice this out into a function because you are repeating yourself a lot

horrible-little-slime (changes requested):

src/yachtzee.ts
Pantocyclus: As asmo has pointed out, this could go up to a 50%meat difference. That's 1k/spleen, which could potentially mean the difference between gregging and chaining, so the term does matter to some degree\*.

\*The estimates aren't *that* accurate yet - the current estimate slightly underestimates our yachtzee profitability, but it could potentially matter when we refine it in the future.

src/yachtzee.ts
```diff
@@ -1003,8 +1003,15 @@ function yachtzeePotionSetup(yachtzeeTurns: number, simOnly?: boolean): number {
   return totalProfits;
 }
 
-function leprechaunMeatBonus(wt: number): number {
-  return 2 * wt + Math.sqrt(220 * wt) - 6;
+function bestFamUnderwaterGear(fam: Familiar): Item {
```
horrible-little-slime, line 1006: This is fine now that it's sliced out, but in the future, we try not to have ternary chains that are _that_ long. A nicer structure for this would be to use nested if statements and several returns.

horrible-little-slime: Will for sure not block a merge, though

horrible-little-slime (approved):

PR #966, opened 2022-06-21
Marginal familiars revisited

Pantocyclus:
Leveraging off the draft revision of `familiar.ts` by phred

### DIFF
--- a/src/familiar.ts
+++ b/src/familiar.ts
@@ -483,55 +483,68 @@ export function setMarginalFamiliar(loc: Location): void {
       return a < b ? a : b;
     });
 
+  let nonJellyExpectedAdv = 0;
   const meatFam = meatFamiliar();
   const meatFamEV = sum(dropFamiliars, (fam) => {
     return fam.familiar === meatFam ? fam.marginalValue - nominalOutfitValue : 0;
   });
   const shouldRunDropFams =
-    sum(
-      dropFamiliars.map((fam) => {
-        const rotFam = rotatingFamiliars.reduce((a, b) => (a.familiar === fam.familiar ? a : b));
-        const { meat, bonus } =
-          cachedOutfits.get(fam.leprechaunMultiplier) ??
-          cacheOutfit(fam.leprechaunMultiplier, fam.familiar);
+    Math.round(
+      sum(
+        dropFamiliars.map((fam) => {
+          const rotFam = rotatingFamiliars.reduce((a, b) => (a.familiar === fam.familiar ? a : b));
+          const { meat, bonus } =
+            cachedOutfits.get(fam.leprechaunMultiplier) ??
+            cacheOutfit(fam.leprechaunMultiplier, fam.familiar);
 
-        const outfitValue = (meat * locBaseMeat) / 100 + bonus - nominalOutfitValue;
-        if (rotFam.familiar === fam.familiar) {
-          const pIdx = get(rotFam.pref);
-          const dropVal = garboValue(rotFam.drop);
-          let expectedAdventures = 0;
-          for (let idx = pIdx; idx < rotFam.expected.length; idx++) {
-            if (dropVal / rotFam.expected[idx] + outfitValue > meatFamEV) {
-              expectedAdventures += rotFam.expected[idx];
-            } else break;
+          const outfitValue = (meat * locBaseMeat) / 100 + bonus - nominalOutfitValue;
+          if (rotFam.familiar === fam.familiar) {
+            const pIdx = get(rotFam.pref);
+            const dropVal = garboValue(rotFam.drop);
+            let expectedAdventures = 0;
+            for (let idx = pIdx; idx < rotFam.expected.length; idx++) {
+              if (dropVal / rotFam.expected[idx] + outfitValue > meatFamEV) {
+                expectedAdventures += rotFam.expected[idx];
+              } else break;
+            }
+            nonJellyExpectedAdv += expectedAdventures;
+            return expectedAdventures;
+          } else if (fam.familiar === $familiar`Space Jellyfish`) {
+            const pIdx = get("_spaceJellyfishDrops");
+            const dropVal =
+              barf && myInebriety() <= inebrietyLimit() ? garboValue($item`stench jelly`) : 0;
+            let expectedAdventures = 0;
+            for (let idx = pIdx; idx < 5; idx++) {
+              if (dropVal / (idx + 1) + outfitValue > meatFamEV) {
+                expectedAdventures += idx + 1;
+              } else break;
+            }
+            if (dropVal / 20 + outfitValue > meatFamEV) expectedAdventures += Infinity;
+            return expectedAdventures;
+          } else {
+            return 0;
           }
-          return expectedAdventures;
-        } else if (fam.familiar === $familiar`Space Jellyfish`) {
-          const pIdx = get("_spaceJellyfishDrops");
-          const dropVal =
-            barf && myInebriety() <= inebrietyLimit() ? garboValue($item`stench jelly`) : 0;
-          let expectedAdventures = 0;
-          for (let idx = pIdx; idx < 5; idx++) {
-            if (dropVal / (idx + 1) + outfitValue > meatFamEV) {
-              expectedAdventures += idx + 1;
-            } else break;
-          }
-          if (dropVal / 20 + outfitValue > meatFamEV) expectedAdventures += Infinity;
-          return expectedAdventures;
-        } else {
-          return 0;
+        }),
+        (val) => {
+          return val;
         }
-      }),
-      (val) => {
-        return val;
-      }
+      )
     ) >=
     myAdventures() - globalOptions.saveTurns;
 
   if (shouldRunDropFams) {
-    if (dropFamiliars[0].familiar !== meatFamiliar()) {
-      const dropFam = dropFamiliars[0].familiar;
-      const dropFamEV = (dropFamiliars[0].marginalValue - nominalOutfitValue).toFixed(2);
+    const idx =
+      globalOptions.ascending &&
+      nonJellyExpectedAdv < 30 && // Assume 30 overdrunk barf adventures
+      myInebriety() <= inebrietyLimit() &&
+      dropFamiliars[idx].familiar !== meatFamiliar()
+        ? dropFamiliars.findIndex((fam) => {
+            fam.familiar === $familiar`Space Jellyfish`;
+          }) ?? 0
+        : 0;
+    if (dropFamiliars[idx].familiar !== meatFamiliar()) {
+      const dropFam = dropFamiliars[idx].familiar;
+      const dropFamEV = (dropFamiliars[idx].marginalValue - nominalOutfitValue).toFixed(2);
       print(
         `Determined that ${dropFam} (EV: ${dropFamEV})} is better than your meat familiar ${meatFam} (EV: ${meatFamEV.toFixed(
           2
@@ -543,6 +556,6 @@ export function setMarginalFamiliar(loc: Location): void {
         "blue"
       );
     }
-    useFamiliar(dropFamiliars[0].familiar);
+    useFamiliar(dropFamiliars[idx].familiar);
   } else useFamiliar(meatFam);
 }
--- a/src/familiar.ts
+++ b/src/familiar.ts
@@ -538,6 +538,7 @@ export function setMarginalFamiliar(loc: Location): void {
         nominalOutfitValue ?? 0;
     const idx =
       globalOptions.ascending &&
+      have($item`Drunkula's wineglass`) &&
       nonJellyExpectedAdv < 30 && // Assume 30 overdrunk barf adventures
       myInebriety() <= inebrietyLimit() &&
       dropFamiliars[0].familiar !== meatFamiliar()
--- a/src/familiar.ts
+++ b/src/familiar.ts
@@ -540,7 +540,6 @@ export function setMarginalFamiliar(loc: Location): void {
       globalOptions.ascending &&
       nonJellyExpectedAdv < 30 && // Assume 30 overdrunk barf adventures
       myInebriety() <= inebrietyLimit() &&
-      jellyfishEV > meatFamEV &&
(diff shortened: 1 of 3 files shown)


src/familiar.ts
```diff
@@ -281,3 +363,149 @@ export function timeToMeatify(): boolean {
...
+  }
+
+  const barf = loc === $location`Barf Mountain`;
+  const dropFamiliars = rotatingFamiliars
+    .map((fam): GeneralFamiliar => {
+      return {
+        familiar: fam.familiar,
+        expectedValue: garboValue(fam.drop) / (fam.expected[get(fam.pref)] ?? Infinity),
+        leprechaunMultiplier: findLeprechaunMultiplier(fam.familiar),
+      };
```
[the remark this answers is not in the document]
horrible-little-slime: I can see why you might not want to deal with it, due to its habit of returning null, but you can use the filterNull function I established earlier

src/familiar.ts
```diff
@@ -281,3 +363,149 @@ export function timeToMeatify(): boolean {
...
+      const isLep = fam.leprechaunMultiplier > 0;
+      let additionalValue = 0;
+      if (isLep) {
+        if (!lepOutfitMeatPercent || !lepOutfitFamWeight) {
+          print("Computing leprechaun outfits for the first time...", "blue");
+          useFamiliar(fam.familiar);
+          meatOutfit(false, undefined, false);
+          lepOutfitMeatPercent = 0;
+          lepOutfitFamWeight = 0;
+          for (const slot of $slots`hat, back, shirt, weapon, off-hand, pants, acc1, acc2, acc3, familiar`) {
```
horrible-little-slime, line 435: yeah, I would prefer a `sum` to using a `for` loop here. A good rule of thumb is to try to avoid using `let` unless you're caching something.

src/familiar.ts
```diff
@@ -281,3 +363,149 @@ export function timeToMeatify(): boolean {
...
+            nonLepOutfitFamWeight += numericModifier(equippedItem(slot), "Familiar Weight");
+          }
+        }
+        additionalValue = (toInt(barf) * nonLepOutfitMeatPercent * baseMeat) / 100;
+      }
+      return {
+        familiar: fam.familiar,
+        expectedValue: fam.expectedValue,
+        leprechaunMultiplier: fam.leprechaunMultiplier,
+        marginalValue: fam.expectedValue + additionalValue,
```
horrible-little-slime, line 475:
```suggestion
      return {
        ...fam,
        marginalValue: fam.expectedValue + additionalValue
      }
```

horrible-little-slime (changes requested):

src/familiar.ts
```diff
@@ -281,3 +363,149 @@ export function timeToMeatify(): boolean {
...
+        : 0,
+      leprechaunMultiplier: 0,
+    })
+    .filter((fam) => have(fam.familiar) && (!underwater || fam.familiar.underwater))
+    .map((fam): MarginalFamiliar => {
+      const isLep = fam.leprechaunMultiplier > 0;
+      let additionalValue = 0;
+      if (isLep) {
+        if (!lepOutfitMeatPercent || !lepOutfitFamWeight) {
+          print("Computing leprechaun outfits for the first time...", "blue");
```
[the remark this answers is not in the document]
Pantocyclus: It only computes it for one leprechaun and one non-leprechaun, and only once per invocation of garbo. Does not take too long for me in my testing.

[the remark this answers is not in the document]
horrible-little-slime: B)

src/familiar.ts
```diff
@@ -281,3 +363,149 @@ export function timeToMeatify(): boolean {
...
+          useFamiliar(fam.familiar);
+          meatOutfit(false, undefined, false);
```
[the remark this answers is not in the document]
Pantocyclus: Indeed. It doesn't even do one per leprechaun multiplier, but just one outright. Both `lepOutfitMeatPercent` and `lepOutfitFamWeight` are set after the first time it maximizes, so it should never be triggered again.

My plan is to cache the outfit for the highest leprechaun multiplier - this is why I explicitly copied this line of code and pasted it above to trigger specifically for robort (if that is successfully triggered, then this maximizer here wouldn't be called)

src/index.ts
```diff
@@ -240,6 +240,8 @@ function barfTurn() {
       myInebriety() <= inebrietyLimit()
     ) {
       useFamiliar($familiar`Space Jellyfish`);
+    } else {
+      setMarginalFamiliar(location, underwater);
```
horrible-little-slime, line 244: won't underwater never be true for this, given its placement in the else-if chain??

Really, I feel like we only want to do this for barf-turns. I don't want us running a jung man against an embezzler. So it might be worth cutting the underwater parameter entirely.

src/familiar.ts
```diff
@@ -281,3 +368,158 @@ export function timeToMeatify(): boolean {
...
+  print(`passiveSkillWeight: +${passiveSkillWeight}lbs`, "blue");
+  print(`totalAdditionalWeight: +${buffFamWeight}lbs`, "blue");
+
+  // Ignore free fight equips in this meat outfit valuation
+  const noFreeFightsReq = new Requirement([], {
+    preventEquip: $items`Kramco Sausage-o-Matic™, cursed magnifying glass, protonic accelerator pack, "I Voted!" sticker`,
+  });
+
+  // Compute optimal outfit for robort (rather than a generic lep) if we have one
+  if (have($familiar`Robortender`) && (!lepOutfitMeatPercent || !lepOutfitFamWeight)) {
```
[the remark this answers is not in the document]
Pantocyclus: Not for a reasonable account imo

src/familiar.ts
```diff
@@ -281,3 +368,158 @@ export function timeToMeatify(): boolean {
...
 }
+
+type MarginalFamiliar = GeneralFamiliar & {
+  marginalValue: number;
+};
+
+let lepOutfitMeatPercent: number | null = null;
+let lepOutfitFamWeight: number | null = null;
+let nonLepOutfitMeatPercent: number | null = null;
+let nonLepOutfitFamWeight: number | null = null;
```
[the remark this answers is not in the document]
Pantocyclus: Stole some code :P

src/index.ts
Pantocyclus: Removed underwater, but retained location because knowing that we are in barf helps us evaluate fairies more accurately (this may be useful when we eventually use the same framework for free fights)

src/familiar.ts
```diff
@@ -270,7 +272,14 @@ const standardFamiliars: () => GeneralFamiliar[] = () =>
...
+      expectedValue:
+        garboValue($item`elemental sugarcube`) / 5 +
+        (Robortender.currentDrinks().includes($item`Feliz Navidad`)
+          ? get("garbo_felizValue", 0) * 0.25
+          : 0) +
+        (Robortender.currentDrinks().includes($item`Newark`)
+          ? get("garbo_newarkValue", 0) * 0.25
+          : 0),
     },
     ...$familiars`Hobo Monkey, Cat Burglar, Urchin Urchin, Leprechaun`.map((familiar) => ({
```
horrible-little-slime, line 284: I would steal the code I put in here--it makes it so we only consider your meatFamiliar(), rather than every lep-like

src/familiar.ts
```diff
@@ -439,84 +446,44 @@ export function setMarginalFamiliar(loc: Location, underwater: boolean): void {
...
+        100;
       return {
         ...fam,
         marginalValue: fam.expectedValue + additionalValue,
       };
     })
     .sort((left, right) => right.marginalValue - left.marginalValue);
 
-  const nominalOutfitValue = ((lepOutfitMeatPercent ?? 0) * baseMeat) / 100;
+  let nominalOutfitValue: number | null = null;
```
horrible-little-slime, line 465:
```
const outfitVal = (value: { meat: number, weight: number, bonus: number }) => value.meat * locBaseMeat / 100;
const nominalOutfitValue = cachedOutfits.reduce((a, b) => outfitVal(a) > outfitVal(b) ? a : b);
```
or something like that

src/familiar.ts
```diff
@@ -439,84 +446,44 @@ export function setMarginalFamiliar(loc: Location, underwater: boolean): void {
...
-          print(`nonLepOutfitMeatPercent: ${nonLepOutfitMeatPercent}%`, "blue");
-          print(`nonLepOutfitFamWeight: ${nonLepOutfitFamWeight}lbs`, "blue");
-        }
-        additionalValue = (toInt(barf) * nonLepOutfitMeatPercent * baseMeat) / 100;
-      }
+      const { meat } =
+        cachedOutfits.get(fam.leprechaunMultiplier) ??
+        cacheOutfit(fam.leprechaunMultiplier, fam.familiar);
+      const additionalValue =
+        (toInt(barf) * (meat + getFamModifier(fam, "Meat Drop")) * locBaseMeat +
```
horrible-little-slime, line 455: I would refrain from using toInt on the boolean and instead use ternary here

src/familiar.ts
```diff
@@ -439,84 +446,44 @@ export function setMarginalFamiliar(loc: Location, underwater: boolean): void {
...
-  const nominalOutfitValue = ((lepOutfitMeatPercent ?? 0) * baseMeat) / 100;
+  let nominalOutfitValue: number | null = null;
+  cachedOutfits.forEach((value) => {
+    if (!nominalOutfitValue || (value.meat * locBaseMeat) / 100 < nominalOutfitValue) {
+      nominalOutfitValue = (value.meat * locBaseMeat) / 100;
+    }
+  });
+
   print("Considering Marginal Familiars:", "blue");
   dropFamiliars.forEach((fam) => {
```
horrible-little-slime, line 473: It looks like this code currently does not account for the bonus equips in each outfit? I could be missing where that's happening

Pantocyclus: Ah, in my head I was thinking that we were always going to be equipping the same bonus equips (katana, mafia ring etc) that it wouldn't matter (since I normalize it with the `nominalOutfitValue`).

I guess this could matter in cases where we don't actually use the same bonus equips across different outfit.

horrible-little-slime:
Resolve conflicts & add something to the relay browser for the new property we're creating

horrible-little-slime:
This final request is unfortunately bigger and harder, but:

When ascending, we want to prioritize using marginal familiars during overdrunk turns when we use a wineglass and thus get lower expected utility out of our meat familiar (we also lose out on Sing Along during those turns, so it's a double whammy).

Regardless of whether we're ascending, when we have any sort of temporary +weight buff from the start of the day (e.g., witchess or familiar.enh or the suzie buff from zatara), we want to prioritize using our meat familiar for those turns and push marginal guys further back to the end of the day.

We can likely use something like ```
getActiveEffects().filter((ef) => haveEffect(ef) < myAdventures() && toSkill(ef) === $skill`none` && getModifier("Familiar Weight", ef) > 0)
```
to determine which expiring +weight buffs we have. Also, obviously not every expiring +weight buff should qualify.

We already have data for most drops familiars on the expected number of turns to get a drop--we can probably use that in conjunction with some math to work backwards to figure out when to stop using our robortender. Because it feels pretty bad to burn start-of-barf turns with a green pixie and then run a low-weight robortender when we're overdrunk.

File: src/familiar.ts [hunk 13]
```diff
@@ -281,3 +384,182 @@ export function timeToMeatify(): boolean {
...
+  const values = {
+    weight: sum(outfit, (eq: Item) => getModifier("Familiar Weight", eq)),
+    meat: sum(outfit, (eq: Item) => getModifier("Meat Drop", eq)),
+    bonus: sum(outfit, (eq: Item) => bonuses.get(eq) ?? 0),
+  };
+  cachedOutfits.set(mult, values);
+  return values;
+}
+
+export function setMarginalFamiliar(loc: Location): void {
```
gausie, line 418: This function desperately needs splitting up. I'd consider [reading about the SLA principle](https://medium.com/javarevisited/slap-that-ugly-code-6ec276d3a4bc) and thinking about how many layers of abstraction this function has currently.

gausie (commented):
I'd like to see familiar.ts broken up into

> .
└── familiar/
    ├── index.ts
    ├── rotating.ts
    ├── standard.ts
    └── drop.ts

or something. It's hard (for me!) to even start understanding this code based on the amount of different things being captured here.

src/familiar.ts
```diff
@@ -281,3 +384,182 @@ export function timeToMeatify(): boolean {
...
+    const idx =
+      globalOptions.ascending &&
+      nonJellyExpectedAdv < 30 && // Assume 30 overdrunk barf adventures
+      myInebriety() <= inebrietyLimit() &&
+      jellyfishEV > meatFamEV &&
+      dropFamiliars[0].familiar !== meatFamiliar()
+        ? dropFamiliars.findIndex((fam) => {
+            fam.familiar === (jellyfishEV > meatFamEV ? $familiar`Space Jellyfish` : meatFam);
+          }) ?? 0
+        : 0;
```
horrible-little-slime, line 548: .findIndex returns -1 for invalid inputs, not undefifined

Also I have no idea what this is trying to do, this logic should be sliced out into a function instead of being done with a massive ternary

Pantocyclus: dropFamiliars[0] gives us the best marginal familiar to run
However, if we are ascending, then we might not want to use the best marginal familiar (saving it instead for overdrunk turns). `idx` thus tries to skip the best marginal familiar if the following conditions are met
1) We must be ascending
2) The number of remaining marginal familiar turns (excluding jellyfish) we expect to run is < 30 (we will then save these ~30 marginal familiar turns to run when overdrunk)
3) We must not be overdrunk
4) ~~I forgot to remove the `jellyfishEV > meatFamEV` check here (it should only be checked in the next instance where you see it)~~

Now if the above conditions are satisfied, we want to skip all our remaining marginal familiars, and only consider either the jellyfish or robort. If jellyfishEV > meatFamEV, we get the index of the jellyfish, else we get the index of the robort. Our marginal familiar is then chosen with dropFamiliars[idx]

gausie: This would be great as comments interspersed into a well-named function

PR #967, opened 2022-06-22
Marginalia

horrible-little-slime:


src/familiar.ts
```diff
@@ -55,177 +63,326 @@ export function meatFamiliar(): Familiar {
...
+  const values = {
+    weight: sum(outfit, (eq: Item) => getModifier("Familiar Weight", eq)),
+    meat: sum(outfit, (eq: Item) => getModifier("Meat Drop", eq)),
+    bonus: sum(outfit, (eq: Item) => bonuses.get(eq) ?? 0),
+  };
+  cachedOutfits.set(mult, values);
+  return values;
+}
+
+export function setMarginalFamiliar(): void {
```
Pantocyclus, line 343: We may eventually want to change the free fight familiars to use a similar structure (valuing which familiar to prioritise makes sense there too). In such a case, it is absolutely crucial that we do not bring out our jellyfish in a non-stench zone, lest we inaccurately value its worth.
I agree with the removal of `underwater`, but I believe there's still value to be had in leaving the `location` in.

src/familiar.ts
```diff
@@ -55,177 +63,326 @@ export function meatFamiliar(): Familiar {
...
 
-  for (const familiar of $familiars`Hobo Monkey, Cat Burglar, Urchin Urchin, Leprechaun`) {
-    if (have(familiar)) familiarValue.push([familiar, 1]);
-  }
+function cacheOutfit(
+  mult: number,
+  familiar: Familiar
+): { weight: number; meat: number; bonus: number } {
+  useFamiliar(familiar);
+  meatOutfit(false, undefined, false);
```
Pantocyclus, line 331: I never did dive deep into the issue, but there were several instances where garbo decided the kramco was optimal for one outfit but not the other, resulting in a terribly wrong comparison between leps and non-leps.

In the ideal case, the `Requirements` here should enforce that the maximized result is exactly our normal barf outfit for that familiar. `undefined` doesn't seem to cut it, especially in edge cases relating to free fights.

src/familiar.ts
```diff
@@ -55,177 +63,326 @@ export function meatFamiliar(): Familiar {
...
+    {
+      familiar: $familiar`Space Jellyfish`,
+      expectedValue:
+        garboValue($item`stench jelly`) /
+        (get("_spaceJellyfishDrops") < 5 ? get("_spaceJellyfishDrops") + 1 : 20),
+      leprechaunMultiplier: 0,
+    },
+  ].map(({ familiar, expectedValue, leprechaunMultiplier }) => {
+    const { meat, weight, bonus } =
+      cachedOutfits.get(leprechaunMultiplier) ?? cacheOutfit(leprechaunMultiplier, familiar);
```
Pantocyclus, line 362: According to Dict, fairychauns are worth 7/6th of a normal lep (or rather, fairies are 1/6th of a lep). I didn't really have time today to work through stuff proper, but I don't know if you intend to group that into the `leprechaunMultiplier`, or if you eventually want to end up with a structure that handles the "hashing of a tuple"

PR #976, opened 2022-07-01
UPC support

horrible-little-slime:


src/outfit.ts
```diff
@@ -205,6 +214,23 @@ export function meatOutfit(embezzlerUp: boolean, requirement?: Requirement, sea?
       forceEquip.push($item`Kramco Sausage-o-Matic™`);
     }
   }
+
+  if (embezzlerUp) {
+    const currentWeapon = 25 * findLeprechaunMultiplier(meatFamiliar());
+    const value = ((75 - currentWeapon) * (750 + baseMeat)) / 100;
```
Pantocyclus, line 220:
```suggestion
    const currentWeapon = equippedItem($slot`weapon`);
    const currentWeaponValue =
      ((numericModifier(currentWeapon, "Meat Drop") +
        2.5 *
          numericModifier(currentWeapon, "Familiar Weight") *
          findLeprechaunMultiplier(meatFamiliar())) *
        (750 + baseMeat)) /
      100;
    const value = 75 * (750 + baseMeat)) / 100 - currentWeaponValue;
```

horrible-little-slime: We can by no means guarantee that the weapon currently occupying that slot is a weapon that is part of our meatoutfit.

src/outfit.ts
```diff
@@ -205,6 +214,23 @@ export function meatOutfit(embezzlerUp: boolean, requirement?: Requirement, sea?
...
+    const embezzlers = globalOptions.ascending
+      ? 30
+      : Math.min(30, embezzlerCount() || digitizedMonstersRemaining());
+    if (value * embezzlers > mallPrice($item`scratch 'n' sniff UPC sticker`)) {
+      acquire(3, $item`scratch 'n' sniff UPC sticker`, value * 30);
```
Pantocyclus, line 225: 30 seems like a magic number given that stickers fall off after 20 turns (I assume you're using a buffer of 1.5x?). I would think `value * embezzlers` is a good hard cap, since that's the exact value at which stickers become unprofitable after.

Also, do we need to always acquire 3? I guess there's no major harm since we won't purchase any more if we already have >3 in our inventory (and used none in previous `useUPCs()`), but we could also in theory acquire only as many stickers as we don't already have stuck on our sword.

horrible-little-slime:
Nope I just got the number wrong

src/outfit.ts
```diff
@@ -205,6 +214,23 @@ export function meatOutfit(embezzlerUp: boolean, requirement?: Requirement, sea?
...
     }
   }
+
+  if (embezzlerUp) {
+    const currentWeapon = 25 * findLeprechaunMultiplier(meatFamiliar());
+    const value = ((75 - currentWeapon) * (750 + baseMeat)) / 100;
+    const embezzlers = globalOptions.ascending
+      ? 30
+      : Math.min(30, embezzlerCount() || digitizedMonstersRemaining());
+    if (value * embezzlers > mallPrice($item`scratch 'n' sniff UPC sticker`)) {
```
Pantocyclus, line 224: `value` here is the value-per-adventure of a 3-sticker sword. Why are we comparing this to the price of a single sticker?

Also, if my suggested code above is used, then the RHS of this equation should be the number of stickers we need to add to our sword (i.e. the additional cost), and not just a flat 3.

src/yachtzee.ts
```diff
@@ -1105,6 +1115,21 @@ function prepareOutfitAndFamiliar() {
...
 
+function stickerSetup(expectedYachts: number) {
+  const currentStickers = $slots`sticker1, sticker2, sticker3`.map((s) => equippedItem(s));
+  const UPC = $item`scratch 'n' sniff UPC sticker`;
+  if (currentStickers.every((sticker) => sticker === UPC)) return;
+  const yachtOpportunityCost = 25 * findLeprechaunMultiplier(bestYachtzeeFamiliar());
+  const embezzlerOpportunityCost = 25 * findLeprechaunMultiplier(meatFamiliar());
+  const value =
+    ((75 - yachtOpportunityCost) * expectedYachts * 2000) / 100 +
+    ((75 - embezzlerOpportunityCost) * Math.min(20, expectedEmbezzlers) * (750 + baseMeat)) / 100;
```
Pantocyclus, line 1126: I don't like that `value` in `outfit.ts` represents valuePerAdventure, but `value` here represents the entirety of yachtzees and embezzlies we expect to run.

Also, the opportunity costs here can use the structure of the suggested code I commented above.

src/outfit.ts
Pantocyclus: I guess we could record our meatOutfit weapon as a global variable (or store it as a garbo mafia preference)?

horrible-little-slime: I feel like that's also a pretty bad solution for a variety of reasons (we don't have a single meatOutfit, or even a single _barf_ meatOutfit), but I'd welcome a more in-depth approximation than 25 * lepMult (which assumes you're either using a garbage sticker or a cursed pirate cutlass, basically)

src/outfit.ts
```diff
@@ -205,6 +213,28 @@ export function meatOutfit(embezzlerUp: boolean, requirement?: Requirement, sea?
...
   }
+
+  const stickerSlots = $slots`sticker1, sticker2, sticker3`;
+  const UPC = $item`scratch 'n' sniff UPC sticker`;
+
+  if (embezzlerUp) {
+    const currentWeapon = 25 * findLeprechaunMultiplier(meatFamiliar());
+    const embezzlers = globalOptions.ascending
+      ? 20
+      : Math.min(30, embezzlerCount() || digitizedMonstersRemaining());
```
Pantocyclus, line 224: Should this not be `!globalOptions.ascending` instead, since we only get to keep the stickers for future garbo runs if we aren't ascending.
Also, why are we using a value of 30 again?

src/outfit.ts
```diff
@@ -205,6 +213,28 @@ export function meatOutfit(embezzlerUp: boolean, requirement?: Requirement, sea?
...
+
+    const addedValueOfFullSword = (embezzlers * ((75 - currentWeapon) * (750 + baseMeat))) / 100;
+    if (addedValueOfFullSword > mallPrice(UPC)) {
```
Pantocyclus, line 227: If we need to add 3 stickers to get the full value of a full sword, shouldn't this be `3 * mallPrice(UPC)?`

src/outfit.ts
Pantocyclus: My idea was to constantly track and update our meatWeapon every barf turn (after gearing, and right before we run an adventure).

But, disregarding opportunity costs, as long as the sticker sword is profitable I don't think the inefficiencies here (which only applies to certain circumstances and not everyone) would be too big of a deal.

horrible-little-slime: I'm going to go ahead and declare this out of scope and we can try to improve it later on. But I don't particularly like any solution and I think this is Fine

PR #977, opened 2022-07-01
gausie said this is legal now

horrible-little-slime:


gausie:
Is the best vote always worth doing twice?

horrible-little-slime:
The best vote isn't always very good, but I'm having a hard time imagining
the second best is any better

gausie (approved):

PR #979, opened 2022-07-02
Only use pasta eyeball when profitable

Pantocyclus:
It is egregiously bad for character levels < 35 (for barf tourists; it may be marginally profitable for embezzlies and yachtzees at a lower level), even with MMJs. We use MMJs as a benchmark to see if it is indeed profitable for the account to run it.

### DIFF
--- a/src/mood.ts
+++ b/src/mood.ts
@@ -29,6 +29,7 @@ import {
 import { baseMeat, questStep, safeRestoreMpTarget, setChoice } from "./lib";
 import { withStash } from "./clan";
 import { usingPurse } from "./outfit";
+import { garboValue } from "./session";
 
 Mood.setDefaultOptions({
   songSlots: [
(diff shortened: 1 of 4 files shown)


Tokoeka (approved):

Pantocyclus:
There seems to be something broken with Libram's handing of Bind Lasagmbie, likely because the skill has different effects depending on your class, and `turnsPerCast` appears to return 0 (so our number of casts is effectively infinity, although it appears to be capped at 100 by the Libram implementation of `Mood`).

At this stage, we should either wait for a Libram fix, or manually handle Bind Lasagmbie outside of any Mood.

src/mood.ts
```diff
@@ -63,7 +64,17 @@ export function meatMood(urKels = false): Mood {
   mood.skill($skill`Drescher's Annoying Noise`);
   mood.skill($skill`Pride of the Puffin`);
 
-  if (myClass() !== $class`Pastamancer`) mood.skill($skill`Bind Lasagmbie`);
+  const mmjCost =
+    (100 -
+      (have($skill`Five Finger Discount`) ? 5 : 0) -
+      (have($item`Travoltan trousers`) ? 5 : 0)) *
+    (200 / (1.5 * myLevel() + 5));
+  const genericManaPotionCost = garboValue($item`generic mana potion`) * (200 / (2.5 * myLevel()));
```
horrible-little-slime, line 72: We want to use garboValue for things that we sell, but mallPrice for things that we buy

horrible-little-slime: Love the restructure

PR #980, opened 2022-07-04
Don't add item to menu if it's more expensive than our price cap

Pantocyclus:


Tokoeka (commented):
Looks good to me, but I dont fully understand the `filter` functionality so will leave approval to someone who does

horrible-little-slime:
I think we want to use mallPrice here on the basis that it's items we're _buying_, not _selling_.

Also, I'm not sure the extent to which this is optimal behavior! Gonna tag in Rever I think.

horrible-little-slime (approved):

PR #983, opened 2022-07-07
add getting High-End Ginger Wine as an option

Tokoeka:
Now we have 3 more Liver Space, grabbing High-End Ginger Wine if we have the outfit might be sensible, and more profitable than cigarettes?

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -13,6 +13,7 @@ import {
   getCampground,
   handlingChoice,
   haveEquipped,
+  haveOutfit,
   inebrietyLimit,
   isBanished,
   Item,
@@ -1506,7 +1507,7 @@ const freeRunFightSources = [
       (get("gingerbreadCityAvailable") || get("_gingerbreadCityToday")) &&
       get("_gingerbreadCityTurns") + (get("_gingerbreadClockAdvanced") ? 5 : 0) >= 10 &&
       get("_gingerbreadCityTurns") + (get("_gingerbreadClockAdvanced") ? 5 : 0) < 19 &&
-      availableAmount($item`sprinkles`) > 5,
+      (availableAmount($item`sprinkles`) > 5 || haveOutfit(`gingerbread best`)),
     (runSource: ActionSource) => {
       propertyManager.setChoices({
         1215: 1, // Gingerbread Civic Center advance clock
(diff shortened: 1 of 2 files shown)


src/fights.ts
```diff
@@ -1535,13 +1539,38 @@ const freeRunFightSources = [
...
+      const boozeVals = Array.from(boozes.values()).map((drink) => {
+        const buff = effectModifier(drink, "Effect");
+        const turnsPerUse = numericModifier(drink, "Effect Duration");
+        const meatDrop = numericModifier(buff, "Meat Drop");
+        const famWeight = numericModifier(buff, "Familiar Weight");
+        const buffValue = ((meatDrop + famWeight*25/10) * turnsPerUse * (baseMeat+750));
+        const advValue = (getAverageAdventures(drink) * get("valueOfAdventure"))
+        return {
+          booze: drink,
+          value: (garboValue(drink) + buffValue + advValue) / drink.inebriety,
```
pstalcup, line 1560: usually when we're consuming something, we prefer "mallPrice" to determine how much it costs to buy. garboValue is for computing how much we would sell it for

src/fights.ts
```diff
@@ -1535,13 +1539,38 @@ const freeRunFightSources = [
...
+        const famWeight = numericModifier(buff, "Familiar Weight");
+        const buffValue = ((meatDrop + famWeight*25/10) * turnsPerUse * (baseMeat+750)) / 100;
+        const advValue = (getAverageAdventures(drink) * get("valueOfAdventure"))
+        return {
+          booze: drink,
+          value: (garboValue(drink) + buffValue + advValue) / drink.inebriety,
+        }
+      });
+      const best = boozeVals.sort((a, b) => b.value - a.value)[0];
+      const gingerWineValue = ((.5*30*(baseMeat+750)) + (getAverageAdventures($item`High-end ginger wine`) * get("valueOfAdventure"))) / 2;
```
pstalcup, line 1564: if we pick up the wine, you should also add it to diet.ts

pstalcup (changes requested):

src/fights.ts
```diff
@@ -1535,13 +1538,44 @@ const freeRunFightSources = [
...
+      (availableAmount($item`sprinkles`) > 5 || haveOutfit(`gingerbread best`)),
     () => {
       propertyManager.setChoices({
         1203: 4, // Gingerbread Civic Center 5 gingerbread cigarettes
         1215: 1, // Gingerbread Civic Center advance clock
+        1209: 2, // enter the gallery at Upscale Midnight
+        1214: 1, // get High-End ginger wine
       });
-      adventureMacro($location`Gingerbread Civic Center`, Macro.abort());
+      const boozes = $items`elemental caipiroska, moreltini, Dreadsylvanian grimlet, Hodgman's blanket, Sacramento wine, iced plum wine, splendid martini, Eye and a Twist, jar of fermented pickle juice, dirt julep, Ambitious Turkey, Friendly Turkey`;
```
pstalcup, line 1549: i would prefer if this list was somehow generated from menu() and potionMenu() instead of being hard coded here. Possibly this whole next block of code could be a function in diet.ts

pstalcup (approved):

src/fights.ts
```diff
@@ -1506,7 +1509,7 @@ const freeRunFightSources = [
       (get("gingerbreadCityAvailable") || get("_gingerbreadCityToday")) &&
       get("_gingerbreadCityTurns") + (get("_gingerbreadClockAdvanced") ? 5 : 0) >= 10 &&
       get("_gingerbreadCityTurns") + (get("_gingerbreadClockAdvanced") ? 5 : 0) < 19 &&
-      availableAmount($item`sprinkles`) > 5,
+      (availableAmount($item`sprinkles`) > 5 || haveOutfit(`gingerbread best`)),
```
horrible-little-slime, line 1512: is it worth checking canEquip?

src/fights.ts
```diff
@@ -1535,13 +1538,30 @@ const freeRunFightSources = [
...
+        (0.5 * 30 * (baseMeat + 750) +
+          getAverageAdventures($item`high-end ginger wine`) * get("valueOfAdventure")) /
+        2;
+      const valueDif = gingerWineValue - best.value;
+      if (
+        availableAmount($item`sprinkles`) < 5 ||
+        (valueDif * 2 > garboValue($item`gingerbread cigarette`) * 5 &&
+          itemAmount($item`high-end ginger wine`) < 11)
+      ) {
+        outfit(`gingerbread best`);
```
horrible-little-slime, line 1560: try to avoid using backticks for things other than template literals

src/fights.ts
Tokoeka: Requires 20 in each stat max, can do if we still think it worth it

PR #985, opened 2022-07-08
don't use garboValue for the cost of things

horrible-little-slime:


horrible-little-slime:
Discussed with Panto (who is good) and this is not the plan for this

PR #990, opened 2022-07-09
Daily to get jick jar

Rinn:


horrible-little-slime (approved):

horrible-little-slime:
Is it worth adding a check to make sure the jickjar is the most valuable psychoanal jar?

Rinn:
We could, but my thought is it shows up so infrequently it's either always going to be or is rare enough we should just grab it to add to the supply because it's hardly going to affect profit margins.

PR #991, opened 2022-07-09
Daily to get pocket wishes

Rinn:


horrible-little-slime:
This is part of breakfast; is there a compelling reason to add it to garbo as well?

Rinn:
Oh is it, I didn't realize I've been doing it in my login script.

PR #992, opened 2022-07-09
Grab wine only if we have the gingerbread outfit

Pantocyclus:


Tokoeka (approved):
Ah yeah, i see the step I missed in the logic there

PR #994, opened 2022-07-10
Daily to unlock the sea and get sea jelly

Rinn:


Rinn:
I need to double check what happens if the property is wrong, don't merge this yet

horrible-little-slime:
sea jelly is in breakfast, although sea unlock isn't.

We do already unlock the sea at the start of operation, near where we unlock the skeleton store et al

PR #995, opened 2022-07-12
Quest wanderers

Tokoeka:
Add Some additional options for obtaining value from wanderers: Dinsey Quests, CI quests

### DIFF
--- a/src/wanderer.ts
+++ b/src/wanderer.ts
@@ -1,5 +1,5 @@
 import { canAdv } from "canadv.ash";
-import { buy, craftType, Item, Location, print, retrieveItem, use } from "kolmafia";
+import { buy, craftType, Item, Location, numericModifier, print, retrieveItem, runChoice, use, visitUrl } from "kolmafia";
 import {
   $effect,
   $item,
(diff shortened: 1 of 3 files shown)


gausie:
A lot of this would be welcome over at libram!

src/wanderer.ts
```diff
@@ -279,11 +282,77 @@ const wandererTargets = [
       return have(guzzlrBooze);
     }
   ),
+  new WandererTarget(
+    "CIOrder",
+    () => realmAvailable("spooky") && (get("questESpOutOfOrder") === "started" || get("questESpOutOfOrder") === "step1" || get("_questESp") === "questESPOutOfOrder"),
```
horrible-little-slime, line 287: questStep could really compress this down. And by "really" I mean "slightly"

horrible-little-slime:
My biggest ongoing concern for this PR is that we should almost certainly never sacrifice an equipment slot during embezzlers, which constitute a vast majority of our wanderers (the rest being start-of-day deterministic kramcos).

In an ideal world, with the `ascend` flag, we would prioritize quests that can be completed.

I know that SBB in particular is weird in that its quests don't reset on ascension. I have no further comment on this.

horrible-little-slime:
Closing this per the wanderer revamp, but now that these are in libram, it should be a more straightforward pr

Issue #745, opened 2022-01-13
[new feature] endOfDay action support for garbo (nightcaps, etc)


docrostov:
this is sourced from the #End-of-Day thread within the garbo chat on the discord. this is rooted in the initial idea that we want to move funfund acquisition to the end of the day. as part of that, we want to create a landing spot for "end-of-day" tasks. some initial thoughts on early things we'd like to add to this:

```
- finishGarboBarfDay()
--- map/saber for pills (implemented)
--- get funfunds (implemented)
--- time prank manny
--- overdrink+stooper via garbo diet script
--- get into your pajamas, check if you will have too many rollover advs
--- use adv resolutions if profitable & won't put you over the cap too much
--- run a few overdrunk turns if you go over the 200 cap
--- give the user a kiss on the forehead and tenderly put them to sleep, zzzz
--- output garbo profit tracking
```

a few design fundamentals for this:
- this should be opt in; garbo should not nightcap unless a user has expressed they want this
- we need to delineate between "end of barf day" tasks and "end of actual KOL day" tasks; i still think we encase them in one block but happy to hear counterarguments
- phred wants us to consider EOD turn-taking tasks; things like melf dupes, sparring, grabbing fused fuses, turning in gerald/ine, etc

horrible-little-slime:
End of Day tasks exist now! Congrats!

We don't actually nightcap specifically yet though, so I'm not personally closing this!

horrible-little-slime:
I actually am going to close this because so much of the body of the issue is devoted to the general functionality. But we do have EOD tasks, we just need to actually add nightcapping.

PR #998, opened 2022-07-13
Initial support for nightcap

pstalcup:


src/diet.ts
```diff
@@ -668,19 +668,31 @@ export function computeDiet(): {
...
   sweatpants: () => Diet<Note>;
+  stooper: () => Diet<Note>;
+  nightcap: () => Diet<Note>;
 } {
   // Handle spleen manually, as the diet planner doesn't support synth. Only fill food and booze.
 
   const orEmpty = (diet: Diet<Note>) =>
     diet.expectedValue(MPA, "net") < 0 ? new Diet<Note>() : diet;
+
+  const DRUNK_FACTOR = 0.75;
```
horrible-little-slime, line 679: nice

src/diet.ts
```diff
@@ -705,10 +717,23 @@ export function computeDiet(): {
...
+      nightcapDietPlanner(menu().filter((menuItem) => itemType(menuItem.item) === "booze")),
   };
 }
 
-type DietName = "FULL" | "SHOTGLASS" | "PANTSGIVING" | "REMAINING" | "SWEATPANTS";
+type DietName =
+  | "FULL"
+  | "SHOTGLASS"
+  | "PANTSGIVING"
+  | "REMAINING"
```
horrible-little-slime, line 733: As this list gets longer, "REMAINING" feels less and less informative a name. I can't think of a better one, though!

src/post.ts
```diff
@@ -149,12 +149,18 @@ function juneCleave(): void {
...
+  if (!skipDiet) {
     fillPantsgivingFullness();
     fillSweatyLiver();
   }
   coldMedicineCabinet();
   safeInterrupt();
   safeRestore();
   updateMallPrices();
+  if (myAdventures() === 0 && globalOptions.nightcap && globalOptions.overdrunk && !skipDiet) {
+    runNightcap();
```
horrible-little-slime, line 164: I would prefer a structure where nightcap gets run only in one spot, rather than being in two separate places for two separate scenarios.

I am by no means married to this. I'm very much a bachelor in this situation.

horrible-little-slime (commented):
Lookin promising, my dude

src/diet.ts
```diff
@@ -982,3 +1007,22 @@ export function runDiet(): void {
...
 }
+
+export function runNightcap(): void {
+  if (getRemainingLiver() === 0) {
+    withVIPClan(() => {
+      const dietBuilder = computeDiet();
+
+      if (have($familiar`Stooper`)) {
+        if (myFamiliar() !== $familiar`Stooper`) {
+          useFamiliar($familiar`Stooper`);
```
Tokoeka, line 1018: Does this take into account case of someone who (for some reason) has but cannot use Stooper? (in post-ronin g-lover for instance) [admittedly no idea why someone would be running garbo in that case, but want to check if its something we want to care about]

pstalcup: yes - this will attempt to use the stooper, and if it can't be used it will just continue to nightcap

src/diet.ts
Tokoeka: Do we want to allow the user to self-define Drunk_Factor (whether via relay, setting a preference or giving an argument when running?

src/post.ts
pstalcup: I will consider this. It comes down into the cleanest way to organize the code, and this (at hte time) seemed easier than reworking bigger parts of the adventuring code

src/diet.ts
pstalcup: It is a way of showing what we have left to consume for our diet as we loop through it.

src/post.ts
horrible-little-slime: Supes reasonable. I'm going to get one more pair of eyes on this maybe and then call it a day

horrible-little-slime:
#745

horrible-little-slime:
Alright I want to rebase and test this finally (sorry rever) but there's a merge conflict, and it's a meaningful one; it's one that deals with the stillsuit vs nightcapping.

We should chug sweat as a stooper booze. I would prefer not to write this, but I can if I must.

src/diet.ts
```diff
@@ -984,3 +1010,22 @@ export function runDiet(): void {
     }
   });
 }
+
+export function runNightcap(): void {
+  if (getRemainingLiver() === 0) {
```
phulin, line 1015:
```suggestion
  if (getRemainingLiver() === (myFamiliar() === $familiar`Stooper` ? 1 : 0)) {
```

src/index.ts
```diff
@@ -383,6 +383,10 @@ export function main(argString = ""): void {
       globalOptions.noDiet = true;
     } else if (arg.match(/yachtzeechain/)) {
       globalOptions.yachtzeeChain = true;
+    } else if (arg.match(/nightcap/)) {
+      globalOptions.nightcap = true;
+    } else if (arg.match(/overdrunk/)) {
+      globalOptions.overdrunk = true;
```
phulin, line 389: what does overdrunk mean? pls add to help message

src/diet.ts
```diff
@@ -984,3 +1010,22 @@ export function runDiet(): void {
...
+    withVIPClan(() => {
+      const dietBuilder = computeDiet();
+
+      if (have($familiar`Stooper`)) {
+        if (myFamiliar() !== $familiar`Stooper`) {
```
phulin, line 1020: this if is unnecessary

src/diet.ts
phulin: .75 seems a little low but fine as an initial estimate, I probably lose something like 1000/6000 MPA when overdrunk

horrible-little-slime:
I think this branch will basically need to be abandoned, but it can be addressed again once the grimoirization of barf-turn is complete.

horrible-little-slime:
Also don't delete the branch, it's still useful

PR #1004, opened 2022-07-18
Marginal Familiars, once again

horrible-little-slime:
Still needs:
 - Some horrible jellyfish math in marginalFamiliars.ts
 - Actual implementation in the main body of code


horrible-little-slime:
I had to do some surgery on the grape about this and I'm not crazy about it but c'est la vie:

after making the `familiarAbilityValue` and `totalFamiliarValue` functions, `barfFamiliar` started to get a little crowded, so I wanted to yoink those functions out of `barfFamiliar`. But the reason they were all there to begin with is because `familiarModifier` is, and `familiarModifier` relies on knowing your `passiveWeight`, which is your weight from effects + passive skills (but NOT from your outfit).

passiveWeight isn't crazy expensive to compute, but also I would prefer not to iterate over all of your effects several times per turn. So I moved it outside of the `barfFamiliar` function, made it a `let` initialized at 0, and made it so we set the passiveWeight before calling any of the familiar value functions.

This feels counterintuitive and hard to read, and I'm looking for a solution. The solution may very well be "this is actually fine you're just in your head about it because you're so used to the code looking like something else."

Pantocyclus (approved):
Shrub and experience familiars may need revisiting in the future, but they are very much out of the current scope of this PR.
From my tests, I believe it should be very much ready to be merged at this stage.

Malibu-Stacey:
> Shrub and experience familiars may need revisiting in the future, but they are very much out of the current scope of this PR. From my tests, I believe it should be very much ready to be merged at this stage.

Agreed. I've tested this the last 4 or 5 days in a row including one with a special build which excluded the Space Jellyfish completely to test the logic for people who don't own that but do own any of the other familiars & it works well in my opinion.

PR #1006, opened 2022-07-20
Support using absentee voter ballot

Pantocyclus:


src/dailies.ts
```diff
@@ -143,7 +146,47 @@ export function postFreeFightDailySetup(): void {
 }
 
 function voterSetup(): void {
-  if (have($item`"I Voted!" sticker`) || !(get("voteAlways") || get("_voteToday"))) return;
+  if (have($item`"I Voted!" sticker`)) return;
+
+  if (!get("voteAlways") && get("_voteToday")) {
```
gausie, line 151: It should be ! to both of these no?

src/dailies.ts
```diff
@@ -143,7 +146,47 @@ export function postFreeFightDailySetup(): void {
...
+        "Item Drop: +15",
+        0.15 *
+          (4 * 100 * 0.3 * embezzlerCount() +
+            3 * 200 * 0.15 * (estimatedTurns() - embezzlerCount())),
+      ],
+      ["Adventures: +1", globalOptions.ascending ? 0 : get("valueOfAdventure")],
+    ]);
+    const availableInitiatives: Map<string, number> = new Map(
+      Object.keys(votingBoothInitiatives(myClass(), myPathId(), myDaycount())).map((s) => {
+        return [s, 0];
```
gausie, line 168: Why iterate twice and not just set the value here?

src/dailies.ts
Pantocyclus: Oh yeah good catch!

src/dailies.ts
Pantocyclus: Not sure if the change was what you envisioned

gausie (dismissed):

src/dailies.ts
gausie: It was!

src/dailies.ts
```diff
@@ -143,7 +146,48 @@ export function postFreeFightDailySetup(): void {
...
+      ],
+      ["Adventures: +1", globalOptions.ascending ? 0 : get("valueOfAdventure")],
+    ]);
+
+    const availableInitiatives: Map<string, number> = new Map(
+      Object.keys(votingBoothInitiatives(myClass(), myPathId(), myDaycount())).map((init) => {
+        let val = 0;
+        valuableInitiatives.forEach((value, key) => {
+          if (init.includes(key)) val = value;
+        });
```
horrible-little-slime, line 172:
```suggestion
        const val = valuableInitiatives.get(init) ?? 0;
```

src/dailies.ts
```diff
@@ -143,7 +146,51 @@ export function postFreeFightDailySetup(): void {
...
+    [
+      "Item Drop: +15",
+      0.15 *
+        (4 * 100 * 0.3 * embezzlerCount() + 3 * 200 * 0.15 * (estimatedTurns() - embezzlerCount())),
+    ],
+    ["Adventures: +1", globalOptions.ascending ? 0 : get("valueOfAdventure")],
+    ["Familiar Experience: +2", 0],
+    ["Monster Level: +10", -1],
+    [`${myPrimestat()} Percent: +25`, -2],
+    [`Experience (${myPrimestat()}): +4`, -3],
```
horrible-little-slime, line 166: Why are these negative now?

Pantocyclus: This was Gausie's request to merge the ranking system of the previous array with the valuation system of the new array. For things that we don't have a good valuation for, we use negative values to prevent the logic from ever deciding to buy a ballot for the wrong reason (while allowing the follow on logic further down to still correctly rank the initiatives)

horrible-little-slime: cool
```
  const voteLocalPriorityArr = [
    [0, initPriority.get(get("_voteLocal1")) || (get("_voteLocal1").indexOf("-") === -1 ? 1 : -1)],
    [1, initPriority.get(get("_voteLocal2")) || (get("_voteLocal2").indexOf("-") === -1 ? 1 : -1)],
    [2, initPriority.get(get("_voteLocal3")) || (get("_voteLocal3").indexOf("-") === -1 ? 1 : -1)],
    [3, initPriority.get(get("_voteLocal4")) || (get("_voteLocal4").indexOf("-") === -1 ? 1 : -1)],
  ];
```
The issue is that this code is going to consider neutral or even negative inits as better than that

Pantocyclus: hmm I guess I could leave them at an ever so slightly positive value instead. Not that you can purchase the ballots from the mall at that price, but even if you could what's losing 10 meat in the grand scheme of things

horrible-little-slime:
I think that's well worth it. Did you account for the three free fights in
the price?

horrible-little-slime (approved):

PR #1008, opened 2022-07-23
Wanderer lovebug currency value

Rinn:
wal-mart gift certificates from VYKEA
Beach bucks from Fun Guy mansion
coinspiracy from The Deep Dark Jungle
Not doing fun funds since we already spend time in barf

Unknown drop rate, using 5% for now based on some casual log searching

### DIFF
--- a/src/wanderer.ts
+++ b/src/wanderer.ts
@@ -280,26 +280,26 @@ const wandererTargets = [
     }
   ),
   // Elemental Airport Currency drops.
-  // TODO: Unknown drop rate, using 10% from a quick log search
+  // TODO: Unknown drop rate, using 5% from a quick log search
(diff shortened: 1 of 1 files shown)


Rinn:
Got some more logs and looks like the wiki is wrong about the Coinspiracy cap at 5 drops

horrible-little-slime:
If there is truly no cap, then I think that doing funfunds actually _is_ helpful, if only because it prevents us from defaulting to the kitchen as much.

This looks good; I don't really care if we're super accurate about the droprates as long as we are eventually. 


This is well out of scope of this PR, but in the long-term it might be worth having the ability to "sum" different wanderer sources--if you're adventuring somewhere for a charter quest AND you get lovebugs, it might edge out some other zone.

horrible-little-slime (approved):

PR #1011, opened 2022-07-26
conditionally add jellyfish as a standard drop familiar

horrible-little-slime:


horrible-little-slime:
I'm going to contemplate a more complex approach to this; will also try to account for the fact that jellyfish can only be used sober, and the fact that when running overdrunk we may not spend all of those turns at barf

PR #1016, opened 2022-07-30
Revert "Support using absentee voter ballot"

horrible-little-slime:
Reverts Loathing-Associates-Scripting-Society/garbage-collector#1006

### DIFF
--- a/src/dailies.ts
+++ b/src/dailies.ts
@@ -21,11 +21,9 @@ import {
   maximize,
   meatPockets,
   myClass,
-  myDaycount,
   myHp,
   myInebriety,
   myMaxhp,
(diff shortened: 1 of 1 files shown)


eegeeZA:
What happened that this was reverted?

horrible-little-slime:
It repeatedly caused crashes on everyone for whom it was applicable, which is a bit of a dealbreaker.

If I had to guess, using `use` made mafia anticipate that it was going to encounter a choice adventure, and then it tried and failed to handle the choice adventure itself. If we switch to `visitUrl` it will (probably) work.

eegeeZA:
Okay cool, I missed the update. See it now.

PR #1017, opened 2022-07-30
use stattype enum in pantogram code

horrible-little-slime:


libraryaddict (approved):
Looks good but in the future please try to maintain higher code quality.

PR #1019, opened 2022-07-30
remove some temporary boost gear from marginal familiar calcs

horrible-little-slime:


Pantocyclus:
I've been running #1020 for the past week or so and it seems to work well. They are functionally the same, but in the future I believe the optional argument can be used as arguments within the individual dropgear functions (to compute some constant/continual bonuses while ignoring the circumstantial/intermittent ones), as opposed to simply removing the consideration of the entire item.

PR #1037, opened 2022-08-13
Support for variable-power potions

Pantocyclus:


src/potions.ts
```diff
@@ -337,6 +342,18 @@ export const farmingPotions = [
...
   ),
+  new Potion($item`papier-mâché toothpicks`, {
+    effect: $effect`Eyes Wide Propped`,
+    canDouble: true,
+    duration: 40,
+    use: (quantity: number) =>
+      new Array(quantity).fill(0).every(() => {
+        acquire(4, $item`Rad Lib`);
+        acquire(4, $item`soda water`);
+        create(1, $item`papier-mâché toothpicks`);
```
horrible-little-slime, line 353: just use retrieveItem to handle item creation imo

src/potions.ts
```diff
@@ -337,6 +342,18 @@ export const farmingPotions = [
           new Array(quantity).fill(0).every(() => cliExecute(`genie effect ${effect}`)),
       })
   ),
+  new Potion($item`papier-mâché toothpicks`, {
+    effect: $effect`Eyes Wide Propped`,
+    canDouble: true,
+    duration: 40,
```
horrible-little-slime, line 348: This is unnecessary given that it is the default behavior

src/potions.ts
```diff
@@ -434,3 +430,121 @@ export function bathroomFinance(embezzlers: number): void {
...
+    // If potion is not tradeable, use retrievePrice instead
+    return this.potion.tradeable
+      ? historical && historicalAge(this.potion) < 14
+        ? historicalPrice(this.potion)
+        : mallPrice(this.potion)
+      : retrievePrice(this.potion);
+  }
+}
+
+function triangleNumber(n2: number, n1?: number) {
```
horrible-little-slime, line 472: assign this a default value of 0 instead of using nullish coalescence

src/potions.ts
```diff
@@ -434,3 +430,121 @@ export function bathroomFinance(embezzlers: number): void {
...
+  return 0.5 * (n2 * (n2 + 1) - (n1 ?? 0) * ((n1 ?? 0) + 1));
+}
+
+export function considerVariableMeatPotions(yachtzees: number, embezzlers: number): void {
```
horrible-little-slime, line 476: feels like much of this logic should be a method on the class of variablemeatpotion, otherwise, what's the point of the class?

src/potions.ts
```diff
@@ -434,3 +430,121 @@ export function bathroomFinance(embezzlers: number): void {
     }
   }
 }
+
+class VariableMeatPotions {
```
horrible-little-slime, line 434: vibe-wise I expect this to extend Potion--I am incredibly unmarried to this. I cannot imagine being less married to something. Also, should be called `VariablePotion` singular, probably

src/potions.ts
horrible-little-slime: also I think using `a` and `b` or something is less confusing than n1 and n2 but, again, very much a bachelor about it

src/potions.ts
```diff
@@ -434,3 +430,121 @@ export function bathroomFinance(embezzlers: number): void {
...
+}
+
+export function considerVariableMeatPotions(yachtzees: number, embezzlers: number): void {
+  const potions = [
+    new VariableMeatPotions($item`love song of sugary cuteness`, 20, 2),
+    new VariableMeatPotions($item`pulled yellow taffy`, 50, 2),
+    // new VariableMeatPotions($item`porcelain candy dish`, 500, 1),
+  ];
+
+  potions.forEach((potion) => {
```
horrible-little-slime: unless you have a compelling reason, we generally prefer `for` loops to `forEach`

src/potions.ts
```diff
@@ -434,3 +430,121 @@ export function bathroomFinance(embezzlers: number): void {
...
+              100;
+            decayDuration -= turns;
+          } else {
+            totalValue +=
+              (value * decayDuration * potion.meatBonusPerTurn * triangleNumber(decayDuration)) /
+              100;
+            decayDuration = 0;
+          }
+        }
+      });
```
horrible-little-slime, line 524: Making this a method on the class should let you avoid having this function-in-a-function constrction

src/potions.ts
```diff
@@ -434,3 +430,121 @@ export function bathroomFinance(embezzlers: number): void {
...
+      const totalCosts = n * potion.price(false);
+      const totalDuration = n * potion.duration;
+      let cappedDuration = Math.max(0, totalDuration - potion.softcap + 1);
+      let decayDuration = Math.min(totalDuration, potion.softcap - 1);
+      let totalValue = 0;
+      [
+        [yachtzees, yachtzeeValue],
+        [embezzlers, embezzlerValue],
+        [barfTurns, barfValue],
+      ].forEach((arr) => {
```
horrible-little-slime, line 499: why won't this let me write a suggestion I have so many suggestions

```
const turnTypes = [[yachtzees, yachtzeeValue], etc];

for (const [quantity, value] of turnTypes) {
```
(the TL;DR is name that variable, use a for loop, and destructure the variable instead of taking it as an array

src/potions.ts
```diff
@@ -434,3 +430,121 @@ export function bathroomFinance(embezzlers: number): void {
...
+      Math.floor(yachtzees / potion.duration),
+      Math.ceil(yachtzees / potion.duration),
+      Math.floor((yachtzees + embezzlers + potion.softcap) / potion.duration),
+      Math.ceil((yachtzees + embezzlers + potion.softcap) / potion.duration),
+      Math.floor((yachtzees + embezzlers) / potion.duration),
+      Math.ceil((yachtzees + embezzlers) / potion.duration),
+      Math.floor((yachtzees + embezzlers + barfTurns + potion.softcap) / potion.duration),
+      Math.ceil((yachtzees + embezzlers + barfTurns + potion.softcap) / potion.duration),
+      Math.floor((yachtzees + embezzlers + barfTurns) / potion.duration),
+      Math.ceil((yachtzees + embezzlers + barfTurns) / potion.duration),
```
horrible-little-slime, line 540: This chunk has a high degree of parallel structure; it will likely be easier to read if you build this through iteration as opposed to hardcoding all fifty-seven cases

src/potions.ts
horrible-little-slime: (which I would personally prefer we avoid when possible)

src/potions.ts
```diff
@@ -434,3 +430,121 @@ export function bathroomFinance(embezzlers: number): void {
...
+    ];
+    const profits = potionAmountsToConsider.map(valueNPotions);
+    const bestProfits = profits.reduce((a, b) => (a < b ? a : b));
+    const bestIdx = profits.indexOf(bestProfits);
```
horrible-little-slime, line 544: I don't think this is necessary; the way bestProfits is constructed, we _can't_ get fellas from outside the array into it

src/potions.ts
```diff
@@ -434,3 +430,121 @@ export function bathroomFinance(embezzlers: number): void {
...
+    if (bestProfits > 0 && bestIdx >= 0) {
+      return potionAmountsToConsider[bestIdx];
```
horrible-little-slime: ahh I see what's happening here.

A nicer construction might be to map `potionAmountsToConsider` into a typed object, with one property that tells us the `potionAmounts`, and one that tells us the `value`. You still reduce the mapped array, but you use `a.value < b.value`, and then you return `bestProfits.potionAmounts`. I personally find this more elegant than tracking the index. There's a word for it that I'm having trouble remembering.

Also maybe rename those variables, but just because I changed their meaning.

horrible-little-slime (changes requested):

src/potions.ts
horrible-little-slime: also, given that at the end we _use_ the potions, `consider` is maybe not the ideal word to use; consider `setup` or `apply` or `use`

src/potions.ts
```diff
@@ -417,3 +430,134 @@ export function bathroomFinance(embezzlers: number): void {
...
+      ? historical && historicalAge(this.potion) < 14
+        ? historicalPrice(this.potion)
+        : mallPrice(this.potion)
+      : retrievePrice(this.potion);
+  }
+
+  getOptimalNumberToUse(yachtzees: number, embezzlers: number): number {
+    const barfTurns = Math.max(0, estimatedTurns() - yachtzees - embezzlers);
+
+    const potionAmountsToConsider = [] as number[];
```
horrible-little-slime, line 478:
```suggestion
    const potionAmountsToConsider: number[] = [];
```

src/potions.ts
```diff
@@ -417,3 +430,134 @@ export function bathroomFinance(embezzlers: number): void {
...
+    for (const fn of [Math.floor, Math.ceil]) {
+      for (const sc of [0, this.softcap]) {
+        for (const em of [0, embezzlers]) {
+          for (const bt of [0, barfTurns]) {
```
horrible-little-slime, line 482: Name these a little more constructively

src/potions.ts
```diff
@@ -417,3 +430,134 @@ export function bathroomFinance(embezzlers: number): void {
...
+            potionAmountsToConsider.push(fn((yachtzees + em + bt + sc) / this.duration));
+          }
+        }
+      }
+    }
+
+    const profits = [] as number[];
```
horrible-little-slime, line 489:
```suggestion
    const profits: number[] = [];
```

src/potions.ts
```diff
@@ -417,3 +430,134 @@ export function bathroomFinance(embezzlers: number): void {
...
+    const profitsMap = new Map(
+      potionAmountsToConsider.map((n) => {
+        const profit = this.valueNPotions(n, yachtzees, embezzlers, barfTurns);
+        profits.push(profit);
```
horrible-little-slime, line 493: why not define profits as `Array.from(profitsMap.keys())`?

I would also probably not use a Map here, because we don't technically know that profits are unique."

So instead, maybe something like

```
const profitsFromPotions = potionAmountsToConsider.map((quantity) => ({ quantity, value: this.valueNPotions(quantity, yachtzees, embezzlers, barfTurns)"))
```

then later
```
const bestOption = profits.reduce((a, b) => a.value > b.value ? a : b);
if (bestOption.value > 0) {
  const potionsToUse = bestOption.quantity - Math.floor(haveEffect(this.effect) / this.duration);
  return Math.max(potionsToUse, 0);
}
```

src/potions.ts
```diff
@@ -417,3 +430,134 @@ export function bathroomFinance(embezzlers: number): void {
...
+    let decayDuration = Math.min(totalDuration, this.softcap - 1);
+    let totalValue = 0;
+    const turnTypes = [
+      [yachtzees, yachtzeeValue],
+      [embezzlers, embezzlerValue],
+      [barfTurns, barfValue],
+    ];
+
+    // eslint-disable-next-line prefer-const
+    for (let [turns, value] of turnTypes) {
```
horrible-little-slime, line 525: We only need to use `let` in for loops of the form `for (let i = start; i <= end; i++)` or similar things. `const` works just fine for for...of loops. Or for...in loops, but those are dumb.

src/potions.ts
```diff
@@ -417,3 +430,134 @@ export function bathroomFinance(embezzlers: number): void {
...
+      if (cappedDuration >= turns) {
+        totalValue += (value * turns * this.cappedMeatBonus) / 100;
+        cappedDuration -= turns;
```
horrible-little-slime, line 528: It might (this is really a "might", not a gently phrased "do this") read better if instead of doing `if...else`, you did
```
if (condition) {
  doTheThing();
  continue;
}
more code here
```
continue works a little like an early return in a function; it'll just move on to the next phase of the loop

horrible-little-slime: It could also be nicer to slice the stuff inside this function out into a loop, and then you can literally use `return`

horrible-little-slime (changes requested):
This is already so much nicer :)

src/potions.ts
```diff
@@ -162,7 +162,8 @@ export class Potion {
       ? historical && historicalAge(this.potion) < 14
         ? historicalPrice(this.potion)
         : mallPrice(this.potion)
-      : retrievePrice(this.potion);
+      : retrievePrice(this.potion, itemAmount(this.potion) + 1) -
+          autosellPrice(this.potion) * itemAmount(this.potion);
```
horrible-little-slime, line 166: I would somehow slice out this logic somewhere and name it so that it makes a little more sense

Also, I think we do actively want to use availableAmount, unless we manually set all of the various autoSatisfy settings. Which, maybe we do!

src/potions.ts
Pantocyclus: I need to decrement turns (although value is left untouched), so I do need a `let` somewhere

src/potions.ts
Pantocyclus: I don't know if I agree with the `continue` comment. I'd use `continue` and early returns when the condition I'm considering is really specific/niche, but this feels more like a red pill/blue pill thing. It feels more like either/or, which I find it easier to read as an if/else.

I don't like this current structure either, but I do need to capture `cappedDuration`, `decayDuration` and `totalValue` in this scope (technically I don't really need to capture `totalValue`). I'm not sure if I'm following what you have in mind when you mention slicing the stuff out.

gausie: Really these should probably be two functions that are selected by this if statement. You'll need to work around decrementing the various values in doing so, but it'll be better code.

Pantocyclus: Aight, no more if/else thanks to the power of maths (but mostly thanks to more braincells being available today). Hopefully this is much better code

horrible-little-slime:
Is this tested? It's looking good to me right now

src/potions.ts
```diff
@@ -417,3 +431,132 @@ export function bathroomFinance(embezzlers: number): void {
...
+
+    return totalValue - totalCosts;
+  }
+}
+
+export function variableMeatPotionsSetup(yachtzees: number, embezzlers: number): void {
+  const potions = [
+    new VariableMeatPotion($item`love song of sugary cuteness`, 20, 2),
+    new VariableMeatPotion($item`pulled yellow taffy`, 50, 2),
+    // new VariableMeatPotions($item`porcelain candy dish`, 500, 1),
```
horrible-little-slime, line 555: Is there a reason this is commented out?

src/acquire.ts
```diff
@@ -36,10 +36,13 @@ export function acquire(
   item: Item,
   maxPrice?: number,
   throwOnFail = true,
-  maxAggregateCost?: number
+  maxAggregateCost?: number,
+  tryRetrievingUntradeable = false
```
horrible-little-slime, line 40: This might be getting to be a lot of additional optional parameters; it might be better to start adding an Options object instead

horrible-little-slime (commented):
Looks pretty much ready 2 roll; we can discuss these comments then gavel it out and push

src/potions.ts
Pantocyclus: I wasn't super comfortable playing with potentially millions of meat's worth of a buff, especially when we need at least 500 to get the ball rolling (and I have reservations about estimating prices of bulk purchases)

horrible-little-slime (dismissed):

Pantocyclus:
#1049 is waiting on the merger of this PR to improve our yachtzeechaining profits (as well as extro profits)

PR #1039, opened 2022-08-13
be a little smarter about our selection of fairies

horrible-little-slime:
@SSBBHax I hope I successfully pinged you.

This does:
 - Add the gnome as a possible pull for drum machines
 - Assign the gnome a bonus value under the assumption it'll hit about 70 lbs
 - Sort our fairies by drop-value rather than by leprechaun multiplier
 This does not:
 - Ever use the gnome when we have a remotely better fairy, even if we're wasting it
 - Add the gnome as what the kids call a "constant value familiar"
 - Actively inform you that the oldest human-planted tree with a known planting date resides in the island nation of Sri Lanka, whose chief exports include tea and rubber

### DIFF
--- a/src/familiar/itemFamiliar.ts
+++ b/src/familiar/itemFamiliar.ts
@@ -1,5 +1,5 @@
-import { Familiar } from "kolmafia";
-import { $familiar, $item, findFairyMultiplier, get, have } from "libram";
+import { Familiar, myFamiliar, runChoice, useFamiliar, visitUrl } from "kolmafia";
+import { $familiar, $item, findFairyMultiplier, get, have, set } from "libram";
 import { menu } from "./freeFightFamiliar";
 
 let bestNonCheerleaderFairy: Familiar;
@@ -25,6 +25,23 @@ export function bestFairy(): Familiar {
 
     const highestFairyMult = findFairyMultiplier(viableFairies[0]);
     const goodFairies = viableFairies.filter((f) => findFairyMultiplier(f) === highestFairyMult);
+
+    if (
+      have($familiar`Reagnimated Gnome`) &&
+      !have($item`gnomish housemaid's kgnee`) &&
+      !get("_garbo_triedForKgnee", false)
+    ) {
+      const current = myFamiliar();
+      useFamiliar($familiar`Reagnimated Gnome`);
+      visitUrl("arena.php");
+      runChoice(4);
+      useFamiliar(current);
+      set("_garbo_triedForKgnee", true);
+    }
+
+    if (have($item`gnomish housemaid's kgnee`) && highestFairyMult === 1) {
+      goodFairies.push($familiar`Reagnimated Gnome`);
+    }
 
     const bonuses = [
       ...menu(true, false),
--- a/src/familiar/itemFamiliar.ts
+++ b/src/familiar/itemFamiliar.ts
@@ -55,8 +55,8 @@ export function bestFairy(): Familiar {
 
     goodFairies.sort(
       (a, b) =>
-        bonuses.find(({ familiar }) => familiar === b)?.expectedValue ??
-        0 - (bonuses.find(({ familiar }) => familiar === a)?.expectedValue ?? 0)
+        (bonuses.find(({ familiar }) => familiar === b)?.expectedValue ?? 0) -
+        (bonuses.find(({ familiar }) => familiar === a)?.expectedValue ?? 0)
     );
 
     bestNonCheerleaderFairy = goodFairies[0];
--- a/src/familiar/itemFamiliar.ts
+++ b/src/familiar/itemFamiliar.ts
@@ -1,0 +1,49 @@
+import { Familiar } from "kolmafia";
+import { $familiar, $item, findFairyMultiplier, get, have } from "libram";
+import { menu } from "./freeFightFamiliar";
+
+let bestNonCheerleaderFairy: Familiar;
+
+export function bestFairy(): Familiar {
+  if (have($familiar`Trick-or-Treating Tot`) && have($item`li'l ninja costume`)) {
+    return $familiar`Trick-or-Treating Tot`;
+  }
+
+  if (
+    have($familiar`Steam-Powered Cheerleader`) &&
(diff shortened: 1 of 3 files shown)


src/familiar/itemFamiliar.ts
```diff
@@ -0,0 +1,66 @@
+import { Familiar, myFamiliar, runChoice, useFamiliar, visitUrl } from "kolmafia";
+import { $familiar, $item, findFairyMultiplier, get, have, set } from "libram";
+import { menu } from "./freeFightFamiliar";
+
+let bestNonCheerleaderFairy: Familiar;
+
+export function bestFairy(): Familiar {
+  if (have($familiar`Trick-or-Treating Tot`) && have($item`li'l ninja costume`)) {
+    return $familiar`Trick-or-Treating Tot`;
+  }
```
Pantocyclus, line 10: If we're expecting our item familiars to be at least 70lbs, a Jumpsuited Hounddog matches the ninja tot at 67.8lbs (while regular fairies match it at 84.7lbs). I'm not sure if we want an early return here if we're running with that assumption.

src/familiar/itemFamiliar.ts
```diff
@@ -0,0 +1,66 @@
...
+        leprechaunMultiplier: 0,
+        limit: "none",
+      },
+    ];
+
+    goodFairies.sort(
+      (a, b) =>
+        bonuses.find(({ familiar }) => familiar === b)?.expectedValue ??
+        0 - (bonuses.find(({ familiar }) => familiar === a)?.expectedValue ?? 0)
+    );
```
Pantocyclus, line 60: Pretty sure this is missing a bracket
```
goodFairies.sort(
    (a, b) =>
      (bonuses.find(({ familiar }) => familiar === b)?.expectedValue ?? 0) -
      (bonuses.find(({ familiar }) => familiar === a)?.expectedValue ?? 0)
  );
```

The above code should compile as
```
goodFairies.sort(
  (a, b) => {
    const temp = 0 - (bonuses.find(({ familiar }) => familiar === a)?.expectedValue ?? 0);
    return bonuses.find(({ familiar }) => familiar === b)?.expectedValue ?? temp;
});
```
which is most probably not what you want

src/familiar/itemFamiliar.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  if (!bestNonCheerleaderFairy) {
+    const viableFairies = Familiar.all()
+      .filter(
+        (f) => have(f) && findFairyMultiplier(f) && f !== $familiar`Steam-Powered Cheerleader`
+      )
+      .sort((a, b) => findFairyMultiplier(b) - findFairyMultiplier(a));
+
+    const highestFairyMult = findFairyMultiplier(viableFairies[0]);
```
Pantocyclus, line 26: Will this crash if a Guy does not have any fairies at all?

horrible-little-slime: yes. I'm very fine with that

src/familiar/itemFamiliar.ts
horrible-little-slime: This is duplicating existing functionality; I do think that's a reasonable shout, especially because a decked out account is hitting something like 50 lbs without breaking a sweat, but it's hard to decide what exactly is optimal. We could start using `+switch` in our maximizer strings, but that would be problematic with the way outfit caching works

src/familiar/itemFamiliar.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (have($familiar`Trick-or-Treating Tot`) && have($item`li'l ninja costume`)) {
+    return $familiar`Trick-or-Treating Tot`;
+  }
+
+  if (
+    have($familiar`Steam-Powered Cheerleader`) &&
+    findFairyMultiplier($familiar`Steam-Powered Cheerleader`) > 1.25
+  ) {
+    return $familiar`Steam-Powered Cheerleader`;
+  }
```
Pantocyclus, line 17: If we are specifically handling cheerleader, should we also separately handle mutant fire ant?

horrible-little-slime: No; cheerleader changes over the course of the day, which is bad with caching. fire ant changes over the course of the month/week/whatever, and thus can be cached properly just fine and dandy.

That being said, we could add handling to make sure the cheerleader is in fact the best fairy.

Pantocyclus: I thought that the multipliers were wonky (read: different from a weight multiplier) for mutant familiars, which was a massive headache in the handling in yachtzee. 
It appears that we do not in fact use multipliers for calculations in yachtzee anymore specifically for this reason. I'm not sure if this would be properly captured here (or for meatFamiliar for that matter)

src/familiar/itemFamiliar.ts
Pantocyclus: Could we not first choose the best fairy, then calculate if that best fairy would beat out a ninja tot? 
I suppose that would require outfit caching which is not done here; but a clear cut case would be one where our best fairy beats out the ninja tot pre-maximizer (i.e. in our current outfit), then clearly we shouldn't be using the tot over that other option.

src/familiar/itemFamiliar.ts
horrible-little-slime: [here](https://github.com/Loathing-Associates-Scripting-Society/libram/blob/d203a4bc24ea6752de3934073b16263ccfb395aa/src/lib.ts#L679-L687) is where findFairyMultiplier is defined; while you're right that the efficacy multiplier is different from the weight multiplier, but I still think treating them as more-or-less the same is "basically" fine

src/familiar/itemFamiliar.ts
horrible-little-slime: I think using `+switch` in our maximizer string is probably the easist way to achieve that

horrible-little-slime: This is what current behavior is, so I'm going to merge as-is, but it's something we can address in a little bit. I just will never merge it if I keep widening the scope of this PR.

PR #1048, opened 2022-08-28
MoodWrapper to handle blood bond beating us up

Pantocyclus:


horrible-little-slime (commented):
Just make this a libram PR instead, my dude

Also as this currently stands it doesn't appear to remove beaten up if we get beaten up doing blood bond

Pantocyclus:
Does it not? We do a `safeRestore()` after executing, but pass an argument to prevent it from throwing

horrible-little-slime:
Probably obsoleted by #1053 but lmk

PR #1049, opened 2022-08-28
Support shorter length chains

Pantocyclus:
Automatically triggered if we cannot get sufficient turns of synth after a full length chain


src/yachtzee.ts
```diff
@@ -612,6 +618,18 @@ function optimizeForFishy(yachtzeeTurns: number, setup?: boolean): number {
         }
       },
     },
+    {
+      name: "Just Fishy Pipe",
+      cost:
+        Math.max(toInt(!haveFishyPipe) * haveEffect($effect`Fishy`), toInt(haveFishyPipe) * 10) +
+          5 * toInt(havePYECCharge) <
```
horrible-little-slime, line 625:
```suggestion
      (haveFishyPipe ? 10 : haveEffect($effect`fishy`)) + (havePYECCharge ? 5 : 0) <
```

horrible-little-slime (approved):

PR #1050, opened 2022-08-28
Tidy up clover opportunity cost & PYEC code

horrible-little-slime:


Pantocyclus (approved):
Untested, but seems correct

PR #1051, opened 2022-08-29
Add love songs and elemental vials to mutually exclusive effects list

Pantocyclus:


horrible-little-slime:
This won't be a blocker, but one thing to note is that we don't currently sort the variablemeatpotions, so if we were to add a pair of mutually exclusive potions, it would end up just picking whichever is first in the list. For regular potions, because we sort the array, it's a little smarter about it.

horrible-little-slime (approved):

PR #1052, opened 2022-08-29
Update CLI messaging

horrible-little-slime:


horrible-little-slime:
It's good enough just to know that I can

PR #1053, opened 2022-08-29
Update libram to v0.6.24

horrible-little-slime:


Pantocyclus (approved):
Blood bond fix works
No issue with Libram upgrade

PR #1055, opened 2022-08-30
use visitUrl() for ballot usage

Pantocyclus:


src/dailies.ts
```diff
@@ -145,7 +149,53 @@ export function postFreeFightDailySetup(): void {
...
+    );
+
+    const ballotValue =
+      sum(
+        Object.values(availableInitiatives)
+          .sort((a, b) => b - a)
+          .slice(1),
+        (val) => 2 * val
+      ) +
+      3 * get("garbo_valueOfFreeFight", 2000);
```
horrible-little-slime, line 190: Consider breaking this into 3 variables; `initiativeValue`, `fightsValue`, and `const ballotValue = initiativeValue + fightsValue;`. This is far from a dealbreaker, but if you end up having to make other changes, make that one as well.

src/dailies.ts
```diff
@@ -145,7 +149,53 @@ export function postFreeFightDailySetup(): void {
...
+
+    if (
+      ballotValue > mallPrice($item`absentee voter ballot`) &&
+      acquire(1, $item`absentee voter ballot`, ballotValue)
```
horrible-little-slime, line 194:
```suggestion
      acquire(1, $item`absentee voter ballot`, ballotValue, false)
```
we really need to update the behavior of `acquire`

horrible-little-slime (approved):

PR #1056, opened 2022-08-30
Better dynamically-stepped yachtzee

Pantocyclus:
We lose significant gains from chaining if we are not able to cover our barfTurns with synth all the way till the end of day.

Previous estimatedTurns() did not factor in organ cleansers, shotglass, borrowed time and chocolates, which massively underestimated the number of turns of synth / organ space we needed to fully cover our run till the end of day.

PR also adds an additional chain-length of 20/25, which fits nicely with pocket wishes (our most expensive buff), allowing us to potentially earn more from a longer chain while also fully benefitting from pocket wish buffs.

### DIFF
--- a/src/embezzler.ts
+++ b/src/embezzler.ts
@@ -912,11 +912,29 @@ export function estimatedTurns(): number {
   const thumbRingMultiplier = usingThumbRing() ? 1 / 0.96 : 1;
 
   // We need to estimate adventures from our organs if we are only dieting after yachtzee chaining
-  const fullnessAdventures = (fullnessLimit() - myFullness()) * 8;
-  const inebrietyAdventures = (inebrietyLimit() - myInebriety()) * 7;
+  const distentionPillSpace = have($item`distention pill`) && !get("_distentionPillUsed") ? 1 : 0;
+  const syntheticPillSpace =
+    have($item`synthetic dog hair pill`) && !get("_syntheticDogHairPillUsed") ? 1 : 0;
+  const shotglassSpace = have($item`mime army shotglass`) && !get("_mimeArmyShotglassUsed") ? 1 : 0;
+  const sweatSpace = have($item`designer sweatpants`) ? 3 - get("_sweatOutSomeBoozeUsed") : 0;
+  const fullnessAdventures = (fullnessLimit() - myFullness() + distentionPillSpace) * 8;
+  const inebrietyAdventures =
+    (inebrietyLimit() - myInebriety() + syntheticPillSpace + sweatSpace + shotglassSpace) * 7;
+  const borrowedTimeAdventures = globalOptions.ascending && !get("_borrowedTimeUsed") ? 20 : 0;
+  const chocolateAdventures = ((3 - get("_chocolatesUsed")) * (4 - get("_chocolatesUsed"))) / 2;
+  const yachtzeeTurns = 30; // guesstimate
+  const bufferAdventures = 30; // We don't know if garbo would decide to use melange/voraci tea/sweet tooth to get more adventures
   const adventuresAfterChaining =
     globalOptions.yachtzeeChain && !get("_garboYachtzeeChainCompleted")
-      ? Math.max(fullnessAdventures + inebrietyAdventures - 30, 0)
+      ? Math.max(
+          fullnessAdventures +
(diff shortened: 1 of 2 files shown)


src/embezzler.ts
```diff
@@ -912,11 +912,29 @@ export function estimatedTurns(): number {
...
-      ? Math.max(fullnessAdventures + inebrietyAdventures - 30, 0)
+      ? Math.max(
+          fullnessAdventures +
+            inebrietyAdventures +
+            borrowedTimeAdventures +
+            chocolateAdventures +
+            bufferAdventures -
+            yachtzeeTurns,
+          0
+        )
```
horrible-little-slime, line 937: Could we slice all of the various organ space predictors into another function so that when we're estimating turns it appears as one big thing?

Maybe also slice out the chocolate and borrowed item stuff?

After that, add double-spacing in places where the vibe feels right to make it easier to follow.

Really, just SLAP it a bit.

src/yachtzee.ts
```diff
@@ -673,9 +675,11 @@ export function yachtzeeChainDiet(simOnly?: boolean): boolean {
     (3 - get("_sweatOutSomeBoozeUsed", 0));
   const spleenAvailable = currentSpleenLeft + filters;
   const organsAvailable = fullnessAvailable + inebrietyAvailable + spleenAvailable;
-  const sufficientOrgans =
+  const sufficientOrgansFor30 =
     organsAvailable - synthCastsToCoverRun - extroSpleenSpace >= 30 + 5 * toInt(havePYECCharge);
-  const baseYachtzeeTurns = sufficientOrgans ? 30 : 10;
+  const sufficientOrgansFor20 =
+    organsAvailable - synthCastsToCoverRun - extroSpleenSpace >= 20 + 5 * toInt(havePYECCharge);
```
horrible-little-slime, line 681: If we plan on adding more than these 2 chain lengths, i would structure these two into an object. Something like
```
const sufficientOrgans = { 20: [long mathy expression], 30: [slightly different expression] };
```

Alternately, SLAP this logic out into another function. I think that would be nice because instead of using the nested ternary on line 682, you can use early returns.

Hell, actually,


```
const cleanableSpleen = organsAvailable - synthCastsToCoverRun - extroSpleenSpace; // Rename this variable to something better if you can 
const sufficientOrgansFor = (yachtzees: number) => cleanableSpleen  >= yachtzees + adventureExtensionBonus;
```
I really do not like the use of `toInt(boolean)` here, and great news, you don't need to use it, we already have a variable for it.

horrible-little-slime (changes requested):

src/yachtzee.ts
```diff
@@ -675,11 +675,14 @@ export function yachtzeeChainDiet(simOnly?: boolean): boolean {
...
-    organsAvailable - synthCastsToCoverRun - extroSpleenSpace >= 20 + 5 * toInt(havePYECCharge);
-  const baseYachtzeeTurns = sufficientOrgansFor30 ? 30 : sufficientOrgansFor20 ? 20 : 10;
+
+  const cleanableSpleen = organsAvailable - synthCastsToCoverRun - extroSpleenSpace;
+  const sufficientOrgansFor = (yachtzees: number) =>
+    cleanableSpleen >= yachtzees + (havePYECCharge ? 5 : 0);
+  const possibleBaseYachtzeeTurns = [30, 20, 10];
+  const baseYachtzeeTurns = possibleBaseYachtzeeTurns.reduce((a, b) =>
+    sufficientOrgansFor(a) ? a : b
+  );
```
horrible-little-slime, line 685: This feels more like a `find` than a `reduce` situation;

```
const baseYachtzeeTurns = [30, 20, 10].find(sufficientOrgansFor) ?? 0;
```
Or something like that. This is assuming that 0 is the default. You could also do 
```
const baseYachtzeeTurns = [30, 20].find(sufficientOrgansFor) ?? 10;
```

It won't let me make this as a formal suggestion

If you do this, leave an empty line after the `sufficientOrgansFor` function, I think it'll be cleaner looking.

horrible-little-slime (dismissed):
Approved. I gave a suggestion, but follow it only if the spirit compels you.

horrible-little-slime (approved):
hell yeah

PR #1057, opened 2022-09-02
Restructure barfTurn

horrible-little-slime:


src/barfturn.ts
```diff
@@ -0,0 +1,270 @@
...
+
+const sober = () => myInebriety() <= inebrietyLimit();
+
+type Turn = {
+  name: string;
+  available: () => boolean;
+  perform: () => boolean;
+};
+
+const turns: Turn[] = [
```
docrostov, line 76: okay so this is just like, a big ol list of turns where you want to do them, and you want to check availability? 

while i think it makes some sense i have a few questions. okay, more than a few questions.

- does order matter here? was this ordered specifically or just randomly?
- "perform" feels like a weird way to name that part of it; runTurn or runAdventure feels better, "perform" just seems like a technically accurate but not fully definied thing that adds a level of obfuscation
- i would probably add a docstring explaining the order piece and how to add new things to this in the future, as you will do this pretty often i expect

src/barfturn.ts
```diff
@@ -0,0 +1,270 @@
...
+  {
+    name: "Lights Out",
+    available: () =>
+      totalTurnsPlayed() % 37 === 0 && totalTurnsPlayed() !== get("lastLightsOutTurn"),
+    perform: () => {
+      const steveRoom = get("nextSpookyravenStephenRoom");
```
docrostov, line 82: have we tested whether or not names cause any issues w/ using this pref this way? ensured the pref's wording doesn't like, have an extra the or removes a the etc? i would imagine we have tested this but i must ask.

src/barfturn.ts
```diff
@@ -0,0 +1,270 @@
...
+        const fightingSteve = steveRoom === $location`The Haunted Laboratory`;
+        // Technically drops 500 meat, but that's close enough for me.
+        if (fightingSteve) embezzlerPrep();
+        const plan = steveAdventures.get(steveRoom);
+        if (plan) {
+          withMacro(
+            Macro.if_($monster`Stephen Spookyraven`, Macro.basicCombat()).abort(),
+            () => {
+              visitUrl(toUrl(steveRoom));
+              for (const choiceValue of plan) {
```
docrostov, line 94: i am very confused about what plan is doing here and why you're accessing it. what's the point of "plan" rather than just runChoice within haunted laboratory. maybe this would make more sense with better var names?

src/barfturn.ts
```diff
@@ -0,0 +1,270 @@
...
+        }
+        return fightingSteve;
+      }
+      return false;
+    },
+  },
+  {
+    name: "Proton Ghost",
+    available: () =>
+      sober() &&
```
docrostov, line 110: i don't get why we are doing the sober check here. even if you can't shoot ghost, you should be able to still kill it if you're running while overdrunk w/ the wine glass. you can't ghostbust but it's still a free fight to increment tiny bits and bobs of value.

src/barfturn.ts
```diff
@@ -0,0 +1,270 @@
...
+      return get("_sourceTerminalDigitizeMonsterCount") !== start;
+    },
+  },
+  {
+    name: "Guaranteed Kramco",
+    available: () => sober() && kramcoGuaranteed(),
+    perform: () => {
+      freeFightPrep(new Requirement([], { forceEquip: $items`Kramco Sausage-o-Matic™` }));
+      adventureMacroAuto(determineDraggableZoneAndEnsureAccess(), Macro.basicCombat());
+      return !kramcoGuaranteed();
```
docrostov, line 176: what is this return actually doing for us in any way

src/barfturn.ts
```diff
@@ -0,0 +1,270 @@
...
+  },
+  {
+    name: "Barf",
+    available: () => true,
+    perform: () => {
+      useFamiliar(barfFamiliar());
+      const lubing = get("dinseyRollercoasterNext") && have($item`lube-shoes`);
+      meatOutfit(
+        false,
+        lubing ? new Requirement([], { forceEquip: $items`lube-shoes` }) : undefined
```
docrostov, line 230: what is undefined here in the meatoutfit doing

src/barfturn.ts
```diff
@@ -0,0 +1,270 @@
...
+    },
+  },
+];
+
+export default function barfTurn(): void {
+  if (SourceTerminal.have()) SourceTerminal.educate([$skill`Extract`, $skill`Digitize`]);
+
+  tryFillLatte();
+  meatMood().execute(estimatedTurns());
+  safeRestore();
```
docrostov, line 251: is this dependent on universal recovery or is this just base game junk?

docrostov (commented):
it seems fine to me i guess, but i will candidly say there's a bunch of stuff here i don't get that probably could stand to be better documented. it really does look like you built `turns` in encounter ontology order and that is extremely important information future script editors emphatically need to know.

src/barfturn.ts
horrible-little-slime: Garbo has been doing this since May 22nd.

src/barfturn.ts
horrible-little-slime: I will probably attach most of the documentation to the `Turn` type, but that tracks. I'll run some language by you.

src/barfturn.ts
horrible-little-slime: We have historically always checked for sobriety because some people have trouble killing physically resistant monsters with only weapon attacks. It's worth re-investigating.

src/barfturn.ts
horrible-little-slime: it's saying "if we aren't lubing, we aren't passing a Requirement here"

src/barfturn.ts
horrible-little-slime: We import this from `lib`

src/barfturn.ts
horrible-little-slime: We need the plan because many of these adventures will lead through several phases of the choice without changing choiceID, because the game is dumb.

It's essentially a "plan" for what choice options we're going to submit in what order.

src/barfturn.ts
horrible-little-slime: This might not actually be useful; I want to have a check to make sure people don't end up in some sort of weird infinite loop, where they try to take an action but never actually do so, thereby never changing the value of the `available` function. So the idea is we return a value that tells us "did we end up going through with this?"

A failure here could range from "jesus christ what" to "oh, yeah, hiccup, just do the next thing." For the most part, we just do the next thing here.

src/barfturn.ts
phulin: we should have `meatOutfit` take an options object rather than the current structure probably, just for clarity

src/barfturn.ts
```diff
@@ -0,0 +1,284 @@
...
+
+type Turn = {
+  name: string;
+
+  // Can we take this turn-action right now?
+  available: () => boolean;
+
+  // Take the turn-action. Returns whether we succeeded in doing so.
+  execute: () => boolean;
+};
```
phulin, line 90: I'd prefer jsdoc format here so VS Code will turn it into something nice

src/index.ts
```diff
@@ -121,201 +86,6 @@ function ensureBarfAccess() {
...
-  if (
-    have($item`designer sweatpants`) &&
-    myAdventures() === 1 + globalOptions.saveTurns &&
-    !globalOptions.noDiet
-  ) {
-    while (get("_sweatOutSomeBoozeUsed", 0) < 3 && get("sweat", 0) >= 25 && myInebriety() > 0) {
-      useSkill($skill`Sweat Out Some Booze`);
-    }
-    consumeDiet(computeDiet().sweatpants(), "SWEATPANTS");
-  }
```
phulin, line 306: the above code all didn't make it over, I think

phulin (commented):
I'd also capitalize the filename barfTurn

src/barfturn.ts
phulin: we should put a warning out when a task fails i think

src/barfturn.ts
```diff
@@ -0,0 +1,284 @@
...
+        (getModifier("Adventure Underwater") ||
+          waterBreathingEquipment.some((item) => have(item))) &&
+        (getModifier("Underwater Familiar") ||
+          familiarWaterBreathingEquipment.some((item) => have(item))) &&
+        (have($effect`Fishy`) || (have($item`fishy pipe`) && !get("_fishyPipeUsed"))) &&
+        !have($item`envyfish egg`) &&
+        canAdventure($location`The Briny Deeps`);
+      const targetLocation = shouldGoUnderwater
+        ? $location`The Briny Deeps`
+        : determineDraggableZoneAndEnsureAccess();
```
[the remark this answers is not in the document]
horrible-little-slime: This is some legacy code, but there are players who will encounter so few embezzlers that they won't hit their first digitize during their initial push. It is a pretty niche case, though, that might be worth ceasing to support. Especially because we aren't actually checking if the envyfish egg is worth it.

src/barfturn.ts
horrible-little-slime: Just printing, or erroring out? I don't think we should error out just because there are enough dumb cases where we hit zone intro adventures.

src/barfturn.ts
horrible-little-slime: I will probably start to handle that as I work on stateful outfit code. Which may or may not happen in the near future.

src/barfturn.ts
```diff
@@ -0,0 +1,284 @@
...
+  meatMood().execute(estimatedTurns());
+  safeRestore();
+
+  const startTurns = totalTurnsPlayed();
+  for (const turn of turns) {
+    if (turn.available()) {
+      const success = turn.execute();
+      if (success) {
+        const spentATurn = totalTurnsPlayed() - startTurns === 1;
+        const foughtAnEmbezzler = get("lastEncounter") === "Knob Goblin Embezzler";
```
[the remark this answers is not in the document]
horrible-little-slime: It needs to cost a turn as wlel, which feels like it gets rid of most of the weird edge cases.

src/barfturn.ts
```diff
@@ -0,0 +1,369 @@
...
+    },
+    spendsTurn: true,
+  },
+  {
+    name: "Spit Acid",
+    available: () =>
+      sober() && have($item`Jurassic Parka`) && !have($effect`Everything Looks Yellow`),
+    execute: () => {
+      const jellyfishing = have($familiar`Space Jellyfish`) && realmAvailable("stench");
+      const familiarChoice = jellyfishing ? $familiar`Space Jellyfish` : freeFightFamiliar(true);
```
Pantocyclus, line 247: Is there a reason to intentionally run jellyfish over a general marginal familiar (which could very well be the jellyfish)?

horrible-little-slime: We don't currently have a way to pass additional familiars to freeFightFamiliar, but I suppose I should just change that

src/barfTurn.ts
```diff
@@ -0,0 +1,369 @@
...
+    },
+    spendsTurn: true,
+  },
+  {
+    name: "Spit Acid",
+    available: () =>
+      sober() && have($item`Jurassic Parka`) && !have($effect`Everything Looks Yellow`),
+    execute: () => {
+      const jellyfishing = have($familiar`Space Jellyfish`) && realmAvailable("stench");
+      const familiarChoice = jellyfishing ? $familiar`Space Jellyfish` : freeFightFamiliar(true);
```
Pantocyclus, line 247: Just a reminder (it's probably already in your list of considerations) that after the small revamp, `jellyfishing` should be conditioned upon `familiarChoice`, not this other way around (i.e. `jellyfishing = familiarChoice === $familiar'Space Jellyfish'`)

horrible-little-slime: I wrote draft logic that essentially went
can we jellyfish? if so, tell the function that our location is pirate barges
did we get jellyfish? if so, our location is pirate barges! else, our location is a draggable zone.

This has a bit of a flaw: what if the value of placing a wanderer would make a different familair beat the jellyfish? but also that is so marginal I cannot be arsed to care. And I'm not even british!

s-k-z:
since sober() shows up so frequently, what if it was a separate flag on AdventureAction? Something like
```ts
enum sobrietyRequirement {
either,
soberOnly,
drunkOnly,
}

PR #1061, opened 2022-09-04
Bell spikes yachtzee

horrible-little-slime:
Depends on barf-turn changes

### DIFF
--- a/src/yachtzee/buffs.ts
+++ b/src/yachtzee/buffs.ts
@@ -1,5 +1,27 @@
-import { cliExecute, Effect, haveEffect, itemAmount, maximize, myMeat, print, toInt, use } from "kolmafia";
-import { $effect, $item, $items, clamp, get, getActiveEffects, getActiveSongs, have, isSong, Mood, set } from "libram";
+import {
+  cliExecute,
+  Effect,
+  haveEffect,
+  itemAmount,
+  maximize,
+  myMeat,
+  print,
+  toInt,
+  use,
+} from "kolmafia";
+import {
+  $effect,
+  $item,
+  $items,
+  clamp,
+  get,
+  getActiveEffects,
+  getActiveSongs,
+  have,
+  isSong,
+  Mood,
+  set,
+} from "libram";
 import { acquire } from "../acquire";
 import { withStash } from "../clan";
 import { baseMeat, burnLibrams, turnsToNC } from "../lib";
--- a/src/yachtzee/buffs.ts
+++ b/src/yachtzee/buffs.ts
@@ -1,0 +1,215 @@
+import { cliExecute, Effect, haveEffect, itemAmount, maximize, myMeat, print, toInt, use } from "kolmafia";
+import { $effect, $item, $items, clamp, get, getActiveEffects, getActiveSongs, have, isSong, Mood, set } from "libram";
+import { acquire } from "../acquire";
+import { withStash } from "../clan";
+import { baseMeat, burnLibrams, turnsToNC } from "../lib";
+import { farmingPotions, mutuallyExclusive, Potion, variableMeatPotionsSetup } from "../potions";
+import { garboValue } from "../session";
+import { executeNextDietStep } from "./diet";
+import { expectedEmbezzlers, pyecAvailable, shrugIrrelevantSongs } from "./lib";
+
+function yachtzeePotionProfits(potion: Potion, yachtzeeTurns: number): number {
+  // If we have an unused PYEC then
+  // 1) If we don't have an effect, +5 to gained effect duration
+  // 2) If we already have an effect, +5 to existing effect duration
+  // This means that the first use of a potion that we don't already have an effect of is more valuable than the next use
+  const PYECOffset = pyecAvailable() ? 5 : 0;
+  const existingOffset = haveEffect(potion.effect()) ? PYECOffset : 0;
+  const extraOffset = PYECOffset - existingOffset;
+  const effectiveYachtzeeTurns = Math.max(
+    Math.min(
+      yachtzeeTurns - haveEffect(potion.effect()) - existingOffset,
+      potion.effectDuration() + extraOffset
+    ),
+    0
+  );
+  const embezzlerTurns = Math.min(
+    expectedEmbezzlers,
+    Math.max(potion.effectDuration() + extraOffset - effectiveYachtzeeTurns, 0)
+  );
+  const barfTurns = Math.max(
+    potion.effectDuration() + extraOffset - effectiveYachtzeeTurns - embezzlerTurns,
+    0
+  );
+  const embezzlerValue = embezzlerTurns > 0 ? potion.gross(embezzlerTurns) : 0;
+  const yachtzeeValue =
+    (effectiveYachtzeeTurns * 2000 * (potion.meatDrop() + 2.5 * potion.familiarWeight())) / 100; // Every 1lbs of lep ~ 2.5% meat drop
+  const barfValue = (barfTurns * baseMeat * turnsToNC) / (turnsToNC + 1);
+
+  return yachtzeeValue + embezzlerValue + barfValue - potion.price(true);
+}
+
+export function yachtzeePotionSetup(yachtzeeTurns: number, simOnly?: boolean): number {
+  let totalProfits = 0;
+  const PYECOffset = pyecAvailable() ? 5 : 0;
+  const excludedEffects = new Set<Effect>();
+
+  shrugIrrelevantSongs();
+
+  if (have($item`Eight Days a Week Pill Keeper`) && !get("_freePillKeeperUsed", false)) {
+    const doublingPotions = farmingPotions
+      .filter(
+        (potion) =>
+          potion.canDouble &&
+          haveEffect(potion.effect()) + PYECOffset * (have(potion.effect()) ? 1 : 0) <
+            yachtzeeTurns &&
+          yachtzeePotionProfits(potion.doubleDuration(), yachtzeeTurns) > 0 &&
+          potion.price(true) < myMeat()
+      )
+      .sort(
+        (left, right) =>
+          yachtzeePotionProfits(right.doubleDuration(), yachtzeeTurns) -
+          yachtzeePotionProfits(left.doubleDuration(), yachtzeeTurns)
+      );
+    const bestPotion = doublingPotions.length > 0 ? doublingPotions[0].doubleDuration() : undefined;
+    if (bestPotion) {
+      const profit = yachtzeePotionProfits(bestPotion, yachtzeeTurns);
+      const price = bestPotion.price(true);
+      totalProfits += profit;
+      print(`Determined that ${bestPotion.potion} was the best potion to double`, "blue");
+      print(
+        `Expected to profit ${profit} meat from doubling 1 ${bestPotion.potion} @ price ${price} meat`,
+        "blue"
+      );
+      if (!simOnly) {
+        cliExecute("pillkeeper extend");
+        acquire(1, bestPotion.potion, profit + price);
+        bestPotion.use(1);
+      } else excludedEffects.add(bestPotion.effect());
+    }
+  }
+
+  for (const effect of getActiveEffects()) {
+    for (const excluded of mutuallyExclusive.get(effect) ?? []) {
+      excludedEffects.add(excluded);
+    }
+  }
+
+  const testPotions = farmingPotions
+    .filter(
+      (potion) =>
+        haveEffect(potion.effect()) + PYECOffset * toInt(haveEffect(potion.effect()) > 0) <
+          yachtzeeTurns && yachtzeePotionProfits(potion, yachtzeeTurns) > 0
+    )
+    .sort(
+      (left, right) =>
+        yachtzeePotionProfits(right, yachtzeeTurns) - yachtzeePotionProfits(left, yachtzeeTurns)
+    );
+
+  for (const potion of testPotions) {
+    const effect = potion.effect();
+    const price = potion.price(true);
+    if (
+      haveEffect(effect) + PYECOffset * toInt(haveEffect(effect) > 0) >= yachtzeeTurns ||
+      price > myMeat()
+    ) {
+      continue;
+    }
+    if (!excludedEffects.has(effect)) {
+      let tries = 0;
+      while (haveEffect(effect) + PYECOffset * toInt(haveEffect(effect) > 0) < yachtzeeTurns) {
+        tries++;
+        print(`Considering effect ${effect} from source ${potion.potion}`, "blue");
+        const profit = yachtzeePotionProfits(potion, yachtzeeTurns);
+        if (profit < 0) break;
+        const nPotions = have(effect)
+          ? clamp(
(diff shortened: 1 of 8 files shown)


src/yachtzee/diet.ts
```diff
@@ -0,0 +1,692 @@
...
+          use(Math.min(n, itemAmount($item`munchies pill`)), $item`munchies pill`);
+        }
+        eat(n, $item`toast with stench jelly`);
+      }),
+      new YachtzeeDietEntry("jumping horseradish", 0, 1, 0, 0, (n: number) => {
+        ensureConsumable("jumping horseradish", n, 1, 0, 0);
+        eat(n, $item`jumping horseradish`);
+      }),
+      new YachtzeeDietEntry("clara's bell", 0, 0, 0, 0, () => use($item`Clara's bell`)),
+      new YachtzeeDietEntry("jurassic parka", 0, 0, 0, 0, useSpikolodonSpikes),
```
Pantocyclus, line 136: I think we can potentially support `sweat out some booze` here

src/yachtzee/diet.ts
```diff
@@ -0,0 +1,692 @@
...
+          (idx > 0 &&
+            dietSchedule[idx - 1].spleen < 0 &&
+            spleenUse + dietSchedule[idx].spleen >= 0)) // And cluster spleen cleansers (continue if the next cleanser can still clean our spleen)
+      ) {
+        spleenUse += dietSchedule[idx++].spleen ?? 0;
+      }
+      dietSchedule.splice(idx, 0, entry);
+    }
+  }
+
```
Pantocyclus, line 349: To support `sweat out some booze`, we can use the same loop as the overconsumption check below. The moment our inebriety is larger than the number of sweats `(x)` we get to use, insert `x` sweat outs

src/yachtzee/diet.ts
```diff
@@ -0,0 +1,692 @@
...
+  const maxYachtzeeTurns = havePYECCharge ? baseYachtzeeTurns + 5 : baseYachtzeeTurns;
+  print(`Synth Casts Wanted: ${synthCastsToCoverRun}`, "blue");
+  print(`Organs Available: ${organsAvailable}`, "blue");
+  print(`Max Yachtzee Turns: ${maxYachtzeeTurns}`, "blue");
+
+  // Plan our diet
+
+  const sliders = Math.floor((fullnessLimit() + toInt(haveDistentionPill) - myFullness()) / 5);
+  const pickleJuice = Math.floor((inebrietyLimit() - myInebriety()) / 5);
```
Pantocyclus, line 447: Once we support sweat out and synthetic pill, we have to add their contributions here (just like the `haveDistentionPill`) contribution to sliders above

src/yachtzee/diet.ts
```diff
@@ -0,0 +1,692 @@
...
+            throw new Error(`consuming ${entry.name} will exceed our spleen limit`);
+          }
+          if (entry.fullness > 0) {
+            if (!get("_milkOfMagnesiumUsed")) {
+              acquire(1, $item`milk of magnesium`, 5 * VOA);
+              use(1, $item`milk of magnesium`);
+            }
+            if (!get("_distentionPillUsed") && have($item`distention pill`)) {
+              use(1, $item`distention pill`);
+            }
```
Pantocyclus, line 272: Finally, support for sweat outs and synthetic pills can be triggered here once we have sufficient inebriety. This should be an "always on" feature much like the distention pill.

src/yachtzee/diet.ts
```diff
@@ -0,0 +1,692 @@
...
+    ["beggin' cologne", cologneToChew],
+    ["jumping horseradish", horseradishes],
+  ];
+
+  const specialEntries: [string, number, (n: number, name?: string) => void][] = (
+    [
+      ["stench jelly", jelliesToChew],
+      ["toast with stench jelly", toastsToEat],
+      ["clara's bell", have($item`Clara's bell`) && !globalOptions.clarasBellClaimed ? 1 : 0],
+      ["jurassic parka", have($item`Jurassic Parka`) ? get("_spikolodonSpikeUses") : 0],
```
Pantocyclus, line 623: I assume this should be `5 - get("_spikolodonSpikeUses")`

src/yachtzee/fishy.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  // Restore here if we potentially need to visit an adventure.php zone to grab fishy turns
+  if (haveEffect($effect`Beaten Up`)) {
+    uneffect($effect`Beaten Up`);
+  }
+  safeRestore();
+
+  const haveFishyPipe = have($item`fishy pipe`) && !get("_fishyPipeUsed");
+  const adventureExtensionBonus = pyecAvailable() ? 5 : 0;
+  const fishySources = [
```
Pantocyclus, line 76: I've actually thought about changing this structure slightly - I'd like a field called `baseFishyTurns` or `fishyTurns` (with/out consideration of PYEC) so that we don't need to change the cost to infinity, which is (1) ugly, and (2) prone to making stupid mistakes like I did at the start when adding pure fishy pipe support for shorter chains

src/yachtzee/fishy.ts
```diff
@@ -0,0 +1,203 @@
...
+          : 0,
+      action: () => {
+        if (haveFishyPipe && haveEffect($effect`Fishy`) < yachtzeeTurns) use(1, $item`fishy pipe`);
+      },
+    },
+  ];
+
+  const bestFishySource = fishySources.reduce((left, right) => {
+    return left.cost < right.cost ? left : right;
+  });
```
Pantocyclus, line 192: Correspondingly, if you're refactoring the above to add the `fishyTurns` field, we would conduct the filtering here to remove sources that don't provide sufficient fishy turns to cover our yachtzee turns

src/yachtzee/lib.ts
```diff
@@ -0,0 +1,102 @@
...
+  useFamiliar(familiar);
+  const mergedRequirements = new Requirement([], { forceEquip: $items`Jurassic Parka` }).merge(
+    run.constraints.equipmentRequirements?.() ?? new Requirement([], {})
+  );
+  freeFightOutfit(mergedRequirements);
+  cliExecute("parka spikolodon");
+
+  const targetZone = canJelly
+    ? $location`Pirates of the Garbage Barges`
+    : $location`The Haunted Kitchen`;
```
Pantocyclus, line 96: We may want to check if we `canAdventure(targetZone)`, if for some reason mafia didn't automatically use the telegram from spookyraven (or if there is a path with no access to the Right Side of the Tracks etc). 
It's not a major issue, but it's really trivial to add I guess.
The other thing is to ensure that we have proper handling for the lights out quest (I'm not sure if that is already globally set)

src/yachtzee/lib.ts
```diff
@@ -0,0 +1,102 @@
...
+  (have($item`Clara's bell`) && !globalOptions.clarasBellClaimed ? 1 : 0) +
+  (have($item`Jurassic Parka`) ? 5 - get("_spikolodonSpikeUses") : 0);
+
+export function yachtzeeBuffValue(obj: Item | Effect): number {
+  return (2000 * (getModifier("Meat Drop", obj) + getModifier("Familiar Weight", obj) * 2.5)) / 100;
+}
+
+export function useSpikolodonSpikes(): void {
+  if (get("_spikolodonSpikeUses") >= 5) return;
+  const run = tryFindFreeRun() ?? ltbRun();
```
Pantocyclus, line 81: Are we saving free runs for this? It might be prudent to head into `fights.ts` to disable all the FreeRunFights with jellyfish if we are chaining and have spikolodon spikes available.

src/yachtzee/lib.ts
```diff
@@ -0,0 +1,102 @@
...
+  cliExecute("parka spikolodon");
+
+  const targetZone = canJelly
+    ? $location`Pirates of the Garbage Barges`
+    : $location`The Haunted Kitchen`;
+  const macro = Macro.familiarActions().step(run.macro);
+  const startingSpikes = get("_spikolodonSpikeUses");
+  do {
+    adventureMacroAuto(targetZone, macro);
+  } while (get("_spikolodonSpikeUses") === startingSpikes);
```
Pantocyclus, line 101: If we are in fact going to reserve jellyfish FreeRunFights as per the above, then we can burn whatever remains here once `get("_spikolodonSpikeUses") >= 5`

src/yachtzee/buffs.ts
```diff
@@ -0,0 +1,215 @@
...
+      if (have(effect) || simOnly) {
+        for (const excluded of mutuallyExclusive.get(effect) ?? []) {
+          excludedEffects.add(excluded);
+        }
+      }
+    }
+  }
+
+  if (!simOnly) {
+    variableMeatPotionsSetup(yachtzeeTurns, expectedEmbezzlers);
```
Pantocyclus, line 160: I have yet to add the profits from variable meat potions to the `totalProfits`, similar to Greenspans, but I think this is a much easier task than the latter.
Not a major issue, but you can consider adding it if you'd like (note that we would then need to pass a simOnly flag into `variableMeatPotionsSetup` because as of now it automatically triggers the acquisition and usage of the potions)

src/yachtzee/fishy.ts
Pantocyclus: Also, as acco has noted in passing, when we are considering shorter chains (e.g. base 10 or 20), we should also consider pocket wishes as a fishy source - they may very well be optimal

src/yachtzee/index.ts
```diff
@@ -0,0 +1,151 @@
...
+      set("_stenchJellyUsed", false);
+    }
+    if (
+      plantCrookweed &&
+      visitUrl("forestvillage.php").includes("friarcottage.gif") &&
+      !get("_floristPlantsUsed").split(",").includes("Crookweed")
+    ) {
+      cliExecute("florist plant Crookweed");
+    }
+    plantCrookweed = false;
```
Pantocyclus, line 133: Any feel on when florist support for Libram will be completed? `barfTurn` could also use some plants for QOL (e.g. the mp restoring and item drop ones)

src/yachtzee/buffs.ts
```diff
@@ -0,0 +1,215 @@
...
+    expectedEmbezzlers,
+    Math.max(potion.effectDuration() + extraOffset - effectiveYachtzeeTurns, 0)
+  );
+  const barfTurns = Math.max(
+    potion.effectDuration() + extraOffset - effectiveYachtzeeTurns - embezzlerTurns,
+    0
+  );
+  const embezzlerValue = embezzlerTurns > 0 ? potion.gross(embezzlerTurns) : 0;
+  const yachtzeeValue =
+    (effectiveYachtzeeTurns * 2000 * (potion.meatDrop() + 2.5 * potion.familiarWeight())) / 100; // Every 1lbs of lep ~ 2.5% meat drop
```
Pantocyclus, line 36: I have been considering for the longest time to add the value of `shrimp cocktail` to this. Even if the mallprice (5k) isn't reflective of what it sells for, it is at the very least worth 5lbs * 20 turns (5 * 0.025 * baseMeat * 20 = 687.5 meat for a base lep) per potion, at a 33% drop rate. Marginal, but not insignificant when it also contributes to 100s of turns yachtzeechains and embezzlers with robort.
(unfortunately the prices still remain high for some reason, so we aren't auto-using them for embezzliers and chains - but they aren't selling so idk maybe somebody needs to do a mall play with them with the stash that we've built up over the thousands of chains we've done)

src/yachtzee/diet.ts
horrible-little-slime: Out of scope of this PR

src/yachtzee/diet.ts
horrible-little-slime: Out of scope of this PR

src/yachtzee/fishy.ts
horrible-little-slime: Out of scope of this PR

src/yachtzee/lib.ts
horrible-little-slime: it's less trivial to add handling for because we would need net another backup location; the haunted kitchen is our current "default" zone to drop things into

src/yachtzee/lib.ts
horrible-little-slime: We can only use one spikes at a time

src/yachtzee/buffs.ts
horrible-little-slime: Out of scope of this PR

src/yachtzee/buffs.ts
horrible-little-slime: Out of scope of this PR

src/yachtzee/lib.ts
Pantocyclus: Yup I understand
But we can still trigger the remainder of the free runs after we are done with the 5th spike

src/yachtzee/lib.ts
horrible-little-slime: I don't know if this is a `when`, given that we have a lot of potential freerun sources. ltbRun technically also includes tennis balls and champagne poppers.

Right now, we are not actively saving freeruns for this; the issue with cutting out the jellyfish runs is that those will just get used fishing for hipsters and whatall.

What we may be able to do is push freeRunFights to after we do yachtzee, when yachtzeeing. We'll have to think about if there's a compelling reason not to do that.

src/yachtzee/lib.ts
horrible-little-slime: I want as few non-yachtzee-related things to happen in the yachtzee code as possible, for maintainability and bugreporting reasons. But I think pushing freerunfights to after yachtzee is probably a decent fix.

src/yachtzee/lib.ts
```diff
@@ -0,0 +1,102 @@
...
+const ignoredSources = [
+  "Orb Prediction",
+  "Pillkeeper Semirare",
+  "Lucky!",
+  "11-leaf clover (untapped potential)",
+];
+export const expectedEmbezzlers = sum(
+  embezzlerSources.filter((source: EmbezzlerFight) => !ignoredSources.includes(source.name)),
+  (source: EmbezzlerFight) => source.potential()
+);
```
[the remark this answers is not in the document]
horrible-little-slime: Yeah I can go ahead and fix this. While this technically refactors yachtzee as a whole, I'd like to mostly save big changes to yachtzee for later.

src/yachtzee/index.ts
horrible-little-slime: It may currently be working, and I have an unreviewed branch that started a refactor that I was unable to get any eyes on, and thus have discontinued work on.

src/yachtzee/lib.ts
Tokoeka: 1 reason I can think of not doing it is yachtzeeing will (out of neccessity) either wholesale use up (or at least partially eat into) fam weight buffs, which overall would reduce the amount of potential familiar free runs we can do. So if we were to push freeRunFights to post yachtzee, we may want to split it out with "free fam runs" pre yachtzee, and others post

PR #1062, opened 2022-09-06
Allow us to pass location and other options to freeFightFamiliar

horrible-little-slime:
@Pantocyclus this will make me able to fiddle with the jellyfish code in the barfTurn restructure more easily

### DIFF
--- a/src/familiar/freeFightFamiliar.ts
+++ b/src/familiar/freeFightFamiliar.ts
@@ -1,17 +1,35 @@
-import { Familiar, familiarWeight } from "kolmafia";
-import { $familiar, findLeprechaunMultiplier } from "libram";
+import { Familiar, familiarWeight, Location } from "kolmafia";
+import { $familiar, $item, findLeprechaunMultiplier, get, have } from "libram";
 import { canOpenRedPresent } from ".";
+import { garboValue } from "../session";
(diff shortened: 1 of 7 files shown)


src/familiar/freeFightFamiliar.ts
```diff
@@ -29,14 +47,25 @@ export function menu(includeExperienceFamiliars = true, canChooseMacro = false):
         familiar: $familiar`Crimbo Shrub`,
         expectedValue: 2500,
         leprechaunMultiplier: 0,
-        limit: "none",
+        limit: "special",
+      });
+    }
+
+    if (location.zone === "Dinseylandfill" && have($familiar`Space Jellyfish`)) {
```
Pantocyclus, line 54: Should we consider other 100% combat stench zones too? May be hard to generalize to grab all such zones, but at the very least I feel like our default stesnch zone `The Haunted Kitchen` should be considered as well

horrible-little-slime: Haunted Kitchen isn't a stench zone, though.

Pantocyclus: oh right my bad

Pantocyclus (approved):

Issue #973, opened 2022-06-29
[resource] Avatar of West of Loathing Extract Oil, Long Con and Meat Buff


Shanakor:
would be awesome if garbo could implement these 3 things.

I have no experience coding with garbo but this might be trivial to implement? (mallsearch the higher price of skin/unusual oil, autocast long con on the garbage tourist and autocast the 100% meat walk buff)

docrostov:
closing this issue because we have implemented the minimum viable product (meat buff + long con). we are doing #1522 to add extract oil; it is possible this will lead to additional beat-ups, please tell us if that is the case (and you are still in AWoL, one point five years later)

PR #1064, opened 2022-09-08
Use Long Con & Lesiurely Amble

Tokoeka:
Initial AWoL support - casts long Con on crates and garbage tourists at appropriate moments, and adds the meat walk to the meat mood

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -290,6 +290,12 @@ export class Macro extends StrictMacro {
         Macro.if_($monster`garbage tourist`, Macro.trySkill($skill`Gallapagosian Mating Call`))
       )
       .externalIf(
+        get("longConMonster") !== $monster`garbage tourist` &&
+          get("_longConUsed") < 5 &&
+          have($skill`Long Con`),
+        Macro.if_($monster`garbage tourist`, Macro.trySkill($skill`Long Con`))
+      )
+      .externalIf(
         !get("_latteCopyUsed") &&
           (get("_latteMonster") !== $monster`garbage tourist` ||
             Counter.get("Latte Monster") > 30) &&
--- a/src/extrovermectin.ts
+++ b/src/extrovermectin.ts
@@ -81,8 +81,10 @@ export function doingExtrovermectin(): boolean {
 export function crateStrategy(): "Sniff" | "Saber" | "Orb" | null {
   if (!doingExtrovermectin()) return null;
   if (
-    have($skill`Transcendent Olfaction`) &&
-    (property.getString("olfactedMonster") === "crate" || get("_olfactionsUsed") < 2)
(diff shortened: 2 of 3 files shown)


Tokoeka:
Partially implements #973 - does not consider use of extract oil as a combat skill due to using it provoking the monster to auto-hit you, which I cannot deal with coding at the moment

horrible-little-slime:
I'm not going to ask you to go AWOL to test this but we should test this. Or I guess just ping whatstheirface when we put it in and ask for bug reports

src/combat.ts
```diff
@@ -289,6 +289,12 @@ export class Macro extends StrictMacro {
           have($skill`Gallapagosian Mating Call`),
         Macro.if_($monster`garbage tourist`, Macro.trySkill($skill`Gallapagosian Mating Call`))
       )
+      .externalIf(
+        get("longConMonster") !== $monster`garbage tourist` &&
```
docrostov, line 293: can we get the log from the dude who ran this in AWOL to make sure it used long con five times? i could see this maybe not actually working if mafia's settings don't properly clear longConMonster when the effect wears off.

src/extrovermectin.ts
```diff
@@ -81,8 +81,10 @@ export function doingExtrovermectin(): boolean {
 export function crateStrategy(): "Sniff" | "Saber" | "Orb" | null {
   if (!doingExtrovermectin()) return null;
   if (
-    have($skill`Transcendent Olfaction`) &&
-    (property.getString("olfactedMonster") === "crate" || get("_olfactionsUsed") < 2)
+    (have($skill`Transcendent Olfaction`) &&
+      (property.getString("olfactedMonster") === "crate" || get("_olfactionsUsed") < 2)) ||
+    (have($skill`Long Con`) &&
```
docrostov, line 86: aw yeah we're conning the fuck outta these crates

docrostov (approved):
i am going to approve these changes but please note the one request above; i think we should check that the logic there worked and mafia's preferences properly changed when long con wore off. it is not a major issue tho and i'm fine with merging this.

docrostov:
we have a log!

https://discord.com/channels/466605739838930955/864070665615179786/1020189698355896420

i am not going to look at this today but my only question is whether the long con logic actually worked and it cast long con 5x times.

src/combat.ts
Tokoeka: Does Long Con wear off, theres nothing on the wiki page about it being limited?

src/extrovermectin.ts
Tokoeka: Having checked in the log, it only cast it once, on garbage tourists. did 0 extrovermectin, so cast it not at all on the crates.

PR #1066, opened 2022-09-09
Add estimated turns for jellyfish

Pantocyclus:
Also use shrub over meat familiar immediately if ELR is on cooldown and its totalFamiliarValue is > meatFamiliarValue (even if we haven't hit our marginal familiar turncount required yet)

### DIFF
--- a/src/familiar/marginalFamiliars.ts
+++ b/src/familiar/marginalFamiliars.ts
@@ -117,7 +117,7 @@ function turnsNeededForFamiliar(
       return 0;
 
     case "shrub":
-      return 0;
+      return Math.ceil(estimatedTurns() / 100);
 
     case "jellyfish":
       return sum(
--- a/src/familiar/marginalFamiliars.ts
+++ b/src/familiar/marginalFamiliars.ts
@@ -116,18 +116,8 @@ function turnsNeededForFamiliar(
     case "none":
       return 0;
 
-    case "shrub":
-      return Math.ceil(estimatedTurns() / 100);
-
-    case "jellyfish":
-      return sum(
-        getAllJellyfishDrops().filter(
-          ({ expectedValue }) =>
-            outfitValue + familiarAbilityValue(familiar) + expectedValue >
-            totalFamiliarValue(baselineToCompareAgainst)
-        ),
-        ({ expectedTurns }) => expectedTurns
-      );
+    case "special":
+      return getSpecialFamiliarLimit({ familiar, outfitValue, baselineToCompareAgainst });
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/familiar/marginalFamiliars.ts
```diff
@@ -153,7 +163,12 @@ export function barfFamiliar(): Familiar {
       turnsNeededForFamiliar(option, meatFamiliarEntry)
     );
 
-    if (turnsNeeded < estimatedTurns()) return meatFamiliar();
+    if (turnsNeeded < estimatedTurns()) {
+      return (
+        viableMenu.find((familiar) => familiar.familiar === $familiar`Crimbo Shrub`)?.familiar ??
+        meatFamiliar()
+      );
```
horrible-little-slime, line 170:
```suggestion
      const shrubAvailable = viableMenu.some(
        ({ familiar }) => familiar === $familiar`Crimbo Shrub`
      );
      return shrubAvailable ? $familiar`Crimbo Shrub` : meatFamiliar();
      ```

File: src/familiar/marginalFamiliars.ts [hunk 2]
```diff
@@ -116,8 +116,18 @@ function turnsNeededForFamiliar(
...
+
+    case "jellyfish":
+      return sum(
+        getAllJellyfishDrops().filter(
+          ({ expectedValue }) =>
+            outfitValue + familiarAbilityValue(familiar) + expectedValue >
+            totalFamiliarValue(baselineToCompareAgainst)
+        ),
+        ({ expectedTurns }) => expectedTurns
+      );
```
horrible-little-slime, line 130: Go back to using a case called `special`, and make a function that provides special handling for special familiars. Something like 
```
function getSpecialFamiliarLimit({
  familiar,
  baselineToCompareAgainst,
  outfitValue,
}: {
  familiar: Familiar;
  outfitValue: number;
  baselineToCompareAgainst: GeneralFamiliar & { outfitWeight: number; outfitValue: number };
}): number {
  switch (familiar) {
    case $familiar`Space Jellyfish`:
      return sum(
        getAllJellyfishDrops().filter(
          ({ expectedValue }) =>
            outfitValue + familiarAbilityValue(familiar) + expectedValue >
            totalFamiliarValue(baselineToCompareAgainst)
        ),
        ({ expectedTurns }) => expectedTurns
      );

    case $familiar`Crimbo Shrub`:
      return Math.ceil(estimatedTurns() / 100);

    default:
      return 0;
  }
}
```
which can live in this file at the bottom.

I don't want the types of limit to grow every time we get a weirdo, so instead we'll just put more cases in the special familiars.

horrible-little-slime (changes requested):

horrible-little-slime (approved):

PR #1072, opened 2022-09-13
Count marginal familiar turns available better

horrible-little-slime:


src/familiar/lib.ts
```diff
@@ -79,3 +82,12 @@ export function canOpenRedPresent(): boolean {
     myInebriety() <= inebrietyLimit()
   );
 }
+
+export function turnsAvailable(): number {
+  const baseTurns = estimatedTurns();
+  const digitizes = digitizedMonstersRemaining();
+  const mapTurns = globalOptions.ascending
+    ? clamp(availableAmount($item`Map to Safety Shelter Grimace Prime`), 0, 60)
```
docrostov, line 90: so i think clamping mapTurns to 60 is totally fine, i know garbo clamps actually generating the maps to 60 [here](https://github.com/Loathing-Associates-Scripting-Society/garbage-collector/blob/dc4a38ceeb96166bd6122c230b06c27bce341098/src/fights.ts#L2152) but do we maybe want to tie this & the max number of maps garbo grabs to one number so we don't have to change the code in multiple places in the event we someday get a better nightcap? maybe like "export const expectedOverdrunkTurns = 60;" above this and clamp both this & the value in fights.ts to that constant?

docrostov (changes requested):
it looks fine and the logic seems reasonable. i did not run it but it looks OK. i think it would be good to add a docstring for turnsAvailable. this is my recommended docstring:

```
/**
 * Number of turns remaining in the barf day, subtracting digitized embezzlers & overdrunk grimacia maps. Used for marginal familair turn tabulations.
 * @returns An integer representing the turns Garbo expects the player to play, subtracting out digitized embezzlers & Grimacia Maps.
 */
```

src/familiar/lib.ts
horrible-little-slime: Yeah I'll name this constant instead of using a magic number

docrostov (approved):

PR #1074, opened 2022-09-14
Add all wormwood locs to skiplist

Tokoeka:
I have had multiple instances of garbo throwing due to trying to place a wanderer in one of the worm wood locations for a Platinum Guzzlr quest, but hitting a worm wood superlikely (with no combat option) instead. 

Hence this proposal to add them all to the skiplist.


horrible-little-slime:
If we're going to do this, we should remove absinthe from our unlockablezone things

horrible-little-slime:
This has a big enough merge conflict post-revamp that I'm going to close it; we can revisit this, but if we do, we should remove the code to unlock the wormwood as well

PR #1084, opened 2022-09-16
Only use nonviolent item familiars

horrible-little-slime:


src/familiar/itemFamiliar.ts
```diff
@@ -12,7 +12,12 @@ export function bestFairy(): Familiar {
   if (!bestNonCheerleaderFairy) {
     const viableFairies = Familiar.all()
       .filter(
-        (f) => have(f) && findFairyMultiplier(f) && f !== $familiar`Steam-Powered Cheerleader`
+        (f) =>
+          have(f) &&
+          findFairyMultiplier(f) &&
```
docrostov, line 17: i took a brief look at libram and i expect this works properly but you know how i am in life and must have known i would ask this -- what does findFairyMultiplier do regarding special-case drop fams, like the peppermint rhino that has item drop & extra +candy drop?

src/familiar/itemFamiliar.ts
```diff
@@ -12,7 +12,12 @@ export function bestFairy(): Familiar {
...
+          f !== $familiar`Steam-Powered Cheerleader` &&
+          !f.physicalDamage &&
```
docrostov, line 19: just so you know i went through and did a runthrough of the familiars on the [wiki](https://kol.coldfront.net/thekolwiki/index.php/Familiars_by_type#Increases_item_drop) for item & damage and they all properly were tagged with their correct damage source, so this looks good.

docrostov (approved):
my only minor gripe with this is that i could see situations where it is technically better to use an attacking item fam (because it's also a marginal fam or there are other benefits to gain from it) over a non-attacking one. but overall the logic is sound and it is a reasonable thing to just say "this is how garbo will operate sorry folks"

PR #1086, opened 2022-09-16
Stone wool

libraryaddict:


horrible-little-slime:
Looks good now! is this tested?

libraryaddict:
Just tried to test, but mafia updated and wrote over the script. Guess I'll try remember in my next leg.

libraryaddict:
I can now confirm that it was tested and worked fine. Didn't test what happens if you can't map the monsters, or stone wool isn't the most $$

PR #1087, opened 2022-09-16
Throw immediately if we're too drunk and can't handle it

horrible-little-slime:


docrostov (approved):
looks good to me. good addition. thank you, slifty.

PR #1088, opened 2022-09-17
Pickpocket grimacia

Shiverwarp:
Equips pickpocket enabling items and attempts to pickpocket items before using Force Friends on Grimacia Map targets.

Survivors never win initiative, so can always pickpocket, and force friends always give the items even if you've pickpocketed first.

We are free running/banishing here so our other equipment isn't relevant except for pickpocketing

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -3,6 +3,7 @@ import {
   availableAmount,
   buy,
   canAdventure,
+  canEquip,
   cliExecute,
   closetAmount,
   create,
@@ -32,6 +33,7 @@ import {
   myLevel,
   myMaxhp,
   myPathId,
+  myPrimestat,
   myThrall,
   numericModifier,
   outfit,
@@ -71,6 +73,7 @@ import {
   $phylum,
   $skill,
   $slot,
+  $stat,
   $thrall,
   ActionSource,
   adventureMacro,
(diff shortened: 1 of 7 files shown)


src/fights.ts
```diff
@@ -872,14 +872,19 @@ const freeFightSources = [
               Macro.trySkill($skill`Feel Hatred`).tryItem($item`divine champagne popper`)
             )
             .if_($monsters`giant rubber spider, time-spinner prank`, Macro.kill())
+            .step("pickpocket")
             .skill($skill`Use the Force`)
+            .setAutoAttack();
```
horrible-little-slime, line 877: Macro.setAutoAttack() doesn't return a `Macro` object; if you want to make this fight use autoattack, you'll need to change the function to be `adventureMacroAuto`

But there's an issue with that! Using `use the force` during an autoattack can cause weird misalignment issues.

src/fights.ts
```diff
@@ -872,14 +872,19 @@ const freeFightSources = [
...
         );
       }
     },
     false,
     {
       requirements: () => [
-        new Requirement([], { forceEquip: $items`Fourth of May Cosplay Saber` }),
+        new Requirement(["1000 Pickpocket Chance"], {
+            forceEquip: $items`Fourth of May Cosplay Saber`,
+            bonusEquip: new Map<Item, number>([[$item`tiny black hole`, 5000],[$item`mime army infiltration glove`, 5000]]),
```
horrible-little-slime, line 886: We only want to use these items if we aren't a moxie class; in addition, we only want to equip one of them, not both. Rather than adding this as a bonus equip, I'd add logic to pick one of them as a forceEquip if you have it.

src/fights.ts
horrible-little-slime: You might be able to do something interesting with the three-input version of `adventureMacroAuto`

horrible-little-slime (changes requested):
This is a great first step. To get linting checks to work, you'll need to install eslint and prettier, and run a linter on this. It'll change some of the formatting to look a little nicer, for some definition of nicer. I also have a few for-sure-surmountable concerns with the approach you're taking.

src/fights.ts
Shiverwarp: Most of the time both will be equipped anyway because they give bonus pickpocket chance that's either better or equivalent to other possibilities.

It's possible that this might not be the case in the future though so I'll try to figure out some logic that makes sense

src/fights.ts
Shiverwarp: I was going to do the autoattack because there was a bug previously that made these pickpocket granting items not work with balls macros, but it seems that's been fixed, so I can just remove the .setAutoAttack() step

horrible-little-slime (approved):

PR #1094, opened 2022-09-18
Add `quick` arg to sacrifice optimal behaviors for quicker execution

Rinn:
Only change currently is to instead use historical price rather than getSalePrice if the prices are less than a week old, for only items that will be sold.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -155,6 +155,7 @@ function printSession(session: Session): void {
   printProfit(highValue);
   print(` You lost meat on ${lowValue.length} items including:`);
   printProfit(lowValue);
+  print(`Quick mode was enabled, results may be less accurate than normal.`);
 }
 
 function garboSaleValue(item: Item) {
--- a/src/session.ts
+++ b/src/session.ts
@@ -1,4 +1,5 @@
 import {
+  autosellPrice,
   Coinmaster,
   historicalAge,
   historicalPrice,
--- a/src/session.ts
+++ b/src/session.ts
@@ -161,7 +161,8 @@ function printSession(session: Session): void {
 function garboSaleValue(item: Item) {
   if (globalOptions.quickMode) {
     if (historicalAge(item) <= 7.0 && historicalPrice(item) > 0) {
-      return historicalPrice(item) * 0.9;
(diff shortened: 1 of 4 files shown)


horrible-little-slime (approved):
Is this tested? It looks generally good

Rinn:
I'll test this after rollover, i'm in the middle of a run atm

Rinn:
I'll follow up after it finishes running, but it did blitz through all the dailies super fast, most notably not searching for anything when it open cargo shorts. The only thing I think may cause issue is the 50k maximize limit and the longer maximize query because of all the -equip brimstone, but i'll know more later.

horrible-little-slime:
Can/should we closet brimstone instead? I guess if we do that we should do
papier mache as well

Rinn:
We don't currently disable acquiring from closet so that wouldn't make a difference. My wrapper script put them in my display but that's really dumb, also if I remember correctly i think fixed the issue with long maximize strings in mafia a while ago.

Rinn:
I had some issues due to a power outage so I'll test this again after another run in a few days.

docrostov:
@phulin gurbo is lit

src/session.ts
```diff
@@ -147,14 +155,24 @@ function printSession(session: Session): void {
   printProfit(highValue);
   print(` You lost meat on ${lowValue.length} items including:`);
   printProfit(lowValue);
+  print(`Quick mode was enabled, results may be less accurate than normal.`);
+}
+
+function garboSaleValue(item: Item) {
+  if (globalOptions.quickMode) {
+    if (historicalAge(item) <= 7.0 && historicalPrice(item) > 0) {
+      return historicalPrice(item) * 0.9;
```
horrible-little-slime, line 164:
```suggestion
      const isMallMin = historicalPrice(item) === Math.min(100, 2 * autosellPrice(item));
      return isMallMin ? autosellPrice(item) : 0.9 * historicalPrice(item);
```

horrible-little-slime (changes requested):
One tiny change

horrible-little-slime (approved):
yell heah

PR #1103, opened 2022-09-20
Add supprt for YR for best item

pstalcup:
###


src/wanderer.ts
```diff
@@ -11,12 +23,13 @@ import {
...
   have,
   SourceTerminal,
+  sumNumbers,
 } from "libram";
 import { estimatedTurns } from "./embezzler";
 import { globalOptions, HIGHLIGHT, propertyManager, realmAvailable } from "./lib";
 import { garboValue } from "./session";
 
-export type DraggableFight = "backup" | "wanderer";
+export type DraggableFight = "backup" | "wanderer" | "yr";
```
horrible-little-slime, line 32: Might want to make this an enum at this point; I think we should also go ahead and call this `"yellow ray"` as well.

src/wanderer.ts
```diff
@@ -146,7 +159,7 @@ function freeCrafts() {
 }
 class WandererTarget {
   name: string;
-  available: () => boolean;
+  available: (type: DraggableFight) => boolean;
```
horrible-little-slime, line 162: This is making me start to wonder if going through the extant wanderer setup is worth it, given how different from everything else this is. Can we use the skipList and related functions without passing through this framework?

Actually this is helpful, comparing YR value to lovebug coinspiracy value is useful. Bleh.

src/wanderer.ts
```diff
@@ -206,6 +219,40 @@ function guzzlrAbandonQuest() {
...
+            .filter((drop) => drop.type !== "c")
+            .map((drop) => garboValue(drop.drop))
+        )
+      )
+    );
+  }
+}
+
+function bestYrLocation(): { location: Location; value: number } {
+  const validLocations = $locations``
```
horrible-little-slime, line 243: I personally have a strong preference for `MafiaClass.all()` instead of using the empty template string thing. But I am a weirdo.

src/wanderer.ts
```diff
@@ -206,6 +219,40 @@ function guzzlrAbandonQuest() {
...
+    .filter(canAdventureOrUnlock)
```
horrible-little-slime, line 244: So this actually does not feel ideal given the plan here; this is a type of wanderer that is inherently extremely spread out, so we probably don't want to spend meat on, say, a transponder just to YR there. I am open for debate on this point, but actually we should really start to look at the price of these things when we're doing guzzlr as well.

horrible-little-slime: so really it's the "or unlock" that gets me

Also, I think we might as well do the combatPercent filtering here, in addition to happening later. We don't want to do math on a really big set of values if we can get away with it.

src/wanderer.ts
```diff
@@ -206,6 +219,40 @@ function guzzlrAbandonQuest() {
...
+    .map((location) => {
+      return { location, value: averageYrValue(location) };
+    })
+    .filter((value) => value.value > 0);
+  if (validLocations.length > 0) {
+    return validLocations[0];
```
horrible-little-slime, line 250: This array is not sorted. We want to either sort (messy) or reduce/argmax

src/wanderer.ts
horrible-little-slime: Backups are really the main one that's worth blowing a transponder on, probably. Actually, are transponders expensive? 

8k. At current guzzlr prices that is almost certainly not worth it.

Oh god I just realized what we should do--it's only plats that require these, right? we have it pay attention to unlock price if the platinum isn't capped yet

That is out of scope. But I do think this should probably use a straight canAdv

src/wanderer.ts
```diff
@@ -276,6 +323,19 @@ const wandererTargets = [
...
+  new WandererTarget(
+    "Yellow Ray",
+    (type: DraggableFight) => type === "yr",
+    () => {
+      const bestLocation = bestYrLocation();
+      if (bestLocation.value > 0) {
+        return bestLocation.location;
+      } else {
+        return null;
+      }
```
horrible-little-slime, line 335: I feel like this should be ternary iff a linter doesn't split it up over multiple lines.

horrible-little-slime (changes requested):

src/wanderer.ts
```diff
@@ -206,6 +219,40 @@ function guzzlrAbandonQuest() {
...
 
+function averageYrValue(location: Location) {
+  const monsters = Object.keys(getLocationMonsters(location))
+    .map((m) => toMonster(m))
+    .filter((m) => !m.attributes.includes("LUCKY") && !m.attributes.includes("ULTRARARE"));
+
+  if (monsters.length === 0) {
+    return 0;
+  } else {
+    return sumNumbers(
```
horrible-little-slime, line 230: You need to divide by monsters.length

src/barfTurn.ts
```diff
@@ -296,7 +296,7 @@ const turns: AdventureAction[] = [
       const locationChoice =
         familiarChoice === $familiar`Space Jellyfish`
           ? $location`Pirates of the Garbage Barges`
-          : determineDraggableZoneAndEnsureAccess("backup");
+          : determineDraggableZoneAndEnsureAccess("yellow ray");
```
horrible-little-slime, line 299: let's go ahead and nix the jellyfish stuff here when we merge this; a 20th of a stench jelly does not beat a well-used YR

horrible-little-slime: Also that'll make it harder to test

src/wanderer.ts
```diff
@@ -206,6 +219,52 @@ function guzzlrAbandonQuest() {
   }
 }
 
+export function averageYrValue(location: Location, debug = false) {
```
horrible-little-slime, line 222: You no longer need to export these

src/wanderer.ts
```diff
@@ -206,6 +219,52 @@ function guzzlrAbandonQuest() {
...
+export function bestYrLocation(): { location: Location; value: number } {
+  const validLocations = $locations``
+    .filter(canAdventureOrUnlock)
+    .filter((l) => canWander(l, "yellow ray"))
+    .map((location) => {
+      return { location, value: averageYrValue(location) };
+    })
+    .filter((value) => value.value > 0);
+  if (validLocations.length > 0) {
+    const res = validLocations.sort((a, b) => b.value - a.value)[3];
```
horrible-little-slime, line 260: let's use a reduce instead of a sort.

Why are we looking at the fourth best? Is that because we haven't hacked the bad zones yet?

src/wanderer.ts
```diff
@@ -206,6 +219,52 @@ function guzzlrAbandonQuest() {
...
+    print(`Best YR Location: ${res.location} for value ${res.value}`);
```
horrible-little-slime, line 261: Honestly I think we should keep this print in even later

horrible-little-slime (dismissed):

src/wanderer.ts
```diff
@@ -206,6 +222,52 @@ function guzzlrAbandonQuest() {
...
+        })
+      ) / monsters.length
+    );
+  }
+}
+
+export function bestYrLocation(): { location: Location; value: number } {
+  const validLocations = $locations``
+    .filter(canAdventureOrUnlock)
+    .filter((l) => canWander(l, "yellow ray"))
```
horrible-little-slime, line 257: We should merge these two filters

src/wanderer.ts
horrible-little-slime: We can get rid of the [3] now that we've nixed the clan dungeon

src/wanderer.ts
gausie: This is not weird, you are right

gausie: We should disallow this in libram, it's horrible!

PR #1104, opened 2022-09-21
Support mush-mouth

pstalcup:


src/post.ts
```diff
@@ -155,6 +159,28 @@ function stillsuit() {
   }
 }
 
+function funguySpores() {
+  if (
+    myLevel() >= 15 &&
+    !have($effect`Mush-Mouth`) &&
+    (!globalOptions.ascending || myAdventures() > 11) &&
+    get("dinseyRollercoasterNext")
```
horrible-little-slime, line 167: Add a comment about why you're doing this

src/post.ts
```diff
@@ -155,6 +159,28 @@ function stillsuit() {
   }
 }
 
+function funguySpores() {
+  if (
+    myLevel() >= 15 &&
```
horrible-little-slime, line 164: possibly add a comment about why this; we should fiddle with this threshold a bit

horrible-little-slime (approved):

horrible-little-slime:
I'm going to consult the Doc on this, but otherwise it's good. The main thing that's spicy is our stat threshold choice.

PR #1109, opened 2022-09-24
add option to restrict bestConsumable by levelreq

Tokoeka:
Noticed garbo get gingerbread cigarettes over high end ginger wine. Apparently next best consumable was dreadsylvanian grimlet, despite not being of level to drink one of those. Hence, adding in option to restrict BestCOnsumable to items you can currently drink based on level requirement


horrible-little-slime:
Looks good other than linting

PR #1110, opened 2022-09-24
occasionally go into a stupor on purpose

horrible-little-slime:


src/barfTurn.ts
```diff
@@ -64,7 +64,15 @@ import {
...
 } from "./outfit";
 import postCombatActions from "./post";
-import { determineDraggableZoneAndEnsureAccess, digitizedMonstersRemaining } from "./wanderer";
+import {
+  determineDraggableZoneAndEnsureAccess,
+  digitizedMonstersRemaining,
+  DraggableFight,
+} from "./wanderer";
+
+const sober = () => myInebriety() <= inebrietyLimit();
```
s-k-z, line 73: Can this run while Stooper is active?

horrible-little-slime: I believe that we unequip the stooper when we run diet code, and that we'll always run diet code by the time we go here. but that is worth special casing regardless.

src/barfTurn.ts
```diff
@@ -64,7 +65,16 @@ import {
...
 import postCombatActions from "./post";
-import { determineDraggableZoneAndEnsureAccess, digitizedMonstersRemaining } from "./wanderer";
+import {
+  determineDraggableZoneAndEnsureAccess,
+  digitizedMonstersRemaining,
+  DraggableFight,
+} from "./wanderer";
+
+const sober = () =>
+  myInebriety() <= inebrietyLimit() + (meatFamiliar() === $familiar`Stooper` ? -1 : 0);
```
s-k-z, line 75: should be `myFamiliar()` not `meatFamiliar()` yeah?

PR #1111, opened 2022-09-25
Don't use shrub against bricko monsters

horrible-little-slime:


src/fights.ts
```diff
@@ -672,7 +688,7 @@ const freeFightSources = [
     },
     false,
     {
-      canOverrideMacro: true,
+      macroAllowsFamiliarActions: true,
```
Pantocyclus, line 691: We don't really need this if it already defaults to true
(Not a major issue, just a thing about standarization - the other FreeFights don't have this)

Pantocyclus (approved):
Seems fine (apart from that one minor cosmetic issue)

PR #1112, opened 2022-09-25
Various yachtzeechain optimisations

Pantocyclus:
(1) Delay freeRunFights() to post-yachtzeechain to save on LTBs 
(2) Consider Pocket Wishes as a Fishy source 
(3) Support liver cleansers within the yachtzeechain diet

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -1877,8 +1877,9 @@ const freeKillSources = [
   ),
 ];
 
-export function freeFights(): void {
+export function freeRunFights(): void {
   if (myInebriety() > inebrietyLimit()) return;
+  if (globalOptions.yachtzeeChain && !get("_garboYachtzeeChainCompleted", false)) return;
   if (
     get("beGregariousFightsLeft") > 0 &&
     get("beGregariousMonster") === $monster`Knob Goblin Embezzler`
@@ -1903,6 +1904,24 @@ export function freeFights(): void {
       freeRunFightSource.runAll();
     }
   });
+}
+
+export function freeFights(): void {
+  if (myInebriety() > inebrietyLimit()) return;
+  if (
+    get("beGregariousFightsLeft") > 0 &&
+    get("beGregariousMonster") === $monster`Knob Goblin Embezzler`
+  ) {
+    return;
+  }
+  visitUrl("place.php?whichplace=town_wrong");
+
+  propertyManager.setChoices({
+    1387: 2, // "You will go find two friends and meet me here."
+    1324: 5, // Fight a random partier
+  });
+
+  freeRunFights();
 
   killRobortCreaturesForFree();
 
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -1886,7 +1886,6 @@ export function freeRunFights(): void {
   ) {
     return;
   }
-  visitUrl("place.php?whichplace=town_wrong");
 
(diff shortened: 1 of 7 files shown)


src/yachtzee/fishy.ts
```diff
@@ -162,6 +162,37 @@ export function optimizeForFishy(yachtzeeTurns: number, setup?: boolean): number
...
+        }
+        cliExecute(`genie effect Fishy`);
+        if (haveFishyPipe && haveEffect($effect`Fishy`) + adventureExtensionBonus < yachtzeeTurns) {
+          use(1, $item`fishy pipe`);
+        }
+      },
+    },
+    {
+      name: "2x Pocket Wish",
+      turns: 40 + (haveFishyPipe ? 10 : 0),
```
horrible-little-slime, line 182: We should probably stop adding this separately to every fishy source (out of scope of this PR)

src/fights.ts
```diff
@@ -1877,8 +1877,9 @@ const freeKillSources = [
   ),
 ];
 
-export function freeFights(): void {
+export function freeRunFights(): void {
```
horrible-little-slime, line 1880: we run FreeRunFights before regular FreeFights and FreeKillFights. Are you worried about unintended consequences of changing this sequencing?

For example, when we `killRobortCreatures`, we use cartography charges to get penguin access. But that means we can't use cartography when doing extinguisher/xo steals. There are a few other similar things like that. FreeRunFights also benefit from having high buffed +famweight, and after yachtzee we'll lose some of those buffs. And this means we're less likely to have our designer sweatpants charged for a nice day of yachtzeeing (although not _that_ much less likely, given that regular free fights exist)

src/fights.ts
```diff
@@ -1903,6 +1904,24 @@ export function freeFights(): void {
...
+
+export function freeFights(): void {
+  if (myInebriety() > inebrietyLimit()) return;
+  if (
+    get("beGregariousFightsLeft") > 0 &&
+    get("beGregariousMonster") === $monster`Knob Goblin Embezzler`
+  ) {
+    return;
+  }
+  visitUrl("place.php?whichplace=town_wrong");
```
horrible-little-slime, line 1917: I think we can remove this now that coolitems exists

src/fights.ts
Pantocyclus: I'm really not too concerned about banderways because it always seems like we either have too many (and garbo does nothing with those extra free runs) or too little (and garbo starts using LTBs for spikolodons). 

In the case where we have too little, we do want to use them over LTBs (although I suppose we should then consider pushing spikolodon forward in the diet schedule).

In the case where we have too many, losing a few banderways due to lost +lbs in most cases literally does not make a difference.

Also, I'd imagine fish heads/penguins are much more valuable than our other uses of cartography (I could be very wrong here), but the old behaviour was to prioritise freeRunFight cartography uses over penguins - if so, wouldn't this new behaviour be better?

horrible-little-slime:
Per discussion in discord, we should move the nonorgan adventures and resequence the diet steps in response to the freerunfight deprioritization

horrible-little-slime:
If we're moving all freekills after yachtzee we get the kind of funny result of not using wolf form for yachtzee. Which is not the end of the world but is sad.

PR #1116, opened 2022-09-26
Attempt at properly valuing smithsness for purse

Shiverwarp:
This is very much a work in progress and has all sorts of problems, but maybe it's a starting point for a more elegant fix to the current problems with Smithsness? I'm probably asking more questions than I'm answering with this PR

So the problems: 
1. This is only considering the value when doing embezzlers. During barf, Half a Purse will be less powerful, but we'd still want to use up our turns of smithsness. Maybe it's out of scope for what I'm attempting to do here to split up the consideration between embezzlers and barf, but there it is anyway. I could do the opposite and just check vs barf turns, and then embezzlers will guaranteed also be better with Purse, as far as I know. But the question is still up in the air which one would be worse to miss out on. Waste turns of smithsness in barf without equipping purse? Or don't use smithsness when it would be more valuable on embezzlers?

2. This equips an outfit just to check whether something is valuable, so it's slow, and also might cause some weirdness with changing HP/MP values that cause recovery checks. But because of the caching already in the function, it will only run once at least, so it's not too bad.

3. It's ugly and truthfully I'm foggy on how some of these things work (meatOutfit for example, as well as the map for the bonus equip, which I basically just copied from the mayfly turns. Also my first time using '() =>' and have no idea if it was even necessary.

4. This will still run into problems until the maximizer bug is fixed

5. It's possible when fully buffed up with effects the Half a Purse could still outperform whatever other offhand we're using instead with familiar weight because of the diminishing returns, but I'm unsure of how to account for this because I'm not sure when during the mood management this would be happening. I moved it to the end of mood.ts in a naïve attempt to account for this

6. And whatever else I've almost certainly missed


Shiverwarp:
Currently breaks with
```
JavaScript error: Cannot read property "M" from undefined
at file:/C:/Shiverwarp/Utilities/KoLMafia/scripts/garbage-collector/garbo-combat.js:17862
```

assuming because of my meatFamiliar() usage, but don't really understand it yet.

In any case I think I've made a few wrong assumptions about how and when this would be run (would we even have our latte ingredients yet?)

horrible-little-slime:
If I had to guess this creates some sort of bizarre circular import, although I'm not sure where it would be. `usingPurse` is called only in `mood` and `potions`, and it's odd that the issue is showing up in garbo_combat.js

Shiverwarp:
So I think the way to actually do this would be to use a maximize simulation call, and then check whether Generated:_spec contains the purse.

https://wiki.kolmafia.us/index.php/Maximize

However I can't seem to get it to work. It might be that this no longer a feature?

src/mood.ts
```diff
@@ -141,6 +138,9 @@ export function meatMood(urKels = false, meat = baseMeat): Mood {
     }
   }
 
+  const flaskValue = usingPurse() ? 0.3 * baseMeat : 5;
```
horrible-little-slime, line 141: Is there a reason you moved these lines? It's NBD but I'm curious

src/outfit.ts
```diff
@@ -348,11 +352,28 @@ export const familiarWaterBreathingEquipment = $items`das boot, little bitty bat
...
-        !canAdventure($location`The Black Forest`));
+    if (myInebriety() <= inebrietyLimit()) {
+      const purseBonus = () => {
+        if (haveEffect($effect`Merry Smithsness`)) {
+          return 0;
+        } else if (SongBoom.have()) {
+          return 1025 * 0.6 - mallPrice($item`Flaskfull of Hollow`) / 150;
+        } else {
+          return 1000 * 0.6 - mallPrice($item`Flaskfull of Hollow`) / 150;
+        }
```
horrible-little-slime, line 363: traditionally we use `750 + baseMeat` to solve this issue

horrible-little-slime (changes requested):
This looks pretty good, and you're right that it's cached so this is fine. I can imagine that this has some weird quibbles, but frankly they're probably still better than current behavior.

src/mood.ts
Shiverwarp: Honestly I haven't looked at this in a long time because I couldn't figure out why it was having the circular import issue.

This was one of the first attempts I had at a PR and I was even more ignorant than I am now so I feel like if I wanted to try to take a stab at this again I'd just do it all over

I'm guessing from my initial PR post I moved them because of this:

> It's possible when fully buffed up with effects the Half a Purse could still outperform whatever other offhand we're using instead with familiar weight because of the diminishing returns, but I'm unsure of how to account for this because I'm not sure when during the mood management this would be happening. I moved it to the end of mood.ts in a naïve attempt to account for this

PR #1117, opened 2022-09-26
Refactor wanderers

pstalcup:


src/wanderer/guzzlr.ts
```diff
@@ -0,0 +1,113 @@
...
+import { canAdventureOrUnlock, canWander, WandererTarget } from "./lib";
+
+function freeCrafts() {
+  return (
+    (have($skill`Rapid Prototyping`) ? 5 - get("_rapidPrototypingUsed") : 0) +
+    (have($skill`Expert Corner-Cutter`) ? 5 - get("_expertCornerCutterUsed") : 0)
+  );
+}
+
+function wandererTurnsAvailableToday(zone: Location) {
```
horrible-little-slime, line 15: This feels like it should be in `wanderer/lib.ts`

src/wanderer/lib.ts
```diff
@@ -0,0 +1,246 @@
...
+    value: number,
+    prepareTurn: () => boolean = () => true
+  ) {
+    this.name = name;
+    this.value = value;
+    this.location = location;
+    this.prepareTurn = prepareTurn;
+  }
+}
+export type WandererFactory = (type: DraggableFight) => WandererTarget[] | undefined;
```
horrible-little-slime, line 136: `null` is more purposeful than `undefined`, but is there a compelling reason to not just return `[]`?

src/wanderer/yellowray.ts
```diff
@@ -0,0 +1,74 @@
...
+          return sumNumbers(items);
+        })
+      ) / monsters.length
+    );
+  }
+}
+
+function yrValues(): Map<Location, number> {
+  const values = new Map<Location, number>();
+  for (const location of Location.all()) {
```
horrible-little-slime, line 39: should we filter down `Location.all()` at all?

I have a change in the pipeline that will let us use historical values for this, which would help with the performance here

src/wanderer/yellowray.ts
```diff
@@ -0,0 +1,74 @@
...
+            .map((drop) => garboValue(drop.drop));
+
+          if (debug) print(`${m}: ${items.join(",")}`);
+
+          return sumNumbers(items);
+        })
+      ) / monsters.length
+    );
+  }
+}
```
horrible-little-slime, line 35: I am probably merging https://github.com/Loathing-Associates-Scripting-Society/garbage-collector/pull/1113 before this, so just watch out for the changes

src/wanderer/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  } else {
+    propertyManager.setChoices(unsupportedChoices.get(candidate.location) ?? {});
+    const targets = candidate.targets.map((t) => t.name).join(",");
+    print(`Wandering at ${candidate.location} for expected value ${candidate.value} (${targets}`);
+
+    return candidate.location;
+  }
+}
+
+export function drunkWander(type: DraggableFight) {
```
horrible-little-slime, line 82: So what's kind of funny is that this is not really quite a correct name for it, because this is meant to be called when sober or drunk. But I don't have a great suggestion off the top of the dome.

src/barfTurn.ts
```diff
@@ -237,7 +227,7 @@ const turns: AdventureAction[] = [
         })
       );
       adventureMacroAuto(
-        isGhost ? noWineglassZone() : determineDraggableZoneAndEnsureAccess(),
+        isGhost ? drunkSafeWander("wanderer") : wanderWhere("wanderer"),
```
horrible-little-slime, line 230: I'm still not 100% sure the meaning of `drunkSafeWander` is clear, and given that it is only used in barfturn, it might make more sense to define it in `barfTurn` instead

src/turns.ts
```diff
@@ -0,0 +1,127 @@
...
+import { $familiar, $item, Counter, get, have, SourceTerminal } from "libram";
+import { usingThumbRing } from "./dropsgear";
+import { embezzlerCount } from "./embezzler";
+import { ESTIMATED_OVERDRUNK_TURNS, globalOptions } from "./lib";
+
+/**
+ * Computes the estimated number of turns during which garbo will run
+ * @returns A guess of how many runs garbo will run in total
+ */
+export function estimatedTurns(): number {
```
horrible-little-slime, line 19: We really gotta spruce this up at some point

src/wanderer/guzzlr.ts
```diff
@@ -0,0 +1,101 @@
...
+    (globalOptions.ascending && wandererTurnsAvailableToday(location) < remaningTurns) // or ascending and not enough turns to finish
+  ) {
+    print("Abandoning...");
+    Guzzlr.abandon();
+  }
+}
+
+function acceptGuzzlrQuest(locationSkiplist: Location[]) {
+  if (Guzzlr.isQuestActive()) considerAbandon(locationSkiplist);
+  while (!Guzzlr.isQuestActive()) {
```
horrible-little-slime, line 36: This `while` will always terminate because you can always accept a bronze quest, right?

src/wanderer/lib.ts
```diff
@@ -0,0 +1,248 @@
...
+
+export function underwater(location: Location): boolean {
+  return location.environment === "underwater";
+}
+
+export function canAdventureOrUnlock(loc: Location): boolean {
+  const skiplist = [
+    ...$locations`The Oasis, The Bubblin' Caldera, Barrrney's Barrr, The F'c'le, The Poop Deck, Belowdecks, 8-Bit Realm, Madness Bakery, The Secret Government Laboratory, The Dire Warren`,
+    ...Location.all().filter((l) => l.parent === "Clan Basement"),
+  ];
```
horrible-little-slime, line 69: Is there any meaningful performance value in defining this (and the other skipLists) outside of the functions, or is that frivolous of me?

src/wanderer/lib.ts
```diff
@@ -0,0 +1,248 @@
...
+  if (location === $location`The Fun-Guy Mansion` && get("funGuyMansionKills", 0) >= 100) {
+    return false;
+  }
+  return canWanderTypeBackup(location);
+}
+
+const wandererSkiplist = $locations`The Batrat and Ratbat Burrow, Guano Junction, The Beanbat Chamber, A-Boo Peak`;
+function canWanderTypeWander(location: Location): boolean {
+  return !wandererSkiplist.includes(location) && location.wanderers;
+}
```
horrible-little-slime, line 100: It feels a _little_ weird to define these in `lib`, but I see why you did it.

src/wanderer/lib.ts
```diff
@@ -0,0 +1,248 @@
...
+  [$location`The Copperhead Club`, { [855]: 4 }],
+  [$location`The Castle in the Clouds in the Sky (Top Floor)`, { [1431]: 1, [677]: 2 }],
+  [$location`The Hidden Office Building`, { [786]: 6 }],
+]);
+
+export function defaultFactory(): WandererTarget[] {
+  return [new WandererTarget("Default", $location`The Haunted Kitchen`, 0)];
+}
+
+export function maxBy<T>(array: T[], key: (t: T) => number): T {
```
horrible-little-slime, line 186: This is worth moving to our regular lib or to libram utils; we might already have some sort of `argMax` function defined somewhere. But yeah we've got to use this more!

src/wanderer/lib.ts
```diff
@@ -0,0 +1,248 @@
...
+  return array
+    .map((t: T) => {
+      return { t, value: key(t) };
+    })
+    .reduce((prev, curr) => (prev.value < curr.value ? curr : prev)).t;
```
horrible-little-slime, line 191: something in my head wants me to compress this into one reduce, but I can't quite make it fit together perfectly like that without either double-calling your `key` function a lot or having a default value for empty arrays.

src/wanderer/lib.ts
```diff
@@ -0,0 +1,248 @@
...
+}
+
+type WanderingSource = {
+  name: string;
+  item: Item;
+  max: number;
+  property: NumericProperty;
+  type: DraggableFight;
+};
+const WanderingSources: WanderingSource[] = [
```
horrible-little-slime, line 201: This information may or may not be in libram's wanderers code. I don't think it has been kept up to date. We should maybe port some of this over.

src/wanderer/lib.ts
```diff
@@ -0,0 +1,248 @@
...
+  },
+  {
+    name: "Voter",
+    item: $item`"I Voted!" sticker`,
+    max: 3,
+    property: "_voteFreeFights",
+    type: "wanderer",
+  },
+  {
+    name: "Backup",
```
horrible-little-slime, line 224: We have a silght issue that we need to figure out:
we have a type of `wanderer` called a `wanderer`. That leaves me confused--I see this "backup" thing, and I immediately thing "but I thought that was a wanderer".

I don't have a great solution to this! but we should rename one of them. I think that is how we ended up with the admittedly inelegant 'DraggableFight' name

src/wanderer/lib.ts
```diff
@@ -0,0 +1,248 @@
...
+    item: $item`backup camera`,
+    max: 11,
+    property: "_backUpUses",
+    type: "backup",
+  },
+];
+
+export function wandererTurnsAvailableToday(location: Location): number {
+  const canWanderCache: { [K in DraggableFight]: boolean } = {
```
horrible-little-slime, line 233: can/should this just be a `Record<DraggableFight, boolean>`? I find that works nicely for enums

src/wanderer/yellowray.ts
```diff
@@ -0,0 +1,71 @@
...
+    .map((m) => toMonster(m))
+    .filter((m) => !badAttributes.some((s) => m.attributes.includes(s) && rates[m.name] > 0));
+
+  if (monsters.length === 0) {
+    return 0;
+  } else {
+    return (
+      sum(monsters, (m) => {
+        const items = itemDropsArray(m).filter((drop) => ["", "n"].includes(drop.type));
+        return sum(items, (drop) => garboValue(drop.drop));
```
horrible-little-slime, line 27:
```suggestion
        return sum(items, (drop) => garboValue(drop.drop, true));
```

to make it use historical values

src/wanderer/yellowray.ts
horrible-little-slime: I am now more confident that we should at least slightly filter `Location.all()` by removing underwater zones, zones from the SkipList, and zones with a combat rate of less than... I guess `70 - getModifier("Combat Rate")`? 70 coming from 100 - the soft cap of 30. 

`averageYrValue` currently doesn't use historical values, so we really want to pare down the number of zones we search to only be ones we can plausibly select. Even with historical values it is probably worth doing, because we'll still trigger mallsearches weeklyish.

src/wanderer/index.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Recursively Check for zones to wander to
+ * @param type type of fight we are looking for
+ * @param nameSkiplist Any wanderer tasks that should be skipped because they could not be prepared
+ * @param locationSkiplist Any locations that should be skipped because they could not be unlocked
+ * @returns A location at which to wander
+ */
+export function wanderWhere(
+  type: DraggableFight,
+  nameSkiplist: string[] = [],
+  locationSkiplist: Location[] = []
```
horrible-little-slime, line 70: I don't currently see the utility in these inputs, but I suppose it can't hurt!

src/wanderer/guzzlr.ts
```diff
@@ -0,0 +1,101 @@
...
+    locationSkiplist.includes(location) ||
+    !canAdventureOrUnlock(location) || // or the zone is marked as "generally cannot adv"
+    (globalOptions.ascending && wandererTurnsAvailableToday(location) < remaningTurns) // or ascending and not enough turns to finish
+  ) {
+    print("Abandoning...");
+    Guzzlr.abandon();
+  }
+}
+
+function acceptGuzzlrQuest(locationSkiplist: Location[]) {
```
horrible-little-slime, line 34: I think eventually guzzlr will be worth so little that we want to basically only do it to cap guzzlr, which will require some logic changes. Will it be difficult to do that? How do we tell it that doing a bronze has value outside of just the bucks?

I think when we're still charging it up we will want to ignore the cost of booze and of zone unlockers, but when we're done we will want to pay attention to them. That can be out of scope of this, though.

horrible-little-slime (commented):
This looks really good; I left a lot of comments, but also you wrote a lot of stuff, so we'll call it a draw.

src/wanderer/index.ts
pstalcup: they mostly facilitate that this function is recursive

src/wanderer/lib.ts
```diff
@@ -61,17 +61,21 @@ export const UnlockableZones: UnlockableZone[] = [
...
+const canAdventureOrUnlockSkipList = [
+  ...$locations`The Oasis, The Bubblin' Caldera, Barrrney's Barrr, The F'c'le, The Poop Deck, Belowdecks, 8-Bit Realm, Madness Bakery, The Secret Government Laboratory, The Dire Warren`,
+  ...Location.all().filter((l) => l.parent === "Clan Basement"),
+];
 export function canAdventureOrUnlock(loc: Location): boolean {
-  const skiplist = [
-    ...$locations`The Oasis, The Bubblin' Caldera, Barrrney's Barrr, The F'c'le, The Poop Deck, Belowdecks, 8-Bit Realm, Madness Bakery, The Secret Government Laboratory, The Dire Warren`,
-    ...Location.all().filter((l) => l.parent === "Clan Basement"),
-  ];
+  const skiplist = [...canAdventureOrUnlockSkipList];
```
horrible-little-slime, line 69: I don't see why we copy the array but c'est la vie

horrible-little-slime (dismissed):
Approved pending garbo not exploding

src/wanderer/yellowray.ts
```diff
@@ -0,0 +1,78 @@
...
+  UnlockableZones,
+  WandererTarget,
+} from "./lib";
+
+function averageYrValue(location: Location) {
+  const badAttributes = ["LUCKY", "ULTRARARE", "BOSS"];
+  const rates = appearanceRates(location);
+  const monsters = Object.keys(getLocationMonsters(location))
+    .map((m) => toMonster(m))
+    .filter((m) => !badAttributes.some((s) => m.attributes.includes(s) && rates[m.name] > 0));
```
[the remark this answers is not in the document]
horrible-little-slime: if I had another leg to test I'd commit this and merge, but I want to absolutely make sure it works beforehand. So probably tonight!

src/wanderer/lib.ts
horrible-little-slime: I understand now, and feel silly for suggesting you yoink them out

PR #1118, opened 2022-09-26
Account for the shitty jellies

horrible-little-slime:
This actually has a big whopping problem (that shouldn't matter right now), and that is that in addition to caring about the location, we deeply care about whether we care about the location. If we plop an embezzler into the deep dark jungle or whatever, this shouldn't matter.


horrible-little-slime:
I've decided I hate this.

gausie:
FWIW hot jelly will sometimes sell to 2crs farmers, sleaze jellies are usually good at crimbo

PR #1125, opened 2022-09-27
Don't recommend Universal Recovery

s-k-z:


Shiverwarp:
As someone who uses UR, what are the problems with it regarding garbo?

I've tried once using mafia's default restores because people recommended it, but it chugged down much more expensive restores compared to what was available and what UR would use.

s-k-z:
It crashes for users with a divide by zero bug that at least two users encountered and you responded to.
It halts on getting beaten up.
It halts on hp hitting exactly zero out of combat.

Shiverwarp:
Are we sure they're using the UR from the proper checkout and not the even more outdated one from the Script Manager in mafia?
I've not had it halt on being beaten up, but perhaps that's a setting I've changed. In fact I have the opposite problem, where it clears beaten up before I can check for it in my scripts

As for the default restoration items, I do want them to be available to use for situations where I'm in ronin, or when they're cheaper. Constantly maintaining that list does not seem like a good answer

s-k-z:
Okay, you can keep using it. What does that have to do with removing the recommendation?

Shiverwarp:
Garbo is about trying to make the most profit, and without UR, we're losing meat on inefficient restorers.

Perhaps instead of removing this recommendation completely, just have a caveat that while it will optimize your restores, it's no longer actively developed, so problems can arise?

s-k-z:
Sample error:
```
 > restore HP

Division by zero (Universal_recovery.ash, line 617)
Could heal 111MP with 5 psychokinetic energy blob for 919.08 meat.
0: fancy blue potion => 695
1: Knob Goblin superseltzer => 830.44
2: palm frond => 832.5
3: palm-frond fan => 835.28
4: Notes from the Elfpocalypse, Chapter II => 837.26
5: phonics down => 841.75
6: cotton candy plug => 845.16
In mallmode, best MP restorative is: fancy blue potion @ 695.0 meat total.
Division by zero (Universal_recovery.ash, line 617)
  at meatper (Universal_recovery.ash:767)
  at daily_handling (Universal_recovery.ash:1774)
  at restore (Universal_recovery.ash:1793)
  at check_restore (Universal_recovery.ash:1859)
  at main (Universal_recovery.ash:1869)
Autorecovery failed.
```

[Here's the Universal Recovery in the scripts manager](https://svn.code.sf.net/p/mafiarecovery/code/scripts/Universal_recovery.ash)

Line 617: 	`if(best_item == null)`

[Here's the Universal Recovery referenced in the README](https://svn.code.sf.net/p/kolmafiascripts/mafiarecovery/code/scripts/Universal_recovery.ash)

Line 617: `set_property("_meatpermp", to_string(item_price(sortit[0])/ heal[sortit[0]].avemp));`

The reported error comes from the one recommended in the README.

Shiverwarp:
Perhaps it was a different divide by zero error fixed by zarqon here?:
https://kolmafia.us/threads/universal-recovery-script.1780/page-120#post-153158

I'm certainly biased towards UR myself because I haven't run into any of these problems personally, and I believe it adds a lot of value in saved meat. But there are a lot of problems with recommending a script that isn't maintained or supported, so I also wouldn't begrudge it being removed from the readme

PR #1128, opened 2022-09-28
Grimoire Daily Tasks

Rinn:


src/tasks/dailyVolcano.ts
```diff
@@ -0,0 +1,116 @@
...
+    acquire: () =>
+      $items`smooth velvet pocket square, smooth velvet socks, smooth velvet hat, smooth velvet shirt, smooth velvet hanky, smooth velvet pants`.map(
+        (x) => <AcquireItem>{ item: x }
+      ),
+    outfit: { modifier: "disco style" },
+  },
+  {
+    name: "Free Volcano Mining",
+    ready: () => realmAvailable("hot") && have($skill`Unaccompanied Miner`),
+    completed: () => get("_unaccompaniedMinerUsed") === 5,
```
Kasekopf, line 104: >= 5, so it doesn't run if preferences are misconfigured

horrible-little-slime:
One mild nitpick I have is that I think the folder should be `/dailies/` rather than `/tasks/`. We will likely eventually grimoirize free fights and embezzler fights, but I still assume that they would be largely separate sections, so not all `task` objects would go in the `/tasks/` folder.

Someone more familiar with grimoire can feel free to swoop in and dissuade me here.

horrible-little-slime:
That works fine then, yeah; I wasn't sure if we'd have to augment the engine in meaningfully different ways for our different sections.

horrible-little-slime (approved):

PR #1131, opened 2022-09-29
Avoid ML when fighting glitch monster

Shiverwarp:


horrible-little-slime:
Is this tested?

Shiverwarp:
> Is this tested?

Not yet, I'll be testing Revers build today.

I can probably test it tomorrow if needed, though I don't have the equipment that causes problems. If nasurte is around I can ask them to test it.

What would be "incorrect" behavior here? Is the reason we're maximizing mainstat just to try to better support people who don't have all the skills needed to kill safely without being hit?

If it equips HOA or Space Trip, is that undesireable?

horrible-little-slime:
Mostly we're looking for:
 - It accomplishes its goal
 - It doesn't set anything on fire from the consequences of the code
 - It doesn't set anything on fire from the consequences of the decisions in the code

Shiverwarp:
Tested now, works fine (avoids equipping my skeleton lord shield)

PR #1132, opened 2022-10-01
Update dependencies.txt to use git if available

Rinn:


horrible-little-slime:
I'm not certain how desirable this is for scripts that are primarily set up for SVN. We _can_ get those Ezandora scripts via git, but do we actively want to?

The answer may very well be "yes." If it is, merge away.

Rinn:
i'll redo this later with the new github syntax that was added

horrible-little-slime (approved):

PR #1133, opened 2022-10-01
Remind users that nobarf may close NEP

JamesDowney:
NEP is a popular farming and leveling zone, there's a pretty good chance someone might run nobarf then expect to adventure in NEP. This will give us something to point at and say "hey you should read the manual, don't complain".

### DIFF
--- a/mafia/data/garbo_help.json
+++ b/mafia/data/garbo_help.json
@@ -3,7 +3,7 @@
 
   {
     "tableItem": "nobarf",
-    "description": "garbo will do beginning of the day setup, embezzlers, and various daily flags, but will terminate before normal Barf Mountain turns."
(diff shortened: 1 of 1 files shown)


gausie:
As someone who uses nobarf without plans to hit up NEP, I would be sad to miss out on the rewards for completing these quests.

I think this falls in with plenty of other resources that people may not want Garbo to use. We should instead consider giving some advice on how to make it avoid using them in your own scripts (like doing a withProperties around Garbo that pretends they're used already or something)

gausie:
Oh wait I got confused

horrible-little-slime:
I'm going to go ahead and merge this, but a more general "garbo will continue to use most resources as normal" disclaimer feels more appropriate.

PR #1134, opened 2022-10-01
Preserve NEP during nobarf

JamesDowney:
NEP is a popular farming and leveling zone, there's a pretty good chance someone might run nobarf then expect to adventure in NEP. This will prevent nobarf from closing NEP during nobarf.


gausie:
As someone who uses nobarf without plans to hit up NEP, I would be sad to miss out on the rewards for completing these quests.

I think this falls in with plenty of other resources that people may not want Garbo to use. We should instead consider giving some advice on how to make it avoid using them in your own scripts (like doing a withProperties around Garbo that pretends they're used already or something)

PR #1135, opened 2022-10-01
keepzonesopen arg

JamesDowney:
Adds the possibility of supplying an argument `keepzonesopen` to prevent garbo from taking actions that would close zones for the day. Currently only applies to NEP, but could be expanded to cover PR or FR if that content is ever utilized to an extent where the zones could be rendered inaccessible by garbo. An example of usage would be someone who wanted to run nobarf then farm little red dresses in NEP.

### DIFF
--- a/mafia/data/garbo_help.json
+++ b/mafia/data/garbo_help.json
@@ -30,6 +30,10 @@
     "description": "*EXPERIMENTAL* garbo will sacrifice some optimal behaviors to run quicker. Estimated and actual profits may be less accurate in this mode."
   },
   {
+    "tableItem": "keepzonesopen",
+    "description": "garbo will try to leave zones open that it could potentially close under normal circumstances - currently only applies to NEP."
+  },
+  {
     "tableItem": "Note on Arguments:",
     "description": "You can use multiple arguments in conjunction, e.g. \"garbo nobarf ascend\"."
   },
--- a/src/dailies.ts
+++ b/src/dailies.ts
@@ -515,7 +515,10 @@ function nepQuest(): void {
 
   if (get("_questPartyFair") === "unstarted") {
     visitUrl(toUrl($location`The Neverending Party`));
-    if (["food", "booze", "trash", "dj"].includes(get("_questPartyFairQuest"))) {
+    if (
+      ["food", "booze"].includes(get("_questPartyFairQuest")) ||
+      (["trash", "dj"].includes(get("_questPartyFairQuest")) && !globalOptions.noBarf) // Preserve NEP if running nobarf
+    ) {
       runChoice(1); // Accept quest
     } else {
       runChoice(2); // Decline quest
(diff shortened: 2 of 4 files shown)


horrible-little-slime:
I like this implementation a lot more than the last one!
The current implementation is still not ideal--right now it only covers one zone (the NEP), and in a way that is easily handled by a wrapper script or something else handled before running garbo. It also creates some scope creep--this feature could require frequent maintenance and have a chilling effect on expansions of the wanderer code or daily tasks.

I'd like to see some additions to our wanderer code to support not closing zones--right now, we theoretically could decide that our best YR target is in the post-war version of McMillicancuddy's Farm, for example, and in doing so we could close it. This is definitely out of scope, but if a user runs garbo at level 4, it would cause them to gain levels and lock them out of the cola wars battlefield (we absolutely do not need to support that). 

What are some other zones garbo could plausibly close by adventuring in them? How can we prevent garbo from doing that?

JamesDowney:
Can I just create a blocklist for zones that shouldn't be touched when `globalOptions.keepZonesOpen === true` and add that to the skiplist for canAdventureOrUnlock? for example:

``` ts
const canAdventureOrUnlockSkipList = [
  ...$locations`The Oasis, The Bubblin' Caldera, Barrrney's Barrr, The F'c'le, The Poop Deck, Belowdecks, 8-Bit Realm, Madness Bakery, The Secret Government Laboratory, The Dire Warren`,
  ...Location.all().filter((l) => l.parent === "Clan Basement"),
];
const closeableZones = $locations`The Neverending Party, McMillicancuddy's Farm`;
export function canAdventureOrUnlock(loc: Location): boolean {
  const skiplist = [...canAdventureOrUnlockSkipList];
  if (!have($item`repaid diaper`) && have($item`Great Wolf's beastly trousers`)) {
    skiplist.push($location`The Icy Peak`);
  }
  if (globalOptions.keepZonesOpen) {
    skiplist.push(...closeableZones);
  }
  const canUnlock = UnlockableZones.some((z) => loc.zone === z.zone && (z.available() || !z.noInv));
  return (
    !underwater(loc) &&
    !canAdventureOrUnlockSkipList.includes(loc) &&
    (canAdventure(loc) || canUnlock)
  );
}
```

JamesDowney:
Unrelated but in the previous example, ignoring what I added, it looks like `skiplist` isn't actually used so the conditional to add Icy Peak to the `skiplist` does nothing. The later check in the block of code uses `canAdventureOrUnlockSkipList`

JamesDowney:
Actually I suppose the issue with this implementation for NEP is that it wouldn't even use your 11 free fights in NEP. I think for NEP specifically the solution in dailies.ts is enough. But this would work for McMillicancuddy's Farm.

Shiverwarp:
As I'm now considering doing some farming at NEP, I would appreciate this being implemented!

PR #1140, opened 2022-10-03
Sssshhsssblllrrggghsssssggggrrgglsssshhssslblgl

Rinn:
do damage


horrible-little-slime:
This generally looks good; it might be worth splitting it up by class or mainstat, but also it might not

PR #1143, opened 2022-10-06
operation garbage cleanup: argmax => maxby

horrible-little-slime:


src/dailies.ts
```diff
@@ -180,12 +180,8 @@ function voterSetup(): void {
       })
     );
 
-    const initiativeValue = sum(
-      Array.from(availableInitiatives.values())
-        .sort((a, b) => b - a)
-        .slice(1),
-      (val) => 2 * val
-    );
+    const initiativeValue = 2 * Math.max(...availableInitiatives.values());
```
docrostov, line 183: this is more readable. good job. my one thing here would be that the actual list of valuations is kind of mediocre but it's fine; a few comments explaining would be nice. in particular, i think you could make a justification that famxp is a bit higher because it helps with goose leveling for the 50k meat skill.

src/dailies.ts
```diff
@@ -495,15 +490,19 @@ function configureVykea() {
...
-        1 * mallPrice($item`VYKEA instructions`));
-
-    if (vykeas.some(([level, cost]) => vykeaProfit(level, cost) > 0)) {
-      const level = vykeas.sort((a, b) => vykeaProfit(...b) - vykeaProfit(...a))[0][0];
+    const vykeaProfit = (vykea: [number, number]) => {
+      const [level, cost] = vykea;
+      return (
+        estimatedTurns() * baseMeat * 0.1 * level -
+        (5 * mallPrice($item`VYKEA rail`) +
+          cost * mallPrice($item`VYKEA dowel`) +
```
docrostov, line 498: okay i realize that this Works :tm: but i would recommend renaming "cost" here. the cost that is being used is the dowelCost, whereas the vykeaCost is the thing actually being calculated. by just naming it cost it reads like there's double counting going on until you reread the whole thing.

src/dailies.ts
```diff
@@ -579,23 +578,27 @@ function checkVolcanoQuest() {
...
-      quantity: get("_volcanoItemCount2"),
-      choice: 2,
-    },
-    {
-      item: property.getItem("_volcanoItem3") ?? $item.none,
-      quantity: get("_volcanoItemCount3"),
-      choice: 3,
-    },
-  ].reduce((a, b) => (volcanoItemValue(a) < volcanoItemValue(b) ? a : b));
+  const bestItem = maxBy(
```
docrostov, line 581: this is a place where maxBy is much more readable; good work.

src/dailies.ts
```diff
@@ -661,7 +664,7 @@ function tomeSummons(): void {
 function extrude(): void {
   if (SourceTerminal.have()) {
     const extrudeConsumables = $items`browser cookie, hacked gibson`;
-    const bestExtrude = extrudeConsumables.sort((a, b) => garboValue(b) - garboValue(a))[0];
+    const bestExtrude = maxBy(extrudeConsumables, garboValue);
```
docrostov, line 667: i have a hot take, which is that garbo should either include handling for software bug or a comment noting why we're ignoring software bug. it's reasonable to ignore it from a "slightly annoying to code" sense, since you can only generate 1/day, but people with ungodly amounts of source essence probably could use a few extra for slight extra profits. at least, used to -- it is possible the market on them has collapsed, and garbo would never buy them anyway. still, just a thought here.

src/dailies.ts
```diff
@@ -715,7 +718,7 @@ function internetMemeShop(): void {
 const teas = $items`cuppa Activi tea, cuppa Alacri tea, cuppa Boo tea, cuppa Chari tea, cuppa Craft tea, cuppa Cruel tea, cuppa Dexteri tea, cuppa Feroci tea, cuppa Flamibili tea, cuppa Flexibili tea, cuppa Frost tea, cuppa Gill tea, cuppa Impregnabili tea, cuppa Improprie tea, cuppa Insani tea, cuppa Irritabili tea, cuppa Loyal tea, cuppa Mana tea, cuppa Mediocri tea, cuppa Monstrosi tea, cuppa Morbidi tea, cuppa Nas tea, cuppa Net tea, cuppa Neuroplastici tea, cuppa Obscuri tea, cuppa Physicali tea, cuppa Proprie tea, cuppa Royal tea, cuppa Serendipi tea, cuppa Sobrie tea, cuppa Toast tea, cuppa Twen tea, cuppa Uncertain tea, cuppa Vitali tea, cuppa Voraci tea, cuppa Wit tea, cuppa Yet tea`;
 function pickTea(): void {
   if (!getCampground()["potted tea tree"] || get("_pottedTeaTreeUsed")) return;
-  const bestTea = teas.sort((a, b) => garboValue(b) - garboValue(a))[0];
+  const bestTea = maxBy(teas, garboValue);
   const shakeVal = 3 * garboAverageValue(...teas);
   const teaAction = shakeVal > garboValue(bestTea) ? "shake" : bestTea.name;
```
docrostov, line 723: honest thing -- i did not realize garbo had valuation for shaking the tea tree. that's awesome. good job fam.

src/dropsgear.ts
```diff
@@ -454,11 +453,8 @@ function juneCleaver(equipMode: BonusEquipMode): Map<Item, number> {
   }
   if (!juneCleaverEV) {
     juneCleaverEV =
-      JuneCleaver.choices.reduce(
-        (total, choice) =>
-          total +
-          valueJuneCleaverOption(juneCleaverChoiceValues[choice][bestJuneCleaverOption(choice)]),
-        0
+      sum([...JuneCleaver.choices], (choice) =>
+        valueJuneCleaverOption(juneCleaverChoiceValues[choice][bestJuneCleaverOption(choice)])
```
docrostov, line 457: i don't have any suggestions for improvements but this is... quite the statement lol. takes a little bit of noodling to figure out what's going on with the [] references and the functions. i don't necessarily think it needs a change but wanted to at least note it.

src/familiar/barfFamiliar.ts
```diff
@@ -164,7 +164,7 @@ export function barfFamiliar(): Familiar {
 
   if (viableMenu.length === 0) return meatFamiliar();
 
-  const best = viableMenu.reduce((a, b) => (totalFamiliarValue(a) > totalFamiliarValue(b) ? a : b));
+  const best = maxBy(viableMenu, totalFamiliarValue);
```
docrostov, line 167: this is more of a general comment than a comment about this explicitly. 

one small issue i have here -- and this is less an issue in vscode (which syntax highlights correctly) and more on github proper -- is that totalFamiliarValue isn't highlighted as a function on github. 

syntax-wise does it also work if you do totalFamiliarValue() or does it have to exclude the parens? if so i might request doing () to make it clear even on github review that this is a function. this is not a major bug in my craw or anything but it would slightly help cross-platform syntax highlighting.

src/fights.ts
```diff
@@ -2404,15 +2401,13 @@ function yachtzee(): void {
...
-              have(familiar) && familiar.underwater && familiar !== $familiar`Robortender`
-          )
-          .sort((a, b) => findLeprechaunMultiplier(b) - findLeprechaunMultiplier(a))[0] ??
-          $familiar.none
+      const familiarOptions = Familiar.all().filter(
+        (familiar) => have(familiar) && familiar.underwater && familiar !== $familiar`Robortender`
       );
+      const familiarChoice = familiarOptions.length
+        ? maxBy(familiarOptions, findLeprechaunMultiplier)
+        : $familiar.none;
```
docrostov, line 2409: this is way more readable than it was before and i am pleased with it. my only mild concern is the use of `$familiar.none;` i think that's a defensible choice given how Guys Always Exist :tm: but it does give me a mild amount of stomach-grumblin' pain to think of a user who uses garbo where garbo literally doesn't select a familiar for them. i do not have a solution for this, to be clear. just wanted to note it.

docrostov: actually wait i do have a solution. what if we just buy and use a lep and have a note/warning for the user telling the user we're going to buy a lep?

src/lib.ts
```diff
@@ -316,15 +310,16 @@ export function printHelpMenu(): void {
...
-    {
-      can: realmAvailable("sleaze"),
-      value: 40000,
-    },
-  ]
-    .filter((x) => x.can)
-    .sort((a, b) => b.value - a.value)[0];
+  const alternateUse = maxBy(
+    [
+      { can: canTreasury, value: 3 * get("valueOfAdventure") },
```
docrostov, line 315: okay i have a thought here. we store embezzlers as 3x VOA in a lot of different places. what if we start changing that to valueOfEmbezzler upwards in lib.ts, make that equal to 3x VOA for now, swap it out in the various locations it's done, and someday have better evaluation procedures? this way if we ever do a better job evaluating it we only have to do it in one spot.

src/lib.ts
```diff
@@ -560,3 +550,29 @@ export function freeCrafts(): number {
     (have($skill`Expert Corner-Cutter`) ? 5 - get("_expertCornerCutterUsed") : 0)
   );
 }
+
+export function maxBy<T>(
```
docrostov, line 554: please add docstrings to this. we're using it everywhere and it's cool as shit but this is exactly the kind of thing that warrants a docstring, especially with your multiple extensions of the core functions and overrides to allow you to do weird shit.

docrostov (changes requested):
the big request in here is the docstring request; i think that's 100% required before i approve. but overall i am highly supportive of this change and think it looks phenomenal. lots of places where this makes things much, much easier to read. great work phred.

src/dailies.ts
horrible-little-slime: Gonna declare that out of scope but a neat PR choice for someone

horrible-little-slime: Is there mafia tracking?

src/dropsgear.ts
horrible-little-slime: Yeah I would love to get better names for these the second someone else comes up with them

src/familiar/barfFamiliar.ts
horrible-little-slime: you would have to do `(x) => totalFamiliarValue(x)`, which I personally think is uglier but we each have our own feelings

horrible-little-slime: `totalFamiliarValue()` doesn't exist, because this function takes an input. When you do give it an input, it's a number; we want the function itself, and that means either passing just `totalFamiliarValue` or using arrows (or the `function` keyword, I guess) to construct a function using it. Which always feels a little silly to me

src/fights.ts
horrible-little-slime: It would need to be an urchin urchin, but frankly this has more to do with me being worried about using `reduce` on an empty array than it does to do with default options

Again, going to declare it out of scope but an interesting thing to explore in the future.

src/familiar/barfFamiliar.ts
docrostov: that's fine then. no big issue from me if it's not doable it's just an annoying syntax thing with github

src/lib.ts
docrostov: made separate issue; resolving convo

docrostov (dismissed):
thanks for the docstring changes. approved.

src/lib.ts
horrible-little-slime: Uhoh this one could crash in cs

horrible-little-slime:
Bman ran this without any notable errors; waiting on one other possible tester before merging

horrible-little-slime:
Also 3bh; merging

PR #1144, opened 2022-10-06
Support pure parka chains

Pantocyclus:
Also support < level 13 chains

### DIFF
--- a/src/yachtzee/diet.ts
+++ b/src/yachtzee/diet.ts
@@ -475,10 +475,10 @@ export function yachtzeeChainDiet(simOnly?: boolean): boolean {
   const possibleJellyYachtzeeTurns = [35, 30, 25, 20, 15, 10];
   const jellyYachtzeeTurns = possibleJellyYachtzeeTurns.find(sufficientOrgansFor) ?? 0;
 
-  const fishyTurnsAvailable =
+  const fishyPipeTurnsAvailable =
     haveEffect($effect`Fishy`) + (have($item`fishy pipe`) && !get("_fishyPipeUsed") ? 10 : 0);
   const canParkaChain =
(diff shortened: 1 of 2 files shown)


src/yachtzee/diet.ts
```diff
@@ -636,7 +647,7 @@ export function yachtzeeChainDiet(simOnly?: boolean): boolean {
       "orange"
     );
   }
-  if (jellyValuePerSpleen < extroValuePerSpleen && !simOnly) {
+  if (jellyValuePerSpleen < extroValuePerSpleen && !simOnly && jellyYachtzeeTurns > 0) {
```
horrible-little-slime, line 650: What's to stop this from deciding to pad out your parka-chain with jelly, then noticing that it's using jelly and aborting?

Pantocyclus: Ah ok I think I see your concern.

I assume what you're saying is that suppose
1) we have enough organ space for jellies, but
2) jellies just aren't profitable to use
then this code here returns without running a pure parka-chain; we should instead reevaluate by setting `jellyYachtzeeTurns=0` so we can continue trudging on without returning since we can still parka-chain
(The code in its current state only addresses allowing parka-chains if we don't have enough organ space to run a jelly-chain)

I can try to fix that later

src/yachtzee/diet.ts
```diff
@@ -472,7 +475,12 @@ export function yachtzeeChainDiet(simOnly?: boolean): boolean {
   const possibleJellyYachtzeeTurns = [35, 30, 25, 20, 15, 10];
   const jellyYachtzeeTurns = possibleJellyYachtzeeTurns.find(sufficientOrgansFor) ?? 0;
 
-  if (jellyYachtzeeTurns === 0) {
+  const fishyTurnsAvailable =
```
horrible-little-slime, line 478: can we rename this to `pipeTurnsAvailable`?

Other than that, looks good

PR #1150, opened 2022-10-08
Handle choices that follow fight

### LINKED ISSUE #1075: Garbo doesn't handle "Is There A Doctor In The House" after a map the monsters fight

libraryaddict:
Fixes #1075

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -886,7 +886,7 @@ const freeFightSources = [
           () => {
             mapMonster($location`Domed City of Grimacia`, $monster`grizzled survivor`);
             runCombat();
-            if (choiceFollowsFight()) runChoice(-1);
+            runChoice(-1);
           }
         );
       } else {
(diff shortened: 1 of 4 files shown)


src/fights.ts
```diff
@@ -1307,6 +1308,7 @@ const freeFightSources = [
           .abort()
           .setAutoAttack();
         mapMonster($location`The Haiku Dungeon`, $monster`amateur ninja`);
+        if (choiceFollowsFight()) runChoice(-1);
```
s-k-z, line 1311: Wouldn't this be simpler if moved inside of `mapMonster` in lib.ts?

libraryaddict: That's only because it was set as the auto attack. I believe sometimes the map monster call was used when a macro was not set as the auto attack.

s-k-z (approved):

Issue #1145, opened 2022-10-06
valuation of embezzlers


docrostov:
okay i have a thought here. we store embezzlers as 3x VOA in a lot of different places. what if we start changing that to valueOfEmbezzler upwards in lib.ts, make that equal to 3x VOA for now, swap it out in the various locations it's done, and someday have better evaluation procedures? this way if we ever do a better job evaluating it we only have to do it in one spot.

_Originally posted by @docrostov in https://github.com/Loathing-Associates-Scripting-Society/garbage-collector/pull/1143#discussion_r989383029_

i think this is a good idea and something we should try to do

horrible-little-slime:
Truly better valuation of embezzler's is garbo's biggest white whale, but this will make life easier when we harpoon her.

PR #1155, opened 2022-10-10
Lower embezzlerDifferential

Shiverwarp:
Just a better estimation for now from discussion in Discord


Tokoeka (dismissed):

Rinn (dismissed):

horrible-little-slime (changes requested):
As discussed in the discord, I'm going to ask for two things:

1. Make a named constant in lib called EMBEZZLER_MULTIPLIER and set it equal to whatever value we use, and apply it any and all places we use it. Except yachtzee, maybe
2. We need to collect some data to verify that 2.5 is correct; while it's true that flat bonuses don't get multiplied, we also run more buffs during embezzlers, which can make them _more_ valuable than the expected 4x.

Shiverwarp:
#1145 This implements what's discussed in this issue.

I'm not sure if we still want to gather data for it. I've stopped using my extros for embezzlers since I made this, so haven't got much more input

phulin:
> As discussed in the discord, I'm going to ask for two things:
> 
> 1. Make a named constant in lib called EMBEZZLER_MULTIPLIER and set it equal to whatever value we use, and apply it any and all places we use it. Except yachtzee, maybe
> 2. We need to collect some data to verify that 2.5 is correct; while it's true that flat bonuses don't get multiplied, we also run more buffs during embezzlers, which can make them _more_ valuable than the expected 4x.

Marginal embezzlers on my garbo are worth even a little less than 2.5x, from 7.5k or so to 17k or so.

Shiverwarp:
> Marginal embezzlers on my garbo are worth even a little less than 2.5x, from 7.5k or so to 17k or so.

Do you happen to know what non-limited buffs you use/purchase for your embezzlers that aren't used for Barf?

I'm suspecting that the 300% meat drop we lose from How To Scam Tourists is not made up for during embezzlers with those buffs, and this is in addition to the less additional value from our static drop items (LGR, Crown/Bjorn, thumb ring, etc)

Pantocyclus:
KGB, summoning chamber, pillkeeper-doubled frosty, zatara and meat.enh are a few that you would expect a shiny account to have for start-of-day embezzlers (that are not present during marginal embezzlers). Maybe horseradish and cologne too, but that heavily depends on the diet.

src/lib.ts
```diff
@@ -115,6 +115,7 @@ export type BonusEquipMode = "free" | "embezzler" | "dmt" | "barf";
 export const WISH_VALUE = 50000;
 export const HIGHLIGHT = isDarkMode() ? "yellow" : "blue";
 export const ESTIMATED_OVERDRUNK_TURNS = 60;
+export const EMBEZZLER_MULTIPLIER = 2.5;
```
docrostov, line 118: my only small note -- any appetite to make this toggleable by a runtime flag or a pref? only reason i ask -- it makes testing different values over a long period of time a little bit easier and i think it would be interesting to run garbo for a week while doing a walk from 1.5 to 3 by increments of 0.2 and seeing where the profitability sweet spot ends up if you take out outlier items and such. it is ok if the answer is no, i know we hate prefs, this just seems like a neat testing one.

docrostov (dismissed):
i honestly thought this was already in live. can we push this up? i think especially in the aftermath of the glooping change, this will have a pretty big impact in changing garbo behavior in a way that helps increase profitability for folks. we definitely overrate our embezzler meatdrop and this really helps allay it even if it isn't perfect. one small note here too but overall comfortable with pushing this guy up.

src/lib.ts
Shiverwarp: To be honest this should be estimateable by creating mock embezzler and barf outfits, and then comparing their meat drop% vs their bonus drops, and also comparing the meat mood for embezzlers vs meat mood for barf.

I just haven't figured out a good way to create these mock outfits to compare.  Using the 5 parameter `maximize` with `include_equip`  will give us a record that includes the equipment it wants to use, but I haven't worked much with this to really know for sure how realistic this is as a solution.

horrible-little-slime:
I'll merge this once the conflicts are resolved, but one of you should add an arg+pref for this as well.

horrible-little-slime (dismissed):

Shiverwarp:
This is the first pref/arg I've added, so definitely let me know if I've done anything incorrectly, or if the help descriptions aren't up to snuff

src/lib.ts
```diff
@@ -88,6 +89,7 @@ export type BonusEquipMode = "free" | "embezzler" | "dmt" | "barf";
 export const WISH_VALUE = 50000;
 export const HIGHLIGHT = isDarkMode() ? "yellow" : "blue";
 export const ESTIMATED_OVERDRUNK_TURNS = 60;
+export const EMBEZZLER_MULTIPLIER = globalOptions.embezzlerMultiplier;
```
horrible-little-slime, line 92: The way that this is set up, passing it as an argument won't work. That's (probably) fine, but worth checking; EMBEZZLER_MULTIPLIER gets initialized before we actually run Args.fill

the easiest fix to this is to make it a () => number and make it get called everywhere it shows up. But we'd also have to see if those are top-level code or not.

The real answer here is 'that's fine, whatever, who is going to be passing this as an arg instead of using a pref.' Which is to say: I'm going to merge this and then fix it myself later maybe

horrible-little-slime (approved):

PR #1157, opened 2022-10-11
June cleaver queue consideration

Shiverwarp:
I said I would stop thinking about it but I could not.

I feel _fairly_ decent about this, but still somewhat unsure.

If it looks okay to others, I can test it once I'm done my stint in hardcore

### DIFF
--- a/src/dropsgear.ts
+++ b/src/dropsgear.ts
@@ -457,6 +457,23 @@ function juneCleaver(equipMode: BonusEquipMode): Map<Item, number> {
         valueJuneCleaverOption(juneCleaverChoiceValues[choice][bestJuneCleaverOption(choice)])
       ) / JuneCleaver.choices.length;
   }
+  // If we're ascending then the chances of hitting choices in the queue is reduced
+  if (globalOptions.ascending && estimatedTurns() <= 180 && JuneCleaver.getInterval() === 30) {
+    const availEV =
+      sum([...JuneCleaver.choicesAvailable()], (choice) =>
+        valueJuneCleaverOption(juneCleaverChoiceValues[choice][bestJuneCleaverOption(choice)])
+      ) / JuneCleaver.choicesAvailable.length;
+    const queueEV =
+      (sum(
+        [...JuneCleaver.queue()],
+        (choice) =>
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
This looks broadly correct, and is exciting!

It is also somewhat cumbersome to navigate; it may be good to [SLAP](https://medium.com/javarevisited/slap-that-ugly-code-6ec276d3a4bc) this out into something that's a little more broken up, either by having a few distinct functions that you call or just by naming several variables. Possibly both!

I can talk through this in a little while and how we might make it read a little nicer

Shiverwarp:
Awesome!

Yeah it looked even worse before I broke it up into queueEV and availEV, so I'd definitely love to make it prettier/readable.

I'll read this SLAP article once I've done my turns for the day (If I still have energy left)

src/dropsgear.ts
```diff
@@ -457,6 +457,21 @@ function juneCleaver(equipMode: BonusEquipMode): Map<Item, number> {
...
+    const availEV =
+      sum([...JuneCleaver.choicesAvailable()], (choice) =>
+        valueJuneCleaverOption(juneCleaverChoiceValues[choice][bestJuneCleaverOption(choice)])
+      ) / JuneCleaver.choicesAvailable().length;
+    const queueEV =
+      sum(
+        [...JuneCleaver.queue()],
+        (choice) =>
+          valueJuneCleaverOption(juneCleaverChoiceValues[choice][bestJuneCleaverOption(choice)]) *
+          ((1 - 2 / 3) ^ Math.max(0, estimatedTurns() - (1 + JuneCleaver.queue().indexOf(choice))))
```
horrible-little-slime, line 471: Okay this line is definitely at least somewhat wrong insofar as estimatedTurns() will often be very large

Shiverwarp: Ah, it should be Math.floor(estimatedTurns()/30) instead of just estimatedTurns().

Do you have any suggestions for how to make this a bit more readable? I read the slap article but I'm not entirely clear on what a level of abstraction really means for something like this.

horrible-little-slime (approved):

PR #1165, opened 2022-10-20
Don't do kramco when romance is in the air

horrible-little-slime:


docrostov (approved):
this is a good idea and i endorse it 100%. good change.

PR #1166, opened 2022-10-20
Request\Demand Sandwich

Rinn:


horrible-little-slime:
Demand sandwich doesn't hurt,but we don't support Boris so it's a bit of an odd choice.

Is this part of breakfast right now?

Rinn:
Request isn't since it has a chance to fail and might need to be cast muliple times, I'm not sure about demand though. But I just added it because it took like 2 seconds to copy and paste the grimoire task

PR #1167, opened 2022-10-20
Open FantasyRealm and get the hat

Rinn:
you can smash extra hats and also this opens it up if we want to get the fat loot token with free kills


horrible-little-slime:
This is absolutely hilarious

PR #1168, opened 2022-10-20
Get shards of double-ice from april shower

Rinn:


Shiverwarp:
Maybe we should check if the player has unlocked all the recipes first from Hot shower? And if not do that instead?

Maximizing MP and doing some libram burning might also be comparable, depending? Would be a bit more effort to optimize that though

horrible-little-slime:
I think prioritizing hot shower for recipes could make sense, although players can always shower before using garbo. Same thing with getting a +%exp shower beforehand.

Rinn:
I'm generally of the opinion that running garbo means you're giving up control over your resources and if you want to use resources on something else then it should be done before running.

Shiverwarp:
I can definitely agree with giving up control. We just do something similar where we run hobo nickels with LGR if we don't have a hobo monkey.

The only other thing is we'd probably want to value the # of libram casts we could get with 1k mp vs the 3.5 double ice we'd get.

This is all pretty small potatoes though

horrible-little-slime:
I'm going to go ahead and say this is meaningfully different from hobo nickels because getting recipes can be accomplished by a wrapper or action outside of garbo. Whereas the closeting is something we do during garbo operations that you would have to, like, run a chat macro while garbo runs to undo. So I'll merge this.

PR #1169, opened 2022-10-20
Duplicate if available when spitting acid

Rinn:


Rinn:
just a note that duplicate will fail on a nocopy monster but it doesn't cause any issues

PR #1188, opened 2022-11-08
Value banishers for pygmy fights

s-k-z:
Don't drunk pygmy when its expensive

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -642,7 +642,7 @@ const pygmyBanishHandlers = [
     pygmy: $monster`pygmy janitor`,
     skill: undefined,
     check: undefined,
-    limit: undefined,
+    limit: 0,
     item: $item`tennis ball`,
   },
 ] as const;


src/fights.ts
```diff
@@ -961,7 +962,12 @@ const freeFightSources = [
       adventureMacro($location`The Hidden Bowling Alley`, pygmyMacro);
     },
     true,
-    {}
+    {
+      cost: () =>
+        sumNumbers(
+          $items`divine champagne popper, Louder Than Bomb, tennis ball`.map((i) => garboValue(i))
+        ) / 11,
```
horrible-little-slime, line 969:
```suggestion
        sum(
          $items`divine champagne popper, Louder Than Bomb, tennis ball`, mallPrice
        ) / 11,
```
We want to use mallPrice instead of garboValue because these are items we buy, not sell; ideally what we'd do is we'd also filter them out based on the skills we use in their stead. Also, go ahead and add the price of the bowl of scorptions to it.

horrible-little-slime (changes requested):

Shiverwarp:
This might be somewhat relevant to this (if not I can make my own PR sometime): Is it possible to maybe run Pygmies before we use up our skill banishes? 

For example, it'll go to haunted kitchen and use up all my feel hatreds and snokebombs getting my Hipster fights, before going to pygmies where it will buy the item banishers.

Perhaps this is already accounted for though, I don't know if hipster fights are worth more than the pygmy fights

Shiverwarp:
My assumption is that the pygmies are considered more valuable because we're actually bothering to buy these item banishers for them.

For the hipster fights, it is only valuing snokebomb and feel hatred as free runs, rather than banishes. Saving two skill banishes for pygmies would probably be ideal in our current situation

s-k-z:
Okay, you can make another PR to address those issues.

PR #1194, opened 2022-11-11
Use buff extenders even without embezzlers

Shiverwarp:
So the implementation is pretty simple (just use the buff extenders before we start running barf if we didn't use them during embezzlers)

But I'm unsure of where the function should live (if it even should?) my guess was we didn't just want just the raw code in index.ts

My suspicion is that the way I set it up (function living in lib.ts) is probably incorrect because it requires an import from clan.ts.

Perhaps it should be put in mood.ts instead?

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -206,10 +206,6 @@ function embezzlerSetup() {
   safeRestore();
   freeFightMood().execute(50);
   useBuffExtenders();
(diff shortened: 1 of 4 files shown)


horrible-little-slime:
Add License to Chill to this

Shiverwarp:
> Add License to Chill to this

Done

PR #1205, opened 2022-11-15
Cookbookbat drop rates are equal as of 2022-11-15

s-k-z:


horrible-little-slime (approved):
Let's go ahead and remove the () => number option later, maybe

PR #1207, opened 2022-11-18
Use writeCcs instead of garbo-combat.js.

phulin:
Reparsing garbo-combat.js every non-autoattack adventure takes unnecessary CPU. This approach should cover all cases; it's basically what grimoire does, and we could easily switch from this to full grimoireization.

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -1,6 +1,8 @@
 import "core-js/features/array/flat";
 import {
+  adv1,
   choiceFollowsFight,
+  cliExecute,
   equippedAmount,
   equippedItem,
   Familiar,
@@ -14,6 +16,7 @@ import {
   Item,
   itemAmount,
   itemType,
+  Location,
   mpCost,
   myAdventures,
   myBjornedFamiliar,
@@ -32,6 +35,7 @@ import {
   setAutoAttack,
   Skill,
   visitUrl,
+  writeCcs,
 } from "kolmafia";
 import {
   $class,
@@ -615,32 +619,11 @@ export class Macro extends StrictMacro {
   }
 }
 
-/**
- * Attempt to perform a nonstandard combat-starting Action with a Macro
- * @param macro The Macro to attempt to use
- * @param action The combat-starting action to attempt
- * @param tryAuto Whether or not we should try to resolve the combat with an autoattack; autoattack macros can fail against special monsters, and thus we have to submit a macro with Macro.save() regardless
- * @returns The output of your specified action function (typically void)
- */
-export function withMacro<T>(macro: Macro, action: () => T, tryAuto = false): T {
-  if (getAutoAttack() !== 0) setAutoAttack(0);
-  if (tryAuto) macro.setAutoAttack();
-  macro.save();
-  try {
-    return action();
-  } finally {
-    Macro.clearSaved();
-    if (tryAuto) setAutoAttack(0);
-  }
-}
-
-export function main(): void {
+function customizeMacro<M extends StrictMacro>(macro: M) {
   if (have($effect`Eldritch Attunement`)) {
-    Macro.if_($monster`Eldritch Tentacle`, Macro.basicCombat())
-      .step(Macro.load())
-      .submit();
+    return Macro.if_($monster`Eldritch Tentacle`, Macro.basicCombat()).step(macro);
   } else if (getTodaysHolidayWanderers().length !== 0) {
-    Macro.ifHolidayWanderer(
+    return Macro.ifHolidayWanderer(
       Macro.externalIf(
         haveEquipped($item`backup camera`) &&
           get("_backUpUses") < 11 &&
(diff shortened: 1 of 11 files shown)


phulin:
Hold - saw some weird stuff while testing today, will investigate to see if it is this PR.

src/combat.ts
```diff
@@ -615,46 +619,90 @@ export class Macro extends StrictMacro {
...
+      )
+    ).step(macro);
+  } else {
+    return macro;
+  }
+}
+
+function makeCcs<M extends StrictMacro>(macro: M) {
+  writeCcs(`[default]\n"${customizeMacro(macro).toString()}"`, "garbo");
+  cliExecute("ccs garbo");
```
horrible-little-slime, line 643:
```suggestion
 propertiesManager.set("customCombatScript", "garbo");
```

src/combat.ts
```diff
@@ -615,46 +619,90 @@ export class Macro extends StrictMacro {
...
+ * Adventure in a location and handle all combats with a given macro.
+ * To use this function you will need to create a consult script that runs Macro.load().submit() and a CCS that calls that consult script.
+ * See examples/consult.ts for an example.
+ *
+ * @category Combat
+ * @param loc Location to adventure in.
+ * @param macro Macro to execute.
+ */
+export function garboAdventure<M extends StrictMacro>(loc: Location, macro: M): void {
+  if (getAutoAttack() !== 0) setAutoAttack(0);
```
horrible-little-slime, line 685: We might want to always call this; I've seen some issues with the `aa` cli command that veracity detailed [here](https://kolmafia.us/threads/mafia-appears-not-to-be-tracking-_spacejellyfishdrops-when-i-autoattack-with-a-macro-r26881.28212/#post-170343).

src/combat.ts
```diff
@@ -615,46 +619,90 @@ export class Macro extends StrictMacro {
...
+ * Adventure in a location and handle all combats with a given autoattack and manual macro.
+ * To use the nextMacro parameter you will need to create a consult script that runs Macro.load().submit() and a CCS that calls that consult script.
+ * See examples/consult.ts for an example.
+ *
+ * @category Combat
+ * @param loc Location to adventure in.
+ * @param autoMacro Macro to execute via KoL autoattack.
+ * @param nextMacro Macro to execute manually after autoattack completes.
+ */
+export function garboAdventureAuto<M extends StrictMacro>(
```
horrible-little-slime, line 700: Do we ever want to use something other than a Garbo macro? Is there a benefit to genericity here?

src/combat.ts
```diff
@@ -615,46 +619,90 @@ export class Macro extends StrictMacro {
...
+function completeCombat<T>(action: () => T) {
+  try {
+    const result = action();
+    while (inMultiFight()) runCombat();
+    if (choiceFollowsFight()) visitUrl("choice.php");
+    return result;
+  } catch (e) {
+    throw `Combat exception! Last macro error: ${get("lastMacroError")}. Exception ${e}.`;
+  }
+}
```
horrible-little-slime, line 655: IMO it is not clear from the name that `action` should be an action that initiates combat. To me, `completeCombat` sounds like a function that finishes an already-started combat, and `action` sounds like something that happens _during_ combat.

src/yachtzee/index.ts
```diff
@@ -115,7 +114,7 @@ function _yachtzeeChain(): void {
         useSkill($skill`The Polka of Plenty`);
       }
     }
-    adventureMacro($location`The Sunken Party Yacht`, Macro.abort());
+    garboAdventure($location`The Sunken Party Yacht`, Macro.abort());
```
horrible-little-slime, line 117: is it worth making a special function for when we want to pass `Macro.abort()`?

horrible-little-slime: Just feels like it happens a decent amount.

webpack.config.js
```diff
@@ -45,7 +45,6 @@ const scriptsConfig = merge(
   {
     entry: {
       garbo: "./src/index.ts",
-      "garbo-combat": "./src/combat.ts",
```
horrible-little-slime, line 48: Hell yeah

src/yachtzee/index.ts
phulin: Yeah, but outside scope of this I think

src/combat.ts
phulin: renamed

src/combat.ts
phulin: There is a compelling reason that I had two days ago and now I have no recollection of what it is.

src/combat.ts
phulin: I think that triggers an extra server request?

src/combat.ts
phulin: 👍

phulin:
Appears to be fully working now.

horrible-little-slime (approved):

PR #1229, opened 2022-12-28
Use train whistle in glitch fight

CheeseyPickle:


CheeseyPickle:
Should probably wait until train whistle is recognized as an item :P

PR #1231, opened 2023-01-02
Get training scrolls from snojo

midgleyc:
Over a long aftercore, you can pick up one of each of the training scrolls from the snojo. Do so.

Implementation is casual, and assumes the user isn't changing the options themselves. It finishes in Moxie, like before.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -1236,19 +1236,11 @@ const freeFightSources = [
   ),
 
   new FreeFight(
-    () => get("snojoAvailable") && clamp(10 - get("_snojoFreeFights"), 0, 10),
+    () =>
+      get("snojoAvailable") &&
+      get("snojoSetting") !== null &&
+      clamp(10 - get("_snojoFreeFights"), 0, 10),
     () => {
-      const snojoSetting = get("snojoSetting", "NONE");
-      if (snojoSetting === "MYSTICALITY" && get("snojoMysticalityWins") > 50) {
-        visitUrl("place.php?whichplace=snojo&action=snojo_controller");
-        runChoice(3);
-      } else if (snojoSetting === "MUSCLE" && get("snojoMuscleWins") > 50) {
-        visitUrl("place.php?whichplace=snojo&action=snojo_controller");
-        runChoice(2);
-      } else if (snojoSetting === "NONE") {
-        visitUrl("place.php?whichplace=snojo&action=snojo_controller");
-        runChoice(1);
-      }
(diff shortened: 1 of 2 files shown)


src/fights.ts
```diff
@@ -1238,9 +1238,16 @@ const freeFightSources = [
...
+      const snojoSetting = get("snojoSetting", "NONE");
+      if (snojoSetting === "MYSTICALITY" && get("snojoMysticalityWins") > 50) {
         visitUrl("place.php?whichplace=snojo&action=snojo_controller");
         runChoice(3);
+      } else if (snojoSetting === "MUSCLE" && get("snojoMuscleWins") > 50) {
+        visitUrl("place.php?whichplace=snojo&action=snojo_controller");
+        runChoice(2);
+      } else if (snojoSetting === "NONE") {
+        visitUrl("place.php?whichplace=snojo&action=snojo_controller");
+        runChoice(1);
```
horrible-little-slime, line 1250: We should make setting the snojo up a daily task instead of being part of this FreeFight, probably.

Then in the task we can have something like ```
type SnojoSetting = "MUSCLE" | "MYSTICALITY" | "MOXIE" | "NONE";

const SETTINGS = {
  MOXIE: { choice: 1, pref: "snojoMoxieWins", },
  MUSCLE: { choice: 2, pref: "snojoMuscleWins", },
  MYSTICALITY: { choice: 3, pref: "snojoMysticalityWins", },
} as const;

const currentSetting = get<SnojoSetting>("snojoSetting", "NONE");
```
and go from there.

But I think now that this logic is becoming something other than "always pick x", it needs to move to dailies.

midgleyc: I guess if it's being split out I should do it properly.

The calculation is something like:
* if you're ascending pick the highest of the values of each of the consumables
* otherwise assume you're in for at least 5 days, and calculate the higher of (scroll + 7 * consumables) / 5 or each consumable and pick accordingly, discarding options where you've already got the scroll.

midgleyc:
Updated dailies, but not tested. How do I rebuild garbo fit for install?

horrible-little-slime:
You should be able to `yarn install` and then `yarn build` if your dev environment is all set up.

If it isn't, the README of this repository should be helpful:
https://github.com/docrostov/kol-ts-starter

horrible-little-slime (approved):
Approved, although it looks like Ryo plans to test it so I'll wait around a minute for that

midgleyc:
I appreciate the vote of confidence in the code at least :)

The daily ran as I expected so I think this is good now.

PR #1233, opened 2023-01-02
Implement Autumnaton Valuation

Shiverwarp:
Well this was an experience! This took a lot of effort, but I'm feeling proud of myself. I'm sure it's sloppy compared to the grown up programmers here though.

So, the possible issues still remaining that I haven't thought of good answers for:

- Doesn't dress for zones we don't have access to. I can't think of a way to start tackling this. Does mafia have some way to see requirements for accessing zones? 
- For actually adventuring in zones to allow autumnaton access, it's possible to find the overall most profitable zone out of all zones in the game, and then compare that to our current best available zone, and if the expected profit over the day is higher than the cost of a free run skill/item, then we do that.  

However this would require the autumnaton stuff being plugged into the more complicated stuff of garbo and I'm not too comfortable there.  It's also possible this ends up being in some zone we can't free run from, or has some other access restrictions we can't meet, etc.
- When considering what upgrades to go for, doesn't consider the extra turns available if we aren't ascending. This could be alleviated or estimated by adding a flat value (perhaps around 400-600) of estimated Turns to the `expectedRemainingExpeditions` function or doubling them if we aren't ascending. This is not "accurate" but it would do a good enough job because we would be going for the upgrades at the start of the day anyway if they're optimal to go for.
- The upgrade choosing code is rather hefty because we need to consider 7 different upgrade locations (plus the overall best one) and then also 3 upgrade states that could change what the optimal location is (The starting upgrade state, the extra acuity state, and the extra slot state) While this code is a bit cumbersome, I was able to adhere better to your Loathers convention of only using `const`
-` expectedRemainingExpeditions` I assume can be done with some math magic (like how estimated remaining digitizes is done)? Instead of my while loop and two lets, but I am not a math magician
-Cookbookbat values are not included in this, but aren't relevant for choosing where we want to send it because they're random anyway, so shouldn't matter

### DIFF
--- a/src/autumnaton.ts
+++ b/src/autumnaton.ts
@@ -1,0 +1,334 @@
+import { maxBy } from "./lib";
+import { garboValue } from "./session";
+import { estimatedTurns } from "./turns";
+import {
+  appearanceRates,
+  getLocationMonsters,
+  Item,
+  itemDropsArray,
+  Location,
+  toMonster,
+} from "kolmafia";
+import { $item, $items, AutumnAton, get, sum } from "libram";
+
+function getAutumnatonUniques(location: Location): [AutumnAton.Upgrade, Item] {
+  switch (location.environment) {
+    case "outdoor":
+      switch (location.difficultyLevel) {
+        case "low":
+          return [AutumnAton.possibleUpgrades[4], $item`autumn leaf`];
+        case "mid":
+          return [AutumnAton.possibleUpgrades[2], $item`autumn debris shield`];
+        case "high":
+          return [AutumnAton.possibleUpgrades[6], $item`autumn leaf pendant`];
+      }
+      break;
+    case "indoor":
+      switch (location.difficultyLevel) {
+        case "low":
+          return [AutumnAton.possibleUpgrades[0], $item`AutumnFest ale`];
+        case "mid":
+          return [AutumnAton.possibleUpgrades[3], $item`autumn-spice donut`];
+        case "high":
+          return [AutumnAton.possibleUpgrades[7], $item`autumn breeze`];
+      }
+      break;
+    case "underground":
+      switch (location.difficultyLevel) {
+        case "low":
+          return [AutumnAton.possibleUpgrades[1], $item`autumn sweater-weather sweater`];
+        case "mid":
+          return [AutumnAton.possibleUpgrades[5], $item`autumn dollar`];
+        case "high":
+          return [AutumnAton.possibleUpgrades[8], $item`autumn years wisdom`];
+      }
+      break;
+  }
+  // Just return a default value if location info is incorrect, libram types don't accept undefined
+  return [AutumnAton.possibleUpgrades[0], $item`AutumnFest ale`];
+}
+
+export function averageAutumnatonValue(
+  location: Location,
+  acuityOverride?: number,
+  slotOverride?: number
+): number {
+  const badAttributes = ["LUCKY", "ULTRARARE", "BOSS"];
+  const rates = appearanceRates(location);
+  const monsters = Object.keys(getLocationMonsters(location))
+    .map((m) => toMonster(m))
+    .filter((m) => !badAttributes.some((s) => m.attributes.includes(s)) && rates[m.name] > 0);
+
+  if (monsters.length === 0) {
+    return 0;
+  } else {
+    interface ItemData {
+      value: number;
+      rate: number;
+      expectedDrops1?: number;
+      expectedDrops2?: number;
+      totalExpectedDrops?: number;
+    }
+
+    const validDrops: ItemData[] = [];
+    // Get all valid drops from monsters in the zone.
+    // If mafia doesn't know the droprate, or it is conditional, we value it at 0
+    monsters.forEach((m) => {
+      itemDropsArray(m).forEach((item) => {
+        if (item.type.includes("c" || "0")) {
+          validDrops.push({ value: 0, rate: item.rate });
+        } else if (item.type === "") {
+          validDrops.push({ value: garboValue(item.drop, true), rate: item.rate });
+        }
+      });
+    });
+
+    let totalZoneExpectedDrops = 0;
+    // Find the expected drops for each valid item
+    validDrops.forEach((d) => {
+      // First two rolls do not care about acuity
+      const acuityCutoff = 20 - (acuityOverride ? acuityOverride : AutumnAton.visualAcuity() * 5);
+      d.expectedDrops1 = Math.min(
+        slotOverride ? slotOverride : AutumnAton.zoneItems(),
+        (d.rate / 100) * 2
+      );
+      // Last 8 rolls do not count items below the acuity cutoff
+      // Our max capacity is reduced by the expected drops from the first 2 rolls
+      if (d.rate < acuityCutoff) {
+        d.rate = 0;
+      }
+      d.expectedDrops2 = Math.min(
+        (slotOverride ? slotOverride : AutumnAton.zoneItems()) - d.expectedDrops1,
+        (d.rate / 100) * 8
+      );
+      totalZoneExpectedDrops += d.expectedDrops1 + d.expectedDrops2;
+    });
+
+    let expectedCollectionValue = 0;
+    validDrops.forEach((d) => {
+      // This makes sure that when we have a larger amount of total expected drops than we have room for, we still return the correct adjusted amount for our available slots
+      if (d.expectedDrops1 && d.expectedDrops2) {
+        d.totalExpectedDrops =
+          ((d.expectedDrops1 + d.expectedDrops2) / totalZoneExpectedDrops) *
+          Math.min(totalZoneExpectedDrops, slotOverride ? slotOverride : AutumnAton.zoneItems());
+      }
+      if (d.totalExpectedDrops) expectedCollectionValue += d.totalExpectedDrops * d.value;
+    });
+    return seasonalItemValue(location) + expectedCollectionValue;
+  }
+}
+
+function seasonalItemValue(location: Location, seasonalOverride?: number): number {
+  // Find the value of the drops based on zone difficulty/type
+  const autumnItems = $items`autumn leaf, AutumnFest ale, autumn breeze, autumn dollar, autumn years wisdom`;
+  const avgValueOfRandomAutumnItem =
+    sum(autumnItems, (item) => garboValue(item, true)) / autumnItems.length;
+  const autumnMeltables = $items`autumn debris shield, autumn leaf pendant, autumn sweater-weather sweater`;
+  const autumnItem = getAutumnatonUniques(location)[1];
+  const seasonalItemDrops = seasonalOverride ? seasonalOverride : AutumnAton.seasonalItems();
+  return (
+    (seasonalItemDrops > 1 ? avgValueOfRandomAutumnItem : 0) +
+    (autumnMeltables.includes(autumnItem)
+      ? avgValueOfRandomAutumnItem
+      : garboValue(autumnItem, true))
+  );
+}
+
+function expectedRemainingExpeditions(legOverride?: number): number {
+  const availableAutumnatonTurns = estimatedTurns() - AutumnAton.turnsLeft();
+  let expeditionTurnSum = 0;
+  let quests = get("_autumnatonQuests", 0);
+  while (expeditionTurnSum < availableAutumnatonTurns) {
+    expeditionTurnSum +=
+      11 *
+      Math.max(
+        1,
+        quests -
+          (legOverride
+            ? legOverride
+            : AutumnAton.currentUpgrades().filter((u) => u.includes("leg")).length)
+      );
+    quests++;
+  }
+
+  return quests - get("_autumnatonQuests", 0);
+}
+
+export function prioritizeUpgradeLocations(fullLocations: Location[]): Location[] {
+  // This function shouldn't be getting called if we don't have an expedition left
+  if (expectedRemainingExpeditions() < 1) {
+    return fullLocations;
+  }
+  const profitRelevantUpgrades = [
+    "leftarm1",
+    "leftleg1",
+    "rightarm1",
+    "rightleg1",
+    "cowcatcher",
+    "periscope",
+    "radardish",
+  ];
+  const currentUpgrades = AutumnAton.currentUpgrades() as string[];
+  const acquirableUpgrades = profitRelevantUpgrades.filter(
+    (upgrade) => !currentUpgrades.includes(upgrade)
+  );
+  // Libram doesn't have amount of leg upgrades specifically like the other types do
+  const legUpgrades = AutumnAton.currentUpgrades().filter((u) => u.includes("leg")).length;
+
+  const acuityUpgrade1Possible = acquirableUpgrades.includes("periscope");
+  const acuityUpgrade2Possible = acquirableUpgrades.includes("radardish");
+  const leg1UpgradePossible = acquirableUpgrades.includes("rightleg1");
+  const leg2UpgradePossible = acquirableUpgrades.includes("leftleg1");
+  const arm1UpgradePossible = acquirableUpgrades.includes("rightarm1");
+  const arm2UpgradePossible = acquirableUpgrades.includes("leftarm1");
+  const cowCatcherUpgradePossible = acquirableUpgrades.includes("cowcatcher");
+
+  const acuity1Locations = fullLocations.filter(
+    (location) => getAutumnatonUniques(location)[0] === "periscope"
+  );
+  const acuity2Locations = fullLocations.filter(
+    (location) => getAutumnatonUniques(location)[0] === "radardish"
+  );
+  const leg1Locations = fullLocations.filter(
+    (location) => getAutumnatonUniques(location)[0] === "rightleg1"
+  );
+  const leg2Locations = fullLocations.filter(
+    (location) => getAutumnatonUniques(location)[0] === "leftleg1"
+  );
+  const arm1Locations = fullLocations.filter(
+    (location) => getAutumnatonUniques(location)[0] === "rightarm1"
+  );
+  const arm2Locations = fullLocations.filter(
+    (location) => getAutumnatonUniques(location)[0] === "leftarm1"
+  );
+  const cowCatcherLocations = fullLocations.filter(
+    (location) => getAutumnatonUniques(location)[0] === "cowcatcher"
+  );
+
+  const bestLocation = maxBy(fullLocations, (l: Location) => averageAutumnatonValue(l));
+  const bestAcuity1Location = maxBy(acuity1Locations, (l: Location) => averageAutumnatonValue(l));
+  const bestAcuity2Location = maxBy(acuity2Locations, (l: Location) => averageAutumnatonValue(l));
+  const bestLocationWithAcuity1 = maxBy(fullLocations, (l: Location) =>
+    averageAutumnatonValue(
+      l,
+      acuityUpgrade1Possible ? AutumnAton.visualAcuity() + 1 : AutumnAton.visualAcuity()
+    )
+  );
+  const bestLocationWithAcuity2 = maxBy(fullLocations, (l: Location) =>
+    averageAutumnatonValue(
+      l,
+      acuityUpgrade2Possible ? AutumnAton.visualAcuity() + 1 : AutumnAton.visualAcuity()
+    )
+  );
+  // Legs do not change the best overall location when we have them upgraded
+  const bestLeg1Location = maxBy(leg1Locations, (l: Location) => averageAutumnatonValue(l));
+  const bestLeg2Location = maxBy(leg2Locations, (l: Location) => averageAutumnatonValue(l));
+
+  const bestArm1Location = maxBy(arm1Locations, (l: Location) => averageAutumnatonValue(l));
+  const bestArm2Location = maxBy(arm2Locations, (l: Location) => averageAutumnatonValue(l));
+  const bestLocationWithArm1 = maxBy(fullLocations, (l: Location) =>
+    averageAutumnatonValue(
+      l,
+      undefined,
+      arm1UpgradePossible ? AutumnAton.zoneItems() + 1 : AutumnAton.zoneItems()
+    )
+  );
+  const bestLocationWithArm2 = maxBy(fullLocations, (l: Location) =>
+    averageAutumnatonValue(
+      l,
+      undefined,
+      arm2UpgradePossible ? AutumnAton.zoneItems() + 1 : AutumnAton.zoneItems()
+    )
+  );
+  // Cow catcher also does not change best overall location after upgrading
+  const bestCowCatcherLocation = maxBy(cowCatcherLocations, (l: Location) =>
+    averageAutumnatonValue(l)
+  );
+
+  const currentExpectedProfit =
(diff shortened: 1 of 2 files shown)


Tokoeka:
>     * Doesn't dress for zones we don't have access to. I can't think of a way to start tackling this. Does mafia have some way to see requirements for accessing zones?

I believe the way to do this may be preValidateAdventure(location). (pVA for short in the below example)

As an example (in ash), pVA($location[the mouldering mansion]) returns true at the moment as I can take a potion to go there, whilst canAdventure($location[the mouldering mansion]) returns false as I do not currently have the relevant effect.

i dont think preValidateAdventure() / canAdventure() is fully working yet (as evidenced by both returning true for $location[the deep machine tunnels] despite my active familiar not being the machine elf and not having the "Inside the snowglobe" effect, but its better then nothing?

src/post.ts
```diff
@@ -189,7 +191,9 @@ function funguySpores() {
   }
 }
 
-const autumnAtonZones = $locations`The Toxic Teacups, The Oasis, The Deep Dark Jungle, The Bubblin' Caldera, The Sleazy Back Alley`;
+const autumnatonLocations = prioritizeUpgradeLocations(
+  Location.all().filter((l) => l.turnsSpent && canAdventure(l))
+);
```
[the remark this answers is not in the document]
Shiverwarp: Sorry I think I'm missing something in this suggested change, do we not want to prioritize the upgrade locations?

src/autumnaton.ts
```diff
@@ -0,0 +1,334 @@
...
+        case "mid":
+          return [AutumnAton.possibleUpgrades[5], $item`autumn dollar`];
+        case "high":
+          return [AutumnAton.possibleUpgrades[8], $item`autumn years wisdom`];
+      }
+      break;
+  }
+  // Just return a default value if location info is incorrect, libram types don't accept undefined
+  return [AutumnAton.possibleUpgrades[0], $item`AutumnFest ale`];
+}
```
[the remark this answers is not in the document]
Shiverwarp: Mostly I just took this from mafia's implementation of it here: 
[](https://github.com/kolmafia/kolmafia/blob/dfe5a6aae0532e81b91a0feaaa555183e093642e/src/net/sourceforge/kolmafia/textui/command/AutumnatonCommand.java#L240)

Is this something I should change now in this PR, or try to put it in libram first?

src/autumnaton.ts
```diff
@@ -0,0 +1,334 @@
...
+    });
+
+    let expectedCollectionValue = 0;
+    validDrops.forEach((d) => {
+      // This makes sure that when we have a larger amount of total expected drops than we have room for, we still return the correct adjusted amount for our available slots
+      if (d.expectedDrops1 && d.expectedDrops2) {
+        d.totalExpectedDrops =
+          ((d.expectedDrops1 + d.expectedDrops2) / totalZoneExpectedDrops) *
+          Math.min(totalZoneExpectedDrops, slotOverride ? slotOverride : AutumnAton.zoneItems());
+      }
```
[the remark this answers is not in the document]
Shiverwarp: Not sure if you'd rather talk here or discord but:

This portion is basically "squashing" down the overall expected amount of total drops to fit within the total amount of collection slots that we have, based on their percentage of the overall total expected drops (before squashing)

src/post.ts
```diff
@@ -205,6 +209,6 @@ export default function postCombatActions(skipDiet = false): void {
   stillsuit();
   funguySpores();
   if (globalOptions.ascending || AutumnAton.turnsForQuest() < myAdventures() + 10) {
-    AutumnAton.sendTo(autumnAtonZones);
+    AutumnAton.sendTo(maxBy(autumnatonLocations, (location) => averageAutumnatonValue(location)));
```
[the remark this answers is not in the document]
Shiverwarp: If we do this, we cannot reduce down our list of locations to prioritize the upgrade locations, or do I misunderstand the libram function?

Shiverwarp:
> > ```
> > * Doesn't dress for zones we don't have access to. I can't think of a way to start tackling this. Does mafia have some way to see requirements for accessing zones?
> > ```
> 
> I believe the way to do this may be preValidateAdventure(location). (pVA for short in the below example)
> 
> As an example (in ash), pVA($location[the mouldering mansion]) returns true at the moment as I can take a potion to go there, whilst canAdventure($location[the mouldering mansion]) returns false as I do not currently have the relevant effect.
> 
> i dont think preValidateAdventure() / canAdventure() is fully working yet (as evidenced by both returning true for $location[the deep machine tunnels] despite my active familiar not being the machine elf and not having the "Inside the snowglobe" effect, but its better then nothing?

I'm unsure of how I would curate the list of possible locations to include zones that require things like this is the only problem (if this were possible, it'd probably also be easier to implement unlocking zones to send autumnaton to as well)

src/post.ts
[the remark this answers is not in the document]
horrible-little-slime:
```suggestion
    AutumnAton.sendTo((locations) => maxBy(prioritizeUpgradeLocations(locations), averageAutumnatonValue));
```

Although it feels like the prioritization logic should just be part of the `value` function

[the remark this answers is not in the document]
Shiverwarp: As it stands I'm a little unsure of how to do that easily because of how I use the `averageAutumnatonValue` function in `prioritizeUpgradeLocations` itself. I'll make this smaller proposed change for now

Veracity0:
> I believe the way to do this may be preValidateAdventure(location). (pVA for short in the below example)
> 
> As an example (in ash), pVA($location[the mouldering mansion]) returns true at the moment as I can take a potion to go there, whilst canAdventure($location[the mouldering mansion]) returns false as I do not currently have the relevant effect.
> 
> i dont think preValidateAdventure() / canAdventure() is fully working yet (as evidenced by both returning true for $location[the deep machine tunnels] despite my active familiar not being the machine elf and not having the "Inside the snowglobe" effect, but its better then nothing?

I think you are missing that adventure validation is in three parts, not two.

preValidateAdventure - returns false if the zone is unavailable without using an expensive item - like a charter day pass.
It will look at container zones - maps to see if we didn’t notice that you’d used a day pass and opened the zone for today

canAdventure - looks at quest setting and active effects and familiars and such to see what you have opened.
It also looks at items which will give you needed effects and familiars which you have in your terrarium.

prepareForAdventure - will use items or pull out familiars or visit NPCs or put on outfits or do other things that do not consume turns to finish the job.

This trio of validations is used by KoLmafia when it automates for you - via GUI or CLI “adventure” command or ASH (JS) adventure or adv1 functions.

If you use preAdventure/canAdventure and they return true, I believe they accurately tell you that you have everything you need to get to the zone and when you actually automate the adventure, THEN the outfits/familiar/item->effect/etc. will happen.

canAdventure for the DMT returns true for the following:

- you can use the Machine Elf - you have one equipped or in terrarium, in standard
- you have the snow globe effect active
- you have a snow globe

prepareForAdventure for the DMT will do one of the following, if needed:

- remove Machine Elf from terrarium
- use snow globe

I have not marked the kolmafia.us thread as Implemented yet because there are a (tiny) handful of zones I don’t fully support yet.

My TBD list is:

13) Batfellow Area
   Is a LimitMode
   Tests?
14) Spelunky Area
   Is a LimitMode
   Tests?
15) Holidays
    Write tests
16) PirateRealm
   - canAdventure
   -- Limited turns per day
   -- Requires pirate eyepatch
   - prepareForAdventure
   -- equip pirate eyepatch
   - Write test
 17) Casino
   - canAdventure
   -- must have casino pass or enough Meat
   - prepareForAdventure
   -- will retrieve/buy casino pass if necessary
   - Write test

Note that Batfellow and Spelunky and Holidays are “write tests to confirm things work.
PirateRealm needs actual support.
Casino? Meh. Before actually submitting a URL there, KoLmafia will get you a casn=ino pass - even in the Relay Browser. So, that’s not part of “automation” - but adventure validation should ensure that it will work.

I am unaware of any other areas that are “not fully working yet”.

Feel free to submit bugs/comments/question on the forum thread.

https://kolmafia.us/threads/ash-location_accessible-function.27836/

Veracity0:
> As an example (in ash), pVA($location[the mouldering mansion]) returns true at the moment as I can take a potion to go there, whilst canAdventure($location[the mouldering mansion]) returns false as I do not currently have the relevant effect.

Wormwood is an example where we know the needed effect and the potion that provides it and are willing to use it to get the effect.
```
> inv absinthe

tiny bottle of absinthe (5)

> ash have_effect($effect[Absinthe-Minded])

Returned: 0

> ash pre_validate_adventure($location[the mouldering mansion])

You are too drunk to continue.
Returned: false

> ash can_adventure($location[the mouldering mansion])

Returned: true
```
Trust me - had I not been overdrunk, pre_validate_adventure() would have returned true. :)

gausie:
> I am unaware of any other areas that are “not fully working yet”.

There are also clan dungeon zones which I started writing but, for obvious reasons, felt a bit snowed-under by and put it on pause.

src/autumnaton.ts
```diff
@@ -0,0 +1,217 @@
...
+  }
+  const profitRelevantUpgrades = [
+    "leftarm1",
+    "leftleg1",
+    "rightarm1",
+    "rightleg1",
+    "cowcatcher",
+    "periscope",
+    "radardish",
+  ] as const;
```
horrible-little-slime, line 144: Let's define this out of the scope of the function

src/autumnaton.ts
```diff
@@ -0,0 +1,217 @@
...
+  // Libram doesn't have amount of leg upgrades specifically like the other types do
+  const legUpgrades = AutumnAton.currentUpgrades().filter((u) => u.includes("leg")).length;
```
horrible-little-slime, line 146: I'll add this to libram and cut a new version Soon:tm:

src/autumnaton.ts
```diff
@@ -0,0 +1,217 @@
...
+    (upgrade) => !currentUpgrades.includes(upgrade)
+  );
+  const currentBestLocation = maxBy(fullLocations, (loc: Location) => averageAutumnatonValue(loc));
+  const currentExpectedProfit =
+    averageAutumnatonValue(currentBestLocation) * expectedRemainingExpeditions();
+
+  const upgradeValuations = acquirableUpgrades.map((upgrade) => {
+    const upgradeLocations = fullLocations.filter(
+      (location) => getAutumnatonUniques(location)[0] === upgrade
+    );
```
horrible-little-slime, line 158: I'm adding the autumnAton uniques to libram and it'll use an object instead of a tuple--we can burn that bridge when we come to it

src/autumnaton.ts
```diff
@@ -0,0 +1,217 @@
...
+    if (upgrade === ("periscope" || "radardish")) {
+      const bestLocWithInstalledUpg = maxBy(fullLocations, (loc: Location) =>
+        averageAutumnatonValue(loc, AutumnAton.visualAcuity() + 1)
+      );
+      const extraExpectedProfit =
+        averageAutumnatonValue(bestLocContainingUpg) +
+        averageAutumnatonValue(bestLocWithInstalledUpg) *
+          Math.max(0, expectedRemainingExpeditions() - 1);
+
+      return { upgrade: upgrade, profit: extraExpectedProfit };
```
horrible-little-slime, line 172:
```suggestion
      return { upgrade, profit: extraExpectedProfit };
```

you can replace all of the `upgrade: upgrade`s with just `upgrade`

src/autumnaton.ts
```diff
@@ -0,0 +1,217 @@
...
+        averageAutumnatonValue(bestLocContainingUpg) +
+        (seasonalItemValue(bestLocWithInstalledUpg) +
+          averageAutumnatonValue(bestLocWithInstalledUpg)) *
+          Math.max(0, expectedRemainingExpeditions() - 1);
+
+      return { upgrade: upgrade, profit: extraExpectedProfit };
+    }
+    return { upgrade: upgrade, profit: 0 };
+  });
+  const mostValuableUpgrade = maxBy(upgradeValuations, ({ profit }) => profit);
```
horrible-little-slime, line 206:
```suggestion
  const mostValuableUpgrade = maxBy(upgradeValuations, "profit");
```

Shiverwarp: Ah I see, maxBy options,  `a key on the objects we're looking at`

Nice

src/autumnaton.ts
```diff
@@ -0,0 +1,224 @@
...
+  "leftarm1",
+  "leftleg1",
+  "rightarm1",
+  "rightleg1",
+  "cowcatcher",
+  "periscope",
+  "radardish",
+] as const;
+
+export function prioritizeUpgradeLocations(fullLocations: Location[]): Location[] {
```
horrible-little-slime, line 142: I think `prioritize` is a misleading name here--it sounds to me like it sorts the list, as opposed to filtering it

Shiverwarp: The best I can think of atm is `trimToBestUpgradeLocations` what do you think?

src/autumnaton.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  const upgradeValuations = acquirableUpgrades.map((upgrade) => {
+    const upgradeLocations = fullLocations.filter(
+      (location) => AutumnAton.getUniques(location)?.upgrade === upgrade
+    );
+    const bestLocContainingUpg = maxBy(upgradeLocations, (loc: Location) =>
+      averageAutumnatonValue(loc)
+    );
+
+    if (upgrade === ("periscope" || "radardish")) {
```
horrible-little-slime, line 135:
```suggestion
    if (["periscope", "radardish"].includes(upgrade)) {
```

The line as is does not do what you want it to do. It will see if upgrade is equal to `"periscope" || "radardish"`, which will evaluate to `"periscope"` (strings are falsy only when empty, and an "or" stops caring once it hits true)

src/autumnaton.ts
```diff
@@ -0,0 +1,189 @@
...
+  const currentExpectedProfit =
+    averageAutumnatonValue(currentBestLocation) * expectedRemainingExpeditions();
+
+  const upgradeValuations = acquirableUpgrades.map((upgrade) => {
+    const upgradeLocations = fullLocations.filter(
+      (location) => AutumnAton.getUniques(location)?.upgrade === upgrade
+    );
+    const bestLocContainingUpg = maxBy(upgradeLocations, (loc: Location) =>
+      averageAutumnatonValue(loc)
+    );
```
horrible-little-slime, line 133:
```suggestion
    const bestLocContainingUpgrade = maxBy(upgradeLocations, averageAutumnatonValue);
```

constructs of the form `(x) => f(x)` are just copying the function `f` in-line. Sometimes that's useful for legibility, but imo it is nicer to read the other way.

horrible-little-slime: Also, letters don't cost us anything; we should use the full word! "upgrade" (that change means that committing this will mean manually fixing some stuff that still references the old name)

src/autumnaton.ts
```diff
@@ -0,0 +1,189 @@
...
+      const bestLocWithInstalledUpg = maxBy(fullLocations, (loc: Location) =>
+        averageAutumnatonValue(loc, AutumnAton.visualAcuity() + 1)
+      );
+      const extraExpectedProfit =
+        averageAutumnatonValue(bestLocContainingUpg) +
+        averageAutumnatonValue(bestLocWithInstalledUpg) *
+          Math.max(0, expectedRemainingExpeditions() - 1);
+
+      return { upgrade, profit: extraExpectedProfit };
+    }
```
horrible-little-slime, line 145: This function feels big and complex for living inside this `map`. Instead of defining it as an arrow function in the map method, make it a regular ol' `function` sitting in the top level of the document (that is, not inside anything else). I haven't read enough of this code to know what to name it off the top of my head

src/autumnaton.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (upgrade === ("rightleg1" || "leftleg1")) {
```
horrible-little-slime, line 146:
```suggestion
    if (upgrade.includes("leg1")) {
```

We can also do the other way. Many array methods also are string methods, because a string can be thought of as an array of unit-length strings

src/autumnaton.ts
```diff
@@ -0,0 +1,189 @@
...
+      const bestLocWithInstalledUpg = currentBestLocation;
+      const extraExpectedProfit =
+        averageAutumnatonValue(bestLocContainingUpg) +
+        averageAutumnatonValue(bestLocWithInstalledUpg) *
+          Math.max(0, expectedRemainingExpeditions(AutumnAton.legs() + 1) - 1);
+
+      return { upgrade, profit: extraExpectedProfit };
+    }
+    if (upgrade === ("rightarm1" || "leftarm1")) {
```
horrible-little-slime, line 155:
```suggestion
    if (upgrade.includes("arm1")) {
```

src/autumnaton.ts
```diff
@@ -0,0 +1,189 @@
...
+      const extraExpectedProfit =
+        averageAutumnatonValue(bestLocContainingUpg) +
+        (seasonalItemValue(bestLocWithInstalledUpg) +
+          averageAutumnatonValue(bestLocWithInstalledUpg)) *
+          Math.max(0, expectedRemainingExpeditions() - 1);
+
+      return { upgrade, profit: extraExpectedProfit };
+    }
+    return { upgrade, profit: 0 };
+  });
```
horrible-little-slime, line 177: Oh that function is even bigger than I thought. We might want to chunk it further once we pull it outside of the nested function.

src/autumnaton.ts
```diff
@@ -0,0 +1,189 @@
...
+  const profitFromBestUpg = mostValuableUpgrade.profit;
+
+  if (profitFromBestUpg > currentExpectedProfit) {
+    const upgradeLocations = fullLocations.filter(
+      (location) => AutumnAton.getUniques(location)?.upgrade === mostValuableUpgrade.upgrade
+    );
+    return upgradeLocations;
+  } else {
+    return fullLocations;
+  }
```
horrible-little-slime, line 188:
```suggestion
  const { upgrade: mostValuableUpgrade, profit: profitFromBestUpgrade }= maxBy(upgradeValuations, "profit");

  if (profitFromBestUpgrade > currentExpectedProfit) {
    const upgradeLocations = fullLocations.filter(
      (location) => AutumnAton.getUniques(location)?.upgrade === mostValuableUpgrade
    );
    return upgradeLocations;
  } else {
    return fullLocations;
  }
```

src/post.ts
```diff
@@ -162,7 +162,7 @@ export default function postCombatActions(skipDiet = false): void {
   updateMallPrices();
   stillsuit();
   funguySpores();
-  if (globalOptions.ascend || AutumnAton.turnsForQuest() < myAdventures() + 10) {
-    AutumnAton.sendTo(autumnAtonZones);
+  if (globalOptions.ascend || AutumnAton.turnsForQuest() < estimatedTurns()) {
+    AutumnAton.sendTo((locations) => maxBy(mostValuableUpgrade(locations), averageAutumnatonValue));
   }
```
horrible-little-slime, line 167: I think instead of defining the thing in-line here using maxBy, mostValuableUpgrade, and averageAutumnAtonValue, `autumnaton.ts` should export a single function that either goes into the `AutumnAton.sendTo` or does the sendTo itself. That is, this should either end up looking like
`AutumnAton.sendTo(bestAutumnatonLocation)` or `sendFallbot()`, depending upon which of those feels more right

src/autumnaton.ts
horrible-little-slime: Also, it looks like you're separately considering several cases using a series of `if` statements; we can either use a `switch` or some sort of iterable to achieve this nicer. That is, you might have something like
```
switch (upgrade) {
  case "periscope":
  case "radardish": {
    all this stuff
  }
  case "rightleg1":
  case "rightleg2": {
    et cetera
  }
  case default:
    return DEFAULT_VALUE;
}
```
or something like
```
for (const { upgrades, profitFunction } of upgradeCases) {
  if (upgrades.includes(upgrade)) return { upgrade, profit: profitFunction() }
}
return DEFAULT_VALUE;
```
with better variable names than the ones I chose, and defining `upgradeCases` outside the scope of this function. We can wait to decide how exactly to set this up until after we've rejiggered the rest of it, though

horrible-little-slime (commented):
I am 85% sure your math is correct but this still has a little bit to go to be as maintainable and legible as possible. I'm sorry it's taking me so long to get through this, but we each have our lives to live.

src/autumnaton.ts
horrible-little-slime: (This'll need linting ofc, I'm freestyling it from here)

src/autumnaton.ts
Shiverwarp: I was originally going to do the switch case but I ran into "no fall-through" lint errors before, and thought that this counted as such.

I will try doing it again

src/post.ts
Shiverwarp: I think a `bestAutumnatonLocation` feels better. Later if we decide it's worth the effort to figure out how to prepare and unlock zones we want to send autumnaton to, I think it will make more sense (And would probably change into two separate functions of `bestAvailableAutumnatonLocation` and `bestUnlockableAutumnatonLocation`

src/autumnaton.ts
Shiverwarp: I've instead pulled the extra profit portions out into functions, let me know if you feel this is sufficient for cleaning this up.

horrible-little-slime (commented):
Later on we should account for the ascending flag, and value upgrades much higher when it's not up, but I think this looks good(?)

Gonna try to get one more person's eyes on it but very, very nice work :)

src/autumnaton.ts
```diff
@@ -0,0 +1,238 @@
...
+      Math.max(
+        1,
+        quests -
+          (legOverride ?? AutumnAton.currentUpgrades().filter((u) => u.includes("leg")).length)
+      );
+    quests++;
+  }
+
+  return quests - get("_autumnatonQuests", 0);
+}
```
horrible-little-slime, line 101:
```suggestion
function expectedRemainingExpeditions(legs = AutumnAton.legs()): number {
  const availableAutumnatonTurns = estimatedTurns() - AutumnAton.turnsLeft();
  const quests = get("_autumnatonQuests");
  const legOffsetFactor = 11 * Math.max(quests - legs - 1, 0);
  return Math.floor(
    Math.sqrt(
      quests ** 2 +
      2 * (
        availableAutumnatonTurns  - legOffsetFactor
      ) / 11
    )
  )
}
```
@docrostov I think this is the right solution for this finite arithmetic series?

Shiverwarp:
> Later on we should account for the ascending flag, and value upgrades much higher when it's not up, but I think this looks good(?)
> 
> Gonna try to get one more person's eyes on it but very, very nice work :)

Yeah, I was planning on doing this in the `expectedRemainingExpeditions` function, but wasn't sure the best way to do it. My first instinct was to just assume double our `availableAutumnatonTurns` (This would naturally matter less as the day went on, but we would want to get our upgrades at the beginning anyway if they're useful, so it should work "fine")

The other option was to just add a flat value if ascending. Neither is particularly elegant though.

src/autumnaton.ts
Shiverwarp: Not related to the math here, but this will still need the override input for number of legs so that the extra upgrade valuations are accurate

horrible-little-slime (dismissed):

PR #1234, opened 2023-01-03
Added support for switching worksheds

Pantocyclus:
Requires the user to correctly set the preference _garboNextWorkshed, which we will attempt to cast it to an item with toItem() (and throw upon failing) if the preference is set


src/diet.ts
```diff
@@ -1005,6 +1005,7 @@ export function consumeDiet(diet: Diet<Note>, name: DietName): void {
   }
 }
 
+export let dietCompleted = globalOptions.noDiet;
```
horrible-little-slime, line 1008: don't export the variable here, export a getter function

src/potions.ts
```diff
@@ -371,6 +371,7 @@ export function doublingPotions(embezzlers: number): Potion[] {
     .map((pair) => pair.potion);
 }
 
+export let potionSetupCompleted = false;
```
horrible-little-slime, line 374: Ditto

src/diet.ts
horrible-little-slime: Alternately, make this part of globalOptions but that doesn't feel ideal

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+  // eslint-disable-next-line libram/verify-constants
+  return getWorkshed() === $item`model train set` && getTrainsetPositionsUntilConfigurable() <= 0;
+}
+
+function setTrainsetConfiguration(pieces: TrainsetPiece[]) {
+  visitUrl("campground.php?action=workshed");
+
+  const pieceIds = pieces
+    .map((p) => getPieceId(p))
+    .map((pieceId, index) => `slot[${index}]=${pieceId}`);
```
horrible-little-slime, line 95:
```suggestion
    .map((piece, index) =>`slot[${index}]=${getPieceId(piece)}`);
```

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+
+  constructor(workshed: Item, done: () => boolean, action: () => void) {
+    this.workshed = workshed;
+    this.done = done;
+    this.action = action;
+  }
+}
+
+export function currentWorkshed(): GarboWorkshed {
+  return worksheds.find((workshed) => workshed.workshed === getWorkshed()) ?? defaultWorkshed;
```
horrible-little-slime, line 127:
```suggestion
  return worksheds.find(({ workshed }) => workshed === getWorkshed()) ?? defaultWorkshed;
```

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+}
+
+export function getGarboWorkshed(item: Item): GarboWorkshed {
+  return worksheds.find((workshed) => workshed.workshed === item) ?? defaultWorkshed;
+}
```
horrible-little-slime, line 132:
```suggestion
export function getGarboWorkshed(item: Item): GarboWorkshed {
  return worksheds.find(({ workshed }) => workshed === item) ?? defaultWorkshed;
}
```

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+  workshed: Item;
+  done: () => boolean;
+  action: () => void;
+
+  constructor(workshed: Item, done: () => boolean, action: () => void) {
+    this.workshed = workshed;
+    this.done = done;
+    this.action = action;
+  }
+}
```
horrible-little-slime, line 124: Right now I don't see any particular reason this should be a Class rather than just a typed object, given that it has no methods

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+export function currentWorkshed(): GarboWorkshed {
+  return worksheds.find((workshed) => workshed.workshed === getWorkshed()) ?? defaultWorkshed;
+}
+
+export function getGarboWorkshed(item: Item): GarboWorkshed {
+  return worksheds.find((workshed) => workshed.workshed === item) ?? defaultWorkshed;
+}
+
+const trainset = new GarboWorkshed(
+  // eslint-disable-next-line libram/verify-constants
```
horrible-little-slime, line 135: Upgrade!

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+      TrainsetPiece.RANDOM_BOOZE,
+      TrainsetPiece.DROP_LAST_FOOD_OR_RANDOM,
+      TrainsetPiece.ORE,
+      TrainsetPiece.EFFECT_MP,
+      TrainsetPiece.GAIN_STATS,
+    ];
+    if (!isTrainsetConfigurable()) return;
+    else if (pieces.join(",") === get("trainsetConfiguration", "")) return;
+    setTrainsetConfiguration(pieces);
+  }
```
horrible-little-slime, line 155: You can declare this out of scope, but later on we should do math to do the 42* turns trick


*may not end up being 42

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+
+const pizzaCube = new GarboWorkshed(
+  $item`diabolic pizza cube`,
+  () => {
+    return dietCompleted;
+  },
+  () => {
+    return;
+  }
+);
```
horrible-little-slime, line 199: I'd leave this out until we actually support pizza in diet. It's not complicated enough code that it's worth commenting anywhere.

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+    const currentFuel = getFuel();
+    const requiredFuel = Math.max(Math.ceil(requiredTurns / 30) * 37 - currentFuel, 0);
+    if (requiredFuel > 10000) {
+      throw new Error(`Wanted to fill ${requiredFuel} fuel in our Asdon, probably a bug!`);
+    }
+    fillTo(requiredFuel);
+    for (let x = 0; x < Math.ceil(requiredTurns / 30); x++) {
+      cliExecute("asdonmartin drive observantly");
+    }
+  }
```
horrible-little-slime, line 218: just use `AsdonMartin.drive`

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+  },
+  () => {
+    return;
+  }
+);
+
+const asdonMartin = new GarboWorkshed(
+  $item`Asdon Martin keyfob`,
+  () => {
+    return haveEffect($effect`Driving Observantly`) >= estimatedTurns();
```
horrible-little-slime, line 204: We will want to deliberately overbuff for at least 1 extra day unless you're ascending

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+
+  if (expected !== get("trainsetConfiguration", "")) {
+    throw new Error(
+      `Expected trainset configuration to have changed, expected "${expected}" but instead got ${get(
+        "trainsetConfiguration",
+        ""
+      )}`
+    );
+  }
+}
```
horrible-little-slime, line 112: All of this belongs in either libram or its own file here. But probably libram.

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+
+    const bestItem = maxBy([...itemChoices.keys()], garboValue);
+    const bestChoice = itemChoices.get(bestItem);
+    if (bestChoice && bestChoice > 0) {
+      visitUrl("campground.php?action=workshed");
+      runChoice(bestChoice);
+    }
+    if (handlingChoice()) visitUrl("main.php");
+  }
+);
```
horrible-little-slime, line 189: define this as a `function` function before `const CMC`, and then just plop the function in there.

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+const snowMachine = new GarboWorkshed(
+  $item`snow machine`,
+  () => {
+    // Assume garden has already been harvested(?)
+    return true;
+  },
+  () => {
+    return;
+  }
+);
```
horrible-little-slime, line 256: This and all of the warbear ones that suck feel like they can just be defined nicely with a `.map`

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+      TrainsetPiece.EFFECT_MP,
+      TrainsetPiece.GAIN_STATS,
+    ];
+    if (!isTrainsetConfigurable()) return;
+    else if (pieces.join(",") === get("trainsetConfiguration", "")) return;
+    setTrainsetConfiguration(pieces);
+  }
+);
+
+const CMC = new GarboWorkshed(
```
pstalcup, line 158: all caps variable imply something that this is not (that is to say, a class or constant variable)

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+const asdonMartin = new GarboWorkshed(
+  $item`Asdon Martin keyfob`,
+  () => {
+    return haveEffect($effect`Driving Observantly`) >= estimatedTurns();
+  },
+  () => {
+    const currentTurns = haveEffect($effect`Driving Observantly`);
+    const requiredTurns = estimatedTurns() - currentTurns;
+    const currentFuel = getFuel();
+    const requiredFuel = Math.max(Math.ceil(requiredTurns / 30) * 37 - currentFuel, 0);
```
pstalcup, line 210: If a function does not exist in libram to handle this yet, we should move some of the logic into libram

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+  $item`warbear induction oven`,
+  () => {
+    return dietCompleted;
+  },
+  () => {
+    return;
+  }
+);
+
+const drillPress = new GarboWorkshed(
```
pstalcup, line 311: For all of these worksheds that do nothing, it would be better to have a function generate them:

```ts
const drillPress = noopWorkshed($item`warbear jackhammer drill press`);
```

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+  () => {
+    // Garbo does not support smashing equipment
+    return true;
+  },
+  () => {
+    return;
+  }
+);
+
+const ROMBurner = new GarboWorkshed(
```
pstalcup, line 322: Same,

```ts
const romBurner = potionWorkshed($item`warbear LP-ROM burner`)
```

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+  $item`none`,
+  () => {
+    return true;
+  },
+  () => {
+    return;
+  }
+);
+
+const worksheds = [
```
pstalcup, line 342: Given that we don't use these intermediate variables for anything, I'm not sure it makes sense to have them exist in that format rather than just being pure members of this array

src/workshed.ts
pstalcup: To be specific, he means changing the `>=` to a `>` unless ascending is true

src/workshed.ts
```diff
@@ -0,0 +1,390 @@
...
+  }
+);
+
+const CMC = new GarboWorkshed(
+  $item`cold medicine cabinet`,
+  () => {
+    return get("_coldMedicineConsults") >= 5;
+  },
+  () => {
+    if (property.getNumber("_nextColdMedicineConsult") > totalTurnsPlayed()) return;
```
pstalcup, line 164: This is unrelated to your PR except you are using worksheds - It would be good to update this code to not hit the CMC every turn and instead use the mafia properties to check

pstalcup (changes requested):

src/workshed.ts
horrible-little-slime: That's not actually quite what I mean!

I mean that if someone's in long aftercore, the best behavior for asdon isn't going to be "buff up until the end of the day, then switch to cmc", it's going to be "buff up for a week, then switch between cmc and trainrealm until you need asdon again". And even if you aren't in long aftercore, on your non-ascending leg you'll want enough asdon turns for that day and the next.

It's not obvious what best behavior is, but I'd say at least two days of turns if you aren't ascending, maybe a pref or an argument for doing more.

Pantocyclus: What would be the `estimatedTurns()` for the extra day? Something like `400 + clamp((VOA-4000)/10, 0, 300)`?

src/workshed.ts
Pantocyclus: I used default methods instead, which I think looks nicer?

src/workshed.ts
Pantocyclus: Same as above

src/workshed.ts
Pantocyclus: Used `AsdonMartin.drive()` as mentioned above

src/workshed.ts
Pantocyclus: Moved workshed into a new folder with its own util file

src/workshed.ts
Pantocyclus: Cleaned things up by using default functions, which interfaces do not support

src/workshed.ts
Pantocyclus: I think this is out of scope for the current PR (I basically copied the existing `coldMedicineCabinet()` code).
The current code hits the CMC to check what items are available, then selects the best amongst them (since we don't want to choose an item we cannot currently select). 
I guess we can handle it similarly by parsing the most common character in `lastCombatEnvironment` and mapping it to its corresponding items to generate the same list of available items (only if "?" is not present in the preference), but this also seems like it would be more useful as a libram thing than a garbo thing as well (much like the trainset)

src/workshed.ts
Pantocyclus: It's just "reset configuration to the default configuration if it is allowed and `turnsToMeat >= 2`" right?

src/workshed.ts
Pantocyclus: It needs to be in so that we can swap out of it if a user starts with pizzaCube at the start of day

src/workshed/index.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  done(): boolean {
+    return true;
+  }
+
+  action(): void {
+    return;
+  }
+}
```
horrible-little-slime, line 39:
```suggestion
  workshed: Item;
  done = () => true;
  action = () => { return };

  constructor(workshed: Item, done?: () => boolean, action?: () => void) {
    this.workshed = workshed;
    if (done) this.done = done;
    if (action) this.action = action;
  }
}
```

src/workshed/index.ts
```diff
@@ -0,0 +1,157 @@
...
+      setTrainsetConfiguration(defaultPieces);
+    }
+  ),
+  new GarboWorkshed(
+    $item`cold medicine cabinet`,
+    () => {
+      return get("_coldMedicineConsults") >= 5;
+    },
+    () => {
+      if (property.getNumber("_nextColdMedicineConsult") > totalTurnsPlayed()) return;
```
horrible-little-slime, line 70:
```suggestion
      if (get("_nextColdMedicineConsult") > totalTurnsPlayed()) return;
```

src/workshed/index.ts
```diff
@@ -0,0 +1,157 @@
...
+      // Since we don't actually support using the syringe in combat at this time, the counter will never change
+      return get("_dnaPotionsMade") >= 3;
+    },
+    () => {
+      // Just grab whatever tonics for now, since we don't actually have support for DNA
+      const page = visitUrl("campground.php?action=workshed");
+      if (page.includes("Human-")) {
+        visitUrl("campground.php?action=dnapotion");
+      }
+    }
```
horrible-little-slime, line 104: we should use libram's DNALab for this

yarn.lock
```diff
@@ -3318,10 +3318,10 @@ regenerate@^1.4.2:
   resolved "https://registry.yarnpkg.com/regenerate/-/regenerate-1.4.2.tgz#b9346d8827e8f5a32f7ba29637d398b69014848a"
   integrity sha512-zrceR/XhGYU/d/opr2EKO7aRHUeiBI8qjtfHqADTwZd6Szfy16la6kqD0MIUs5z5hx6AaKa+PixpPrR289+I0A==
 
-regenerator-runtime@^0.13.4:
-  version "0.13.9"
-  resolved "https://registry.yarnpkg.com/regenerator-runtime/-/regenerator-runtime-0.13.9.tgz#8925742a98ffd90814988d7566ad30ca3b263b52"
-  integrity sha512-p3VT+cOEgxFsRRA9X4lkI1E+k2/CtnKtU4gcxyaCUreilL/vqI6CdZ3wxVUx3UOUg+gnUOQQcRI7BmSI656MYA==
+regenerator-runtime@^0.13.11, regenerator-runtime@^0.13.4:
```
horrible-little-slime, line 3321: What's going on over here

src/workshed/index.ts
```diff
@@ -57,7 +57,12 @@ const worksheds = [
     },
     () => {
       if (!isTrainsetConfigurable()) return;
-      else if (defaultPieces.join(",") === get("trainsetConfiguration", "")) return;
+      else {
+        const pos = get("trainsetPosition", 0);
+        const configured = get("lastTrainsetConfiguration", 0);
+        const idx = (pos - configured) % 8;
+        if (defaultPieces.join(",") === get("trainsetConfiguration", "") && idx <= 1) return;
+      }
```
horrible-little-slime, line 60: eventually we'll want to not hardcode this, but we'll burn that bridge when we come to it

yarn.lock
Pantocyclus: idk tbh. I'll manually fix it I guess

src/workshed/index.ts
```diff
@@ -0,0 +1,154 @@
...
+class GarboWorkshed {
+  workshed: Item;
+  done = () => true;
+  action = () => {
+    return;
+  };
+  constructor(workshed: Item, done?: () => boolean, action?: () => void) {
+    this.workshed = workshed;
+    if (done) this.done = done;
+    if (action) this.action = action;
```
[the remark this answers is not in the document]
Pantocyclus: I don't think this works. `done` could be undefined; but also, `this.done` is already defined so `done` wouldn't be assigned to it.

src/workshed/index.ts
```diff
@@ -0,0 +1,171 @@
...
+        return setTrainsetConfiguration(defaultPieces);
+      } else {
+        const pieces = getTrainsetConfiguration();
+        const offset = toInt(get("trainsetPosition")) % 8;
+        const nextPiece = pieces[offset];
+
+        print(`Reconfiguring trainset, as our next station is ${String(nextPiece)}`, "blue");
+        if ([TrainsetPiece.DOUBLE_NEXT_STATION, TrainsetPiece.GAIN_MEAT].includes(nextPiece)) {
+          return;
+        }
```
horrible-little-slime, line 75:
```suggestion
        if ([TrainsetPiece.DOUBLE_NEXT_STATION, TrainsetPiece.GAIN_MEAT].includes(nextPiece)) {
          return;
        }
       print(`Reconfiguring trainset, as our next station is ${String(nextPiece)}`, "blue");
```

src/workshed/index.ts
```diff
@@ -0,0 +1,171 @@
...
+
+        const newPieces: TrainsetPiece[] = [];
+        for (let i = 0; i < 8; i++) {
+          const newPos = (i + offset) % 8;
+          newPieces[newPos] = defaultPieces[i];
+        }
```
horrible-little-slime, line 81: SLAP this off into a separate function

horrible-little-slime (commented):
I think probably it will be better to make `post` its own module and make `workshed` one page from it, but we can address that later probably

src/workshed/index.ts
```diff
@@ -0,0 +1,165 @@
...
+
+function currentWorkshed(): GarboWorkshed {
+  return getGarboWorkshed(getWorkshed());
+}
+
+let _nextWorkshed: GarboWorkshed | null = null;
+function nextWorkshed(): GarboWorkshed {
+  _nextWorkshed ??= getGarboWorkshed(stringToWorkshedItem());
+  return _nextWorkshed ?? defaultWorkshed;
+}
```
horrible-little-slime, line 52: Make these all statics on the GarboWorkshed class

src/workshed/index.ts
```diff
@@ -0,0 +1,165 @@
...
+) {
+  print(
+    `Warning: We currently do not support switching from the model train set to another workshed, so ${
+      nextWorkshed().workshed
+    } will not be set-up during this run of garbo!`,
+    "red"
+  );
+}
+
+export function handleWorkshed(): void {
```
horrible-little-slime, line 154:
```suggestion
export default function handleWorkshed(): void {
```

src/workshed/index.ts
```diff
@@ -0,0 +1,165 @@
...
+  new GarboWorkshed(
+    $item`spinning wheel`,
+    () => {
+      return get("_spinningWheel");
+    },
+    () => {
+      // We simply assume you will not gain a level while garboing, since we do not do powerlevellings
+      // So we will just use the spinning wheel immediately
+      visitUrl("campground.php?action=spinningwheel");
+    }
```
horrible-little-slime, line 129: If this isn't in daily, put it in daily

src/workshed/utils.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  const bestItem = maxBy([...itemChoices.keys()], garboValue);
+  const bestChoice = itemChoices.get(bestItem);
+  if (bestChoice && bestChoice > 0) {
+    visitUrl("campground.php?action=workshed");
+    runChoice(bestChoice);
+  }
+  if (handlingChoice()) visitUrl("main.php");
+}
```
horrible-little-slime, line 136: Move this to appear after all the trainset functions, instead of betwixt

src/config.ts
```diff
@@ -36,6 +36,11 @@ You can use multiple options in conjunction, e.g. "garbo nobarf ascend"',
       help: "*EXPERIMENTAL* garbo will sacrifice some optimal behaviors to run quicker. Estimated and actual profits may be less accurate in this mode.",
       default: false,
     }),
+    workshed: Args.string({
+      setting: "",
+      help: "workshed that garbo should intelligently switch to after fully exploiting the benefits of your current workshed.",
+      default: "",
+    }),
```
horrible-little-slime, line 43: Make this `Args.custom`, have it convert from a string to an item _here_, that way `help` can explicitly list the aliases we support

src/workshed/utils.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export function stringToWorkshedItem(): Item {
+  // An empty string is a subset of every string and will match all the worksheds
+  // So we explicitly handle this case here
+  if (globalOptions.workshed === "") return $item`none`;
+
+  const lowerCaseWorkshed = globalOptions.workshed.toLowerCase();
+  const validWorksheds = [
+    ...new Map([
```
horrible-little-slime, line 158: I don't think there's any utility in wrapping this chunk in the map constructor and then calling entries, we never use the mappiness of it.

horrible-little-slime (changes requested):

src/workshed/index.ts
Pantocyclus: I don't really see a need though - we would still need this here if low-end users are swapping into the spinning wheel

src/config.ts
Pantocyclus: I sure hope this doesn't result in a circular import somewhere down the line...

src/config.ts
```diff
@@ -1,5 +1,7 @@
 import { Args } from "grimoire-kolmafia";
-import { get } from "libram";
+import { Item } from "kolmafia";
+import { $item, get } from "libram";
+import { stringToWorkshedItem } from "./workshed/utils";
```
horrible-little-slime, line 4: To prevent circular imports, just define this in `config` instead of `workshed`

src/config.ts
```diff
@@ -36,6 +38,14 @@ You can use multiple options in conjunction, e.g. "garbo nobarf ascend"',
       help: "*EXPERIMENTAL* garbo will sacrifice some optimal behaviors to run quicker. Estimated and actual profits may be less accurate in this mode.",
       default: false,
     }),
+    workshed: Args.custom<Item>(
+      {
+        default: $item`none`,
+        help: "Parses the substring of the item name (e.g dna, mayo, asdon) into the corresponding workshed for garbo to intelligently switch into. Additional accepted aliases: cmc, trainset.",
```
horrible-little-slime, line 44: Instead of explicitly listing acceptable aliases, consider the `options` parameter

src/config.ts
Pantocyclus: It really doesn't look like it belongs in config because of how specific it is to the worksheds

src/config.ts
Pantocyclus: Setting options makes it appear as though cmc and trainset are the only 2 options for this argument, unless we want to explicitly list everything

Pantocyclus: image
This is probably the best I can do

mafia/data/garbo_item_lists.json
```diff
@@ -230,5 +230,53 @@
     "Angry Farmer's Wife Candy",
     "cotton candy pillow",
     "sugar shillelagh"
+  ],
+  "trainset": [
```
docrostov, line 234: are we -positive- these are the only candies from the trainset? i can't recall getting anything else but i don't remember where we sourced this stuff

src/config.ts
```diff
@@ -1,5 +1,31 @@
 import { Args } from "grimoire-kolmafia";
-import { get } from "libram";
+import { Item } from "kolmafia";
+import { $item, $items, get } from "libram";
+
+const workshedAliases = [
+  { item: $item`cold medicine cabinet`, aliases: ["cmc"] },
+  { item: $item`model train set`, aliases: ["trainset", "trainrealm", "train"] },
+  { item: $item`Asdon Martin keyfob`, aliases: ["breadcar", "car", "asdon", "aston"] },
```
docrostov, line 8: aston kutcher eat your hart out

src/workshed/index.ts
```diff
@@ -0,0 +1,143 @@
...
+import { globalOptions } from "../config";
+import { potionSetupCompleted } from "../potions";
+import { estimatedTurns } from "../turns";
+import { getBestCycle, grabMedicine, offsetDefaultPieces } from "./utils";
+
+let _nextWorkshed: GarboWorkshed | null = null;
+let _currentWorkshed: GarboWorkshed | null = null;
+class GarboWorkshed {
+  workshed: Item;
+  done = () => true;
```
docrostov, line 13: it took a little bit to wrap my head around the boolean logic being used for the "done" guy here tbh. it's fine but it did take a bit of thought and isn't super intuitive imho. i'm not sure GarboWorkshed as a class is great at introducing users on how to make GarboWorkshed entries for new ones, which doesn't matter now but might be a little annoying cruft down the road.

src/workshed/index.ts
```diff
@@ -0,0 +1,143 @@
...
+    return _currentWorkshed ?? defaultWorkshed;
+  }
+
+  static nextWorkshed(): GarboWorkshed {
+    _nextWorkshed ??= GarboWorkshed.getGarboWorkshed(globalOptions.workshed);
+    return _nextWorkshed ?? defaultWorkshed;
+  }
+}
+const defaultWorkshed = new GarboWorkshed($item`none`);
+const estimatedTurnsTomorrow = 400 + clamp((get("valueOfAdventure") - 4000) / 8, 0, 600);
```
docrostov, line 38: where... where did this number come from? i did some quick math and i think it "works" but i do fear this is a little bit of a "this works now but won't work later" type thing as different turngen sources drop or get nerfed. any appetite to make something like a "maxGarboTurnsPlayed" preference that garbo updates if you spent more turns in a given garbo session than the preference? that way it isn't VoA based but usage based, and using the highest # of turns the user has ever had. idk. just a thought.

docrostov (commented):
i don't feel comfortable approving this because the GarboWorkshed paradigm is a little too finicky for me to feel at home in the code. but i have a handful of comments, some jokes and some potentially useful observations/suggestions. ignore them at your peril. (this is a joke it's very reasonable to ignore these)

src/config.ts
docrostov: for what it's worth (not much!), i agree w/ phred, this feels like a config thing despite being very shed-specific.

mafia/data/garbo_item_lists.json
horrible-little-slime: I got this list from Nas, he posted it December 29 from having run it all season. This isn't set in stone, though, we can edit her when relevant

src/workshed/index.ts
horrible-little-slime: This one is indeed magic numbers, but the `clamp` makes this harder to figure out. I would like to have a fuller discussion about this in the LASS channel but I'm yet to see a better pitch

src/workshed/index.ts
horrible-little-slime: I should make the constructor take an options object to make it easier to read, and maybe use a more informative name than done

src/workshed/index.ts
```diff
@@ -0,0 +1,164 @@
...
+      if (!get("trainsetConfiguration")) {
+        // Visit the workshed to make sure it's actually empty, instead of us having not yet seen it this run
+        visitUrl("campground.php?action=workshed");
+        visitUrl("main.php");
+      }
+      if (!get("trainsetConfiguration")) {
+        print("Reconfiguring trainset, as our next station is empty", "blue");
+        return TrainSet.setConfiguration(getBestCycle());
+      } else {
+        const bestTwoStations = getBestCycle().splice(0, 2);
```
Pantocyclus, line 82: I saw this earlier and was thinking - if we do in fact find something that beats the meat station, is it optimal to do a 42-cycle instead of a 43-cycle? I'm sure we can maths this out, but we can also leave that for a future PR

src/config.ts
```diff
@@ -1,5 +1,31 @@
...
+  { item: $item`diabolic pizza cube`, aliases: ["cube", "pizza", "pizzacube"] },
+  { item: $item`portable Mayo Clinic`, aliases: ["mayo", "clinic"] },
+  { item: $item`Little Geneticist DNA-Splicing Lab`, aliases: ["dna", "lab", "dnalab"] },
```
Pantocyclus, line 11: Like Katarn has also mentioned, I don't really like `lab` since the chemistry lab is also a lab. This also complicates matters if we do intend to reintroduce substring matching in the future, where `lab` (which may be a working implementation in some user's scripts) would then cause issues with the reimplementation of that feature.

src/workshed/index.ts
```diff
@@ -0,0 +1,164 @@
...
+    done: () => {
+      // This will likely always return true or false for now, depending on the start state of garbo
+      // Since we don't actually support using the syringe in combat at this time, the counter will never change
+      return get("_dnaPotionsMade") >= 3;
+    },
+    action: () => {
+      // Just grab whatever tonics for now, since we don't actually have support for DNA
+      if (get("dnaSyringe")) DNALab.makeTonic(3);
+    },
+  }),
```
Pantocyclus, line 124: I was also thinking of setting a global `let _triedMakingTonics = false`, with `done: () => { return _triedMakingTonics || get("_dnaPotionsMade") >=3; }` and `action: () => { _triedMakingTonics = true; if (get("dnaSyringe")) DNALab.makeTonic(3); }`, which should be much better behaviour that respects the user's intent to swap out of DNA

horrible-little-slime: I'm going to declare that out of scope and a good fit for future generations

PR #1236, opened 2023-01-05
Update guzzlr.ts

AmirhPooya:
Guzzlr prioritize for bronze when bronze is not capped

### DIFF
--- a/src/wanderer/guzzlr.ts
+++ b/src/wanderer/guzzlr.ts
@@ -40,12 +40,11 @@ function acceptGuzzlrQuest(locationSkiplist: Location[]) {
       !(get("garbo_prioritizeCappingGuzzlr", false) && Guzzlr.haveFullPlatinumBonus())
     ) {
       Guzzlr.acceptPlatinum();
-    } else if (Guzzlr.canGold() && (Guzzlr.haveFullBronzeBonus() || !Guzzlr.haveFullGoldBonus())) {
-      // if gold is not maxed, do that first since they are limited per day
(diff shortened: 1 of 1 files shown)


src/wanderer/guzzlr.ts
```diff
@@ -40,12 +40,11 @@ function acceptGuzzlrQuest(locationSkiplist: Location[]) {
       !(get("garbo_prioritizeCappingGuzzlr", false) && Guzzlr.haveFullPlatinumBonus())
     ) {
       Guzzlr.acceptPlatinum();
-    } else if (Guzzlr.canGold() && (Guzzlr.haveFullBronzeBonus() || !Guzzlr.haveFullGoldBonus())) {
-      // if gold is not maxed, do that first since they are limited per day
-      Guzzlr.acceptGold();
-    } else {
-      // fall back to bronze when can't plat, can't gold, or bronze is not maxed
+    } else if (!Guzzlr.haveFullBronzeBonus() || !Guzzlr.canGold()) {
```
Pantocyclus, line 43: As per our discussion in discord, this may not be optimal wrt capping the guzzlr progress asap.
I'd do something like `else if ((!Guzzlr.haveFullBronzeBonus() && get("garboGuzzlrPrioritizeBronze", false)) || !Guzzlr.canGold())` where users can choose to set that preference if they want to max out their bronze bonus asap

horrible-little-slime:
Closing for the reasons panto outlined, but I am very open to changing this behavior if we get a better idea of what that change should be

PR #1239, opened 2023-01-06
Switch to Grimoire args

s-k-z:


horrible-little-slime:
Oh, wait, we'll want to update readme.md as well

PR #1250, opened 2023-01-13
Enable usage of porcelain candy dishes

MrFizzyBubbs:
Discussion for this functionality roughly started here: https://discord.com/channels/466605739838930955/756649296774037567/1062567486731603998.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -40,7 +40,7 @@ You can use multiple options in conjunction, e.g. "garbo nobarf ascend"',
       setting: "",
       help: "Print the current version and exit.",
     }),
-    richbitch: Args.flag({
+    candydish: Args.flag({
       setting: "",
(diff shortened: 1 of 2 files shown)


MrFizzyBubbs:
Alternatively we could only use dishes if you have >=100 in inventory.

src/config.ts
```diff
@@ -40,6 +40,12 @@ You can use multiple options in conjunction, e.g. "garbo nobarf ascend"',
       setting: "",
       help: "Print the current version and exit.",
     }),
+    richbitch: Args.flag({
```
Pantocyclus, line 43: As Katarn has mentioned in the discord, we should change this to something that isn't as explicit as this. Perhaps `candydish` or `redtooth` or something else descriptive that isn't too lengthy.

MrFizzyBubbs: :(

Pantocyclus (dismissed):
I am ok with this, but will leave it to someone else to merge it in if they think it's fine

horrible-little-slime:
can we add a pref for this? It seems like something some people would want to leave on indefinitely

PR #1251, opened 2023-01-13
Fight moles

horrible-little-slime:


Rinn (approved):

horrible-little-slime:
@s-k-z has voiced some interesting ideas about how to iterate on this in the future, but I don't see any harm in merging as-is

PR #1254, opened 2023-01-14
More workshed features

Pantocyclus:
Added substring matching and initials matching to config. 

Also supports switching out of DNA lab, as well as trainset -> CMC (and should in general switch to whatever workshed the user intended to after we run out of turns)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -42,6 +42,8 @@ function stringToWorkshedItem(s: string): Item | null {
       )
   );
 
+  // grimoire catches the errors and throws its own errors
+  // so throw new Error(text) would result in the text not getting printed.
   if (validWorksheds.length > 1) {
     print(`Invalid Workshed: ${s} matches multiple worksheds! Matched:`, "red");
     validWorksheds.forEach(({ item }) => print(`${item}`, "red"));
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,16 +1,29 @@
 import { Args } from "grimoire-kolmafia";
-import { Item } from "kolmafia";
+import { Item, print } from "kolmafia";
 import { $item, $items, get } from "libram";
 
 const workshedAliases = [
-  { item: $item`cold medicine cabinet`, aliases: ["cmc"] },
-  { item: $item`model train set`, aliases: ["trainset", "trainrealm", "train"] },
-  { item: $item`Asdon Martin keyfob`, aliases: ["breadcar", "car", "asdon", "aston"] },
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -1,30 +1,57 @@
...
+      )
+  );
+
+  if (validWorksheds.length > 1) {
+    print(`Invalid Workshed: ${s} matches multiple worksheds! Matched:`);
+    validWorksheds.forEach(({ item }) => print(`${item}`));
+    throw new Error();
+  } else if (validWorksheds.length === 0) {
+    print(`Invalid Workshed: ${s} does not match any worksheds!`);
+    throw new Error();
```
Pantocyclus, line 51: Pre-empting that grimoire actually catches the errors and throws its own errors, so `throw new Error(text)` would result in the text not getting printed.

horrible-little-slime: Just make this a code comment

src/workshed/index.ts
```diff
@@ -17,14 +18,20 @@ class GarboWorkshed {
...
     if (options.done) this.done = options.done;
     if (options.action) this.action = options.action;
+    this.minTurns = options.minTurns ?? 0;
   }
 
   canRemove(): boolean {
-    return this.done?.() ?? true;
+    return (
+      (this.done?.() ?? true) ||
+      estimatedTurns() - globalOptions.saveTurns <=
```
horrible-little-slime, line 32: estimatedTurns() _should_ already be accounting for saveTurns

src/workshed/index.ts
```diff
@@ -17,14 +18,20 @@ class GarboWorkshed {
...
+        (GarboWorkshed.next ? GarboWorkshed.next.minTurns ?? 0 : 0)
```
horrible-little-slime, line 33:
```suggestion
        (GarboWorkshed.next?.minTurns ?? 0)
```

src/config.ts
```diff
@@ -1,30 +1,63 @@
...
+    .map((term) => term[0])
+    .join("");
+  return initials.length >= 3 ? initials : "";
+}
+
+function stripString(s: string): string {
+  if (s.includes(" ")) return stripString(s.replace(" ", ""));
+  else if (s.includes("-")) return stripString(s.replace("-", ""));
+  return s;
+}
```
horrible-little-slime, line 32: Instead of this bespoke function, use either replaceAll (may need a polyfill?) or a regexp with the `g` flag in `replace`

horrible-little-slime: I will probably do something about this Some Day

horrible-little-slime (approved):

PR #1255, opened 2023-01-14
Use lodestone in potionSetup before embezzlers/yachtzees

Pantocyclus:


horrible-little-slime:
I don't see any other once/day buffs in this area, which is inconsistent. Is this the only meat/famweight daily buff, should we move this into mood, or should we move mood into this?

Pantocyclus:
Doubling potions with pillkeeper is once-a-day too

Pantocyclus:
Oh hmm... each turn of this meat buff is worth 1k/2k for embezzlers/yachtzees, but 3.5k EV for volcano. 
Since we do volcano before yachtzees before embezzlers I suppose we should actually just grab this at the start of day, and let garbo naturally use it on the most optimal task

horrible-little-slime:
> Oh hmm... each turn of this meat buff is worth 1k/2k for embezzlers/yachtzees, but 3.5k EV for volcano. Since we do volcano before yachtzees before embezzlers I suppose we should actually just grab this at the start of day, and let garbo naturally use it on the most optimal task

make it so, and i'll get it on in

src/tasks/dailyItems.ts
```diff
@@ -289,5 +290,17 @@ export const DailyItemTasks: Task[] = [
...
+      ready: () => have($item`lodestone`) && !get("_lodestoneUsed"),
+      completed: () => get("_lodestoneUsed"),
+      do: () => use($item`lodestone`),
+    },
+    {
+      name: "Update Garbage Tote",
+      ready: () => have($item`January's Garbage Tote`) && !get("_garbageItemChanged"),
+      completed: () => get("_garbageItemChanged"),
+      do: () => cliExecute("fold broken champagne bottle"),
+    },
```
Pantocyclus, line 304: There are 2 solutions to this. Either we update the garbage tote state here at the start of day if it's not updated, or we could check for `!get("_garbageItemChanged") || get("garbageChampagneCharge") > 0` for the drum machines and `!get("_garbageItemChanged") || get("garbageShirtCharge") > 0` for NEP.

I'm more inclined towards this solution since this means we don't have to remember to add the condition to every piece of code that uses the tote.

PR #1256, opened 2023-01-14
Hippoking return arg

horrible-little-slime:


src/session.ts
```diff
@@ -216,7 +216,7 @@ export function valueSession(): void {
   Session.current().toFile("test.json");
 }
 
-export function printGarboSession(): void {
+export function endSession(printLog = true): void {
```
s-k-z, line 219: When do you pass in false to `endSession`?

src/config.ts
```diff
@@ -65,24 +65,24 @@ export const globalOptions = Args.create(
...
       setting: "",
-      help: "do beginning of the day setup, embezzlers, and various daily flags, but will terminate before normal Barf Mountain turns. May close NEP for the day.",
+      help: "operate under the assumption that you're ascending after running it, rather than experiencing rollover. It will use borrowed time, it won't charge stinky cheese items, etc.",
       default: false,
     }),
-    ascend: Args.flag({
+    dumpstash: Args.flag({
       setting: "",
-      help: "operate under the assumption that you're ascending after running it, rather than experiencing rollover. It will use borrowed time, it won't charge stinky cheese items, etc.",
+      help: "return all items to the clan stash, then quit",
```
s-k-z, line 75: `dumpstash` sounds ambiguous, why not `returnstash`?
Which clan stash? I think help should clarify.

src/index.ts
```diff
@@ -118,6 +123,56 @@ export function main(argString = ""): void {
     );
   }
 
+  if (stashItems.length > 0) {
+    if (
+      userConfirmDialog(
+        `Garbo has detected that you have the following items still out of the stash from a previous run of garbo: ${stashItems
+          .map((item) => item.name)
+          .join(",")}. Would you like us to return these to the stash now?`,
```
s-k-z, line 131: Could use a space here

src/index.ts
```diff
@@ -118,6 +123,56 @@ export function main(argString = ""): void {
...
+                stashItems.splice(stashItems.indexOf(item), 1);
+              }
+            }
+          });
+        } else throw new Error("Error: No garbo_stashClan set.");
+      } finally {
+        endSession(false);
+      }
+    } else {
+      stashItems.splice(0, stashItems.length);
```
s-k-z, line 170: I think there's a degenerate case here.

Let's say that you end the day without returning items to the stash.
The next day, you run the script and are prompted to return the items.
You select no.
The `stashItems` array is then emptied.
At the end of this session, the `garboStashItems` property is updated to be the `stashItems` array, which should empty it as well?
If you want to then return items to the clan stash, you have to do so manually, don't you?

s-k-z (approved):

PR #1258, opened 2023-01-15
feat: if quick mode enabled, don't stasis for as long

midgleyc:
Stasising for 20 rounds can drastically increase how long garbo takes to run for incremental gains. In quick mode, don't.

Some of the causes for 20 round stasis are questionable -- I think BittyCar only happens after the first round, and many crown / bjorn drops only in the first 3 rounds -- but I left them as they were.


horrible-little-slime:
The talk page on the wiki suggests that the regular wiki page is incorrect; @taltamir said
> I was told by someone that it can actually activate on other rounds other than the first. I bought one and tested it (in 2019) and it does indeed activate after the first round. I poked an enemy with a seal tooth until it finally activated on the 7th round. Is it possible that this was changed from 1st round only to just going on until it activates?

(This was on April 1st, but I think it is exceedingly unlikely it is a joke)

For the bjorn/crown, nuggets do seem to happen well after round 3. From today's log
> Round 7: BTGGF makes some spooky nuggets for you. What exactly are spooky nuggets? Nobody knows. That's what's so spooky about them.
Although if we find that specifically the elemental fairies can trigger late, then we can narrow down that filter. 

This seems like a good idea regardless, though.

PR #1259, opened 2023-01-15
First draft of Wart Support

horrible-little-slime:


docrostov (approved):
hell yeah

docrostov:
(i know this is a deeply un-detailed review but the code makes sense to me and the logic matches my issue on it. lookin good bud.)

PR #1269, opened 2023-01-24
Change 42 turn trick to 4x turn trick

horrible-little-slime:
This is untested at present

### DIFF
--- a/src/post/workshed.ts
+++ b/src/post/workshed.ts
@@ -4,7 +4,7 @@ import { dietCompleted } from "../diet";
 import { globalOptions } from "../config";
 import { potionSetupCompleted } from "../potions";
 import { estimatedTurns } from "../turns";
-import { getBestCycle, grabMedicine, offsetDefaultPieces } from "./workshed_utils";
+import { getBestCycle, getBestOffset, grabMedicine, offsetDefaultPieces } from "./workshed_utils";
 type WorkshedOptions = {
   workshed: Item;
   done?: () => boolean;
@@ -83,10 +83,11 @@ const worksheds = [
         print("Reconfiguring trainset, as our next station is empty", "blue");
         return TrainSet.setConfiguration(getBestCycle());
       } else {
-        const bestTwoStations = getBestCycle().splice(0, 2);
+        const offsetWanted = getBestOffset();
+        const stationToWatchFor = getBestCycle()[offsetWanted];
         const offset = get("trainsetPosition") % 8;
-        if (bestTwoStations.includes(TrainSet.next())) return;
-        print(`Reconfiguring trainset, as our next station is ${TrainSet.next()}`, "blue");
+        if (TrainSet.next() !== stationToWatchFor) return;
+        print(`Reconfiguring trainset.}`, "blue");
         return TrainSet.setConfiguration(offsetDefaultPieces(offset));
       }
     },
--- a/src/post/workshed.ts
+++ b/src/post/workshed.ts
@@ -4,7 +4,12 @@ import { dietCompleted } from "../diet";
 import { globalOptions } from "../config";
 import { potionSetupCompleted } from "../potions";
 import { estimatedTurns } from "../turns";
-import { getBestTrainConfiguration, getBestTrainStations, grabMedicine, offsetDefaultPieces } from "./workshed_utils";
+import {
+  getBestTrainConfiguration,
+  getBestTrainStations,
+  grabMedicine,
+  offsetDefaultPieces,
+} from "./workshed_utils";
 type WorkshedOptions = {
   workshed: Item;
   done?: () => boolean;
--- a/src/post/workshed.ts
+++ b/src/post/workshed.ts
@@ -4,7 +4,7 @@ import { dietCompleted } from "../diet";
 import { globalOptions } from "../config";
 import { potionSetupCompleted } from "../potions";
 import { estimatedTurns } from "../turns";
-import { getBestCycle, getBestOffset, grabMedicine, offsetDefaultPieces } from "./workshed_utils";
+import { getBestTrainConfiguration, getBestTrainStations, grabMedicine, offsetDefaultPieces } from "./workshed_utils";
 type WorkshedOptions = {
   workshed: Item;
   done?: () => boolean;
@@ -74,6 +74,8 @@ const worksheds = [
     },
     action: () => {
       if (!TrainSet.canConfigure()) return;
+      if (globalOptions.ascend && estimatedTurns() <= 40) return;
+
       if (!get("trainsetConfiguration")) {
         // Visit the workshed to make sure it's actually empty, instead of us having not yet seen it this run
         visitUrl("campground.php?action=workshed");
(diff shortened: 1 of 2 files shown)


src/post/workshed.ts
```diff
@@ -83,10 +83,11 @@ const worksheds = [
...
         return TrainSet.setConfiguration(getBestCycle());
       } else {
-        const bestTwoStations = getBestCycle().splice(0, 2);
+        const offsetWanted = getBestOffset();
+        const stationToWatchFor = getBestCycle()[offsetWanted];
         const offset = get("trainsetPosition") % 8;
-        if (bestTwoStations.includes(TrainSet.next())) return;
-        print(`Reconfiguring trainset, as our next station is ${TrainSet.next()}`, "blue");
+        if (TrainSet.next() !== stationToWatchFor) return;
+        print(`Reconfiguring trainset.}`, "blue");
```
Pantocyclus, line 90:
```suggestion
        print(`Reconfiguring trainset.`, "blue");
```

src/post/workshed.ts
```diff
@@ -83,10 +83,11 @@ const worksheds = [
         print("Reconfiguring trainset, as our next station is empty", "blue");
         return TrainSet.setConfiguration(getBestCycle());
       } else {
-        const bestTwoStations = getBestCycle().splice(0, 2);
+        const offsetWanted = getBestOffset();
+        const stationToWatchFor = getBestCycle()[offsetWanted];
         const offset = get("trainsetPosition") % 8;
-        if (bestTwoStations.includes(TrainSet.next())) return;
-        print(`Reconfiguring trainset, as our next station is ${TrainSet.next()}`, "blue");
+        if (TrainSet.next() !== stationToWatchFor) return;
```
Pantocyclus, line 89: Might be *slightly* non-optimal if we run turns post garbo (to grab the volcoino or do a melf dupe), and we skip by `stationToWatchFor` for the overdrunk/rollover leg (which we will then have to wait a full cycle before garbo reconfigures the trainset).

src/post/workshed_utils.ts
```diff
@@ -67,6 +67,26 @@ export function getBestCycle(): TrainSet.Cycle {
   return [...trainCycle];
 }
 
+export function getBestOffset(): number {
+  const cycle = getBestCycle();
+  const valueTrainStation = (station: TrainSet.Station) =>
+    POTENTIAL_BEST_TRAIN_PIECES.find(({ piece }) => piece === station)?.value?.() ?? 0;
+  const cycleValue =
+    5 * (sum(POTENTIAL_BEST_TRAIN_PIECES, ({ value }) => value()) + valueTrainStation(cycle[0]));
```
Pantocyclus, line 75: I assume `cycle[0]` would be the coal hopper, which does not have any `value()` defined for it (and thus returns `0`). 
This should probably be `cycle[1]`?

src/post/workshed_utils.ts
```diff
@@ -67,6 +67,26 @@ export function getBestCycle(): TrainSet.Cycle {
...
+
+  const potentialOffsets = Array(POTENTIAL_BEST_TRAIN_PIECES.length)
+    .fill(null)
+    .map((_x, i) => i + 2);
```
Pantocyclus, line 79: We should only need to consider offsets [2,3,4,5,6,7], which would only require `POTENTIAL_BEST_TRAIN_PIECES.length - 2` (after all, we are already using the magic number `2` in `i + 2`)

src/post/workshed_utils.ts
```diff
@@ -67,6 +67,26 @@ export function getBestCycle(): TrainSet.Cycle {
...
+
+  return maxBy(potentialOffsets, (offset) => {
+    const tail = [...cycle].splice(1, offset - 1);
+    const tailValue = sum(tail, valueTrainStation);
+    const fullCycleValue = cycleValue + valueTrainStation(cycle[0]) + tailValue;
```
Pantocyclus, line 84: Again I assume this should be `valueTrainStation(cycle[1])`

Pantocyclus, line 84: Again I assume this should be `valueTrainStation(cycle[1])` since the coal hopper takes the value of the next station

src/post/workshed_utils.ts
```diff
@@ -67,6 +67,26 @@ export function getBestCycle(): TrainSet.Cycle {
...
+    POTENTIAL_BEST_TRAIN_PIECES.find(({ piece }) => piece === station)?.value?.() ?? 0;
+  const cycleValue =
+    5 * (sum(POTENTIAL_BEST_TRAIN_PIECES, ({ value }) => value()) + valueTrainStation(cycle[0]));
+
+  const potentialOffsets = Array(POTENTIAL_BEST_TRAIN_PIECES.length)
+    .fill(null)
+    .map((_x, i) => i + 2);
+
+  return maxBy(potentialOffsets, (offset) => {
+    const tail = [...cycle].splice(1, offset - 1);
```
Pantocyclus, line 82: Is `slice` more efficient here, since we don't need to make unnecessary `[...cycle]` copies - (although we should note that it's `slice(idx, idx+delta)` as compared to `splice(idx, delta)`)

src/post/workshed_utils.ts
```diff
@@ -67,6 +67,26 @@ export function getBestCycle(): TrainSet.Cycle {
   return [...trainCycle];
 }
 
+export function getBestOffset(): number {
```
Pantocyclus, line 70: We should probably cache this - I doubt the offset would change over the course of a run

src/post/workshed.ts
```diff
@@ -74,20 +79,21 @@ const worksheds = [
     },
     action: () => {
       if (!TrainSet.canConfigure()) return;
+      if (globalOptions.ascend && estimatedTurns() <= 40) return;
```
Pantocyclus, line 82: Oh are you not allowed to reconfigure after ascending if 40 turns have not passed since the last ascension?

horrible-little-slime: Not quite--your configuration is saved upon ascending, so ascenders will want the option to configure it themselves, ascend, and then have both the configuration they want and the ability to reconfigure at will.

src/post/workshed_utils.ts
```diff
@@ -50,26 +50,57 @@ const POTENTIAL_BEST_TRAIN_PIECES = [
...
+    for (let i = 2; i < 7; i++) {
+      const averageTurnValue =
+        (cycleValue + sum([...goodStations, cycle[i]], valueTrainStation)) / (40 + i);
+      if (averageTurnValue > currentBestAverageTurnValue) {
+        currentBestAverageTurnValue = averageTurnValue;
+        goodStations.push(cycle[i]);
+      } else {
+        bestStations = goodStations;
+        break;
+      }
```
Pantocyclus, line 94: `const goodStations = cycle.slice(0, [...Array(cycle.length-1).keys()].map((i) => (cycleValue + sum(cycle.slice(0, i+2, valueTrainStation))) / (42+i)).map((x, i) => [x, i+2]).reduce((r, a) => (a[0] > r[0] ? a : r))[1] ?? 2)`

src/post/workshed.ts
Pantocyclus: idk if I agree with this since this is technically meat-suboptimal, and only applies to users who would run trainset immediately after ascension, or specifically time their counter such that they swap in their trainset at an opportune moment.
But at the same time this only comes at a cost of, what, 1.9k? I can probably live with that; I just don't know whether this is in the spirit of things (but I also do not want to continually tag on more and more custom prefs/args which would make managing garbo way harder in the future)

horrible-little-slime: I agree re: args and prefs, but also I feel like in the competition between "maybe 2k for loopers, if the numbers work out like that" and "not locking Actual Ascenders into a suboptimal state" I care a lot more about the latter.

src/post/workshed_utils.ts
horrible-little-slime: Given that we're caching this and it's a very small array, I think we can call both of these versions "extremely cheap"

I think I will probably personally abandon this PR and let you take it over slash do it in another PR. That being said, at least one of those reduces looks like it should be a `flat`, and the other map + reduce looks like it should maybe be a `maxBy`

horrible-little-slime:
I am personally abandoning this, but will leave the branch around undeleted in case someone else picks it up. I may resume this at another point, but if I do it'll be a different approach maybe.

PR #1277, opened 2023-01-27
feat: don't count sheet in valuation

midgleyc:
At the end of each session, garbo says I've earned 900m meat from items. This isn't quite true.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -345,6 +345,7 @@ export function main(argString = ""): void {
       valueOfInventory: 2,
       suppressMallPriceCacheMessages: true,
       maximizerCombinationLimit: maximizerCombinationLimit,
+      allowNegativeTally: true,
     });
     let bestHalloweiner = 0;
     if (haveInCampground($item`haunted doghouse`)) {


horrible-little-slime:
Are you starting the day with a spooky putty monster, or is this when garbo crashes and fails to return the item, or something else?

This should be handled by [this ](https://github.com/loathers/libram/blob/98899c87fe3315216f412bccffbae5ed8b06f9c8/src/session.ts#L29-L32) section of libram's session.ts; rather than patching the acute issue, we should figure out the underlying problem at play here.

If this just started happening, it may be a weird consequence of #1256, which _should_ make garbo better at tracking items that we return to the stash later.

midgleyc:
I don't use a clan stash, I own a spooky putty sheet myself.

It ends the day as a sheet, not a monster.

It's happened every run since I started using garbo. Garbo runs all the way through, without crashing.

horrible-little-slime:
Does philter or another inventory management script closet it, and it gets uncloseted?

midgleyc:
No, but I do have "allow item counts in session results to go negative" set to false, so I guess I lose it to make the monster -> I use the monster to regain it -> the session thinks I have +1 compared to where I started.

horrible-little-slime:
Good shout; we can add that setting to the big propertiesManager call at
the start of the session, or make it get called during startSession.

Given that we also do stash returns in that earlier chunk before the main
body of garbo, we'll want to make sure the property gets set before that
code executes.

horrible-little-slime (approved):

PR #1285, opened 2023-02-03
account for nobarf interaction with estimatedTurns

Shiverwarp:


horrible-little-slime:
Instead of hardcoding this specific interaction, we should break `estimatedTurns` into 2(?) functions with divergent behavior--one to represent the number of turns we expect to run today, and one to represent the number of turns we expect _garbo_ to run today.

Or something like that.

Shiverwarp:
> Instead of hardcoding this specific interaction, we should break `estimatedTurns` into 2(?) functions with divergent behavior--one to represent the number of turns we expect to run today, and one to represent the number of turns we expect _garbo_ to run today.
> 
> Or something like that.

It's a little strange for Cleaver (especially during nobarf) specifically because it does increment on free combats and runaways, so technically this whole thing is a bit inaccurate.

We would have to estimate the number of free runs and free combats we're going to do, which I do not have much of an idea for right now.

horrible-little-slime (commented):
I think 'estimatedNonGarboTurns' is not an ideal name, but this is otherwise looking ready to ship--we can maybe talk about this in the LASS channel to see if Katarn or someone has a good suggestion

Pantocyclus:
Is this technically what you want as the solution to the cleaver problem? I'd have thought that it should be more like `remainingGarboTurns` + `remainingFreeTurns`

Shiverwarp:
> Is this technically what you want as the solution to the cleaver problem? I'd have thought that it should be more like `remainingGarboTurns` + `remainingFreeTurns`

So there's some weirdness with cleaver in this respect

- If we're just doing nobarf, then `remainingGarboTurns` will just be the embezzlers we plan to do.
- We most likely will not be using cleaver during those embezzlers
- We currently don't account for the free runs and free kills we do during nobarf which do count for June Cleaver NCs
- We only care about the remaining turns we have if they're below the threshold to cycle through every possible NC in the June Cleaver queue. If we're doing nobarf, it's most likely happening at the start of the day anyway, so it shouldn't matter.

I'm not opposed to changing it to `remainingGarboTurns` + `remainingFreeTurns` but I don't believe it will currently change (or ever change) the result we get.

Shiverwarp:
I think this should be good to merge now?

horrible-little-slime (approved):

PR #1286, opened 2023-02-05
SIT course catalogue support

horrible-little-slime:
Drafted because it is untested rn; not cutting a new eslint-plugin-libram until after the new path, I think

### DIFF
--- a/src/tasks/dailyItems.ts
+++ b/src/tasks/dailyItems.ts
@@ -302,5 +302,15 @@ export const DailyItemTasks: Task[] = [
       completed: () => get("_garbageItemChanged"),
       do: () => cliExecute("fold broken champagne bottle"),
     },
+    {
+      name: "Learn About Bugs",
+      // eslint-disable-next-line libram/verify-constants
+      ready: () => have($item`S.I.T. Course Completion Certificate`),
+      // eslint-disable-next-line libram/verify-constants
+      completed: () => get("_sitCourseCompleted", true) || have($skill`Insectologist`),
+      // eslint-disable-next-line libram/verify-constants
+      do: () => use($item`S.I.T. Course Completion Certificate`),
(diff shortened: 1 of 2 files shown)


docrostov:
image
tested, it works fine it appears

src/tasks/dailyItems.ts
```diff
@@ -302,5 +302,15 @@ export const DailyItemTasks: Task[] = [
       completed: () => get("_garbageItemChanged"),
       do: () => cliExecute("fold broken champagne bottle"),
     },
+    {
+      name: "Learn About Bugs",
+      // eslint-disable-next-line libram/verify-constants
+      ready: () => have($item`S.I.T. Course Completion Certificate`),
+      // eslint-disable-next-line libram/verify-constants
+      completed: () => get("_sitCourseCompleted", true) || have($skill`Insectologist`),
```
docrostov, line 310: my only small note on this: if prefs get misaligned, and the user doesn't have insectologist but mafia thinks the sit course isn't completed, will this construction lead to an infinite loop?

docrostov (approved):
this works under my testing scenario of "not having insectology but being able to use the thing again." i do not see any reason why this wouldn't work under other scenarios; the only small question i have is when in the login process mafia checks skills. this feels like this should work pretty well tho!

src/tasks/dailyItems.ts
horrible-little-slime: a quick search in the mafia git repository tells me that `indelible-for-a-day` appears here:
https://github.com/kolmafia/kolmafia/blob/881e876fc2f41706259097fa9e41ab34b4779fcc/src/net/sourceforge/kolmafia/request/UseItemRequest.java#L6008-L6012

TL;DR: we're fine

PR #1287, opened 2023-02-06
Do a better 40 turns trick; be good to ascenders

horrible-little-slime:


src/post/workshed_utils.ts
```diff
@@ -67,7 +67,30 @@ export function getBestCycle(): TrainSet.Cycle {
...
+
+let bestOffset: number | null = null;
+function getBestOffset(): number {
+  bestOffset ??= maxBy([2, 3, 4, 5, 6, 7], valueOffset);
+  return bestOffset;
+}
+
+export function getPrioritizedStations(): TrainSet.Station[] {
+  return getBestCycle().slice(0, getBestOffset());
+}
```
Pantocyclus, line 91: There's an off-by-one error here I think.
When `bestOffset` is `n`, we consider up to `n-1` stations after the Coal Hopper (hence when `n` is 2, we consider the 1st station after). Here we only consider up to `n=7` in the `maxBy`, which is only 6 after the coal hopper, and we will never consider the last station

src/post/workshed.ts
```diff
@@ -81,13 +81,14 @@ const worksheds = [
...
-        const bestTwoStations = getBestCycle().splice(0, 2);
-        const offset = get("trainsetPosition") % 8;
-        if (bestTwoStations.includes(TrainSet.next())) return;
+        const bestStations = getPrioritizedStations();
+        if (bestStations.includes(TrainSet.next())) return;
+        if (globalOptions.ascend && estimatedTurns() <= 40) return;
         print(`Reconfiguring trainset, as our next station is ${TrainSet.next()}`, "blue");
-        return TrainSet.setConfiguration(offsetDefaultPieces(offset));
+        const position = get("trainsetPosition") % 8;
+        return TrainSet.setConfiguration(getRotatedCycle(position));
```
Pantocyclus, line 91: Minor style nitpicks:
`get("trainsetPosition") % 8` is assigned to `position` in the `else` case, but not in the `if` case.
Also `if (globalOptions.ascend && estimatedTurns() <= 40) return;` can probably be pushed before the `getPrioritizedStations()` compute, since it is cheaper

Pantocyclus (approved):
I think this looks fine

PR #1294, opened 2023-02-13
Add Shadow Rift to item steal zones

gausie:
- Update dependencies
- Add Shadow Rift to item steal zones


src/fights.ts
```diff
@@ -2193,6 +2193,16 @@ const itemStealZones = [
     openCost: () => 0,
     preReq: null,
   },
+  {
+    location: $location`Shadow Rift`,
+    monster: $monster`shadow slab`,
+    item: $item`shadow brick`,
+    requireMapTheMonsters: true,
```
Tokoeka, line 2200: i swear when I tried to map the monsters in the rift, it just didnt work (i think because its nowander?)

gausie: You're probably right. I've tried to address this but am out of my depth without some more familiarity on the item steal zone code

horrible-little-slime:
This is failing linting due to import order but otherwise looks fine

PR #1300, opened 2023-02-28
Use `driftwood beach comb` for buff and free walks

eegeeZA:
If there is a "piece of driftwood" in inventory and it can be used to grant temporary access to Comb the Beach, then use it.

### DIFF
--- a/src/tasks/daily.ts
+++ b/src/tasks/daily.ts
@@ -408,15 +408,23 @@ export const DailyTasks: Task[] = [
     do: () => cliExecute("horsery dark"),
   },
   {
+    name: "Beach Comb One-Day",
+    ready: () =>
+      have($item`piece of driftwood`) &&
+      (!have($item`Beach Comb`) || !have($item`driftwood beach comb`)),
+    completed: () => have($item`driftwood beach comb`),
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
does `BeachComb.tryHead` work properly with driftwood beach combs?

The wiki is down right now--is there a reasonable way to value the driftwood beach comb, or can we really treat it as free?

eegeeZA:
Yeah, tested and it works.
image

I was thinking that one could purchase it before Garbo runs. But if I'm following your train of thought, Garbo could also purchase it. The first free adventure is used for the Familiar Weight buff. The value using the remaining 10 adventures could be done assuming one finds the average of 3 grains of sand.

Based on my last run below, the value was ~12640 meat, with unspectacular combing results. Using the above calculation would place it at ~12600 meat which is close enough.
```
=== RESULTS ===
-Found 0 meat
-Found 1 sea carrot
-Found 27 grain of sand
-Found 3 magenta seashell
-Found 2 gray seashell
-Found 1 kelp
-Found 1 driftwood hat
-Found 1 driftwood pants
-Found 1 driftwood bracelet
```

PR #1301, opened 2023-03-01
changes per scotch's abnormal YR behavior

horrible-little-slime:


docrostov (approved):
looks good to me bucko

Issue #1297, opened 2023-02-24
[bug] need to expand spleen consideration 


docrostov:
https://github.com/Loathing-Associates-Scripting-Society/garbage-collector/blob/00188b1392466335ea159fd01b34c1fc05195cd9/src/diet.ts#L363

user reported the script was not considering agua. i figured there wouldn't be a world where we'd want to do agua but i was wrong. current prices are agua de vida = 5482 for 4 spleen @ 1.88 advs per spleen, versus transdermal smoke patch at 1 spleen @ 2.5 adv per spleen. i did a quick table and the user is correct. the loss is relatively low at the 5500-6000 range most of us are at but is more noticeable at lower levels where min VOA users are probably missing out on roughly 20k meat per day; not much but not nothing. easy PR to just add in other spleeners things but i am going to bed lol

![image](https://user-images.githubusercontent.com/8014761/221105528-306d6b28-61ce-4920-bccb-9d42719ef4a4.png)

horrible-little-slime:
I would describe this as a feature request rather than a bug. but yeah this should be pretty doable.

Could you round up a list of all equivalent spleen items?

docrostov:
resolved in #1303. 

(i will take issue with this as a feature request vs a bug. i think this is a straightforward enough add -- and the PR was easy enough -- that it is worth saying "our diet was not considering any of like 30 readily available spleen items" was kind of a bug because it meant for non-extro/jelly folks garbo was leaving a decent gob of spleen-y value on the table. :<)

PR #1303, opened 2023-03-01
Add more spleen items

spaghetti-squash:
#1297

### DIFF
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -321,7 +321,7 @@ function menu(): MenuItem<Note>[] {
    * standardSpleenItem indicates a spleen item of size 4 with an adventure yield of 5-10. Taken from the wiki. They are all functionally equivalent.
    */
   const standardSpleenItems = $items`agua de vida, gooey paste, oily paste, ectoplasmic paste, greasy paste, bug paste, hippy paste, orc paste, demonic paste, indescribably horrible paste, fishy paste, goblin paste, pirate paste, chlorophyll paste, strange paste, Mer-kin paste, slimy paste, penguin paste, elemental paste, cosmic paste, hobo paste, Crimbo paste, groose grease, Unconscious Collective Dream Jar, grim fairy tale, powdered gold`;
-  
+
   const smallEpics = [...$items`meteoreo, ice rice`, $item`Tea, Earl Grey, Hot`];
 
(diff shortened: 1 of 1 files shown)


src/diet.ts
```diff
@@ -316,6 +316,8 @@ function menu(): MenuItem<Note>[] {
    *  > js Item.all().filter((item) => item.fullness > 0 && item.name.indexOf("lasagna") > 0 && getIngredients(item)["savory dry noodles"]).join(", ")
    */
   const lasagnas = $items`fishy fish lasagna, gnat lasagna, long pork lasagna`;
+
```
docrostov, line 319: can we add a comment noting what standardSpleenItem means, something like:
``` // standardSpleenItem indicates a spleen item of size 4 with an adventure yield of 5-10. They are functionally equivalent.```

**Review by docrostov — APPROVED** [review]
i would appreciate if you added the comment requested but otherwise it's fine.

are we just ignoring beastly paste? it is ok if the answer is "yeah".

File: src/diet.ts (hunk 1 above)
Comment (horrible-little-slime) [reply src/diet.ts:319]: If you make that a formal suggestion I can commit it right now otherwise it'll have to wait until I'm at my computer

**Comment by horrible-little-slime** [comment]
> i would appreciate if you added the comment requested but otherwise it's fine.
> 
> are we just ignoring beastly paste? it is ok if the answer is "yeah".

Opposite! we already account for beastly paste in our potionmenu

File: src/diet.ts [hunk 2]
```diff
@@ -316,6 +316,12 @@ function menu(): MenuItem<Note>[] {
...
+  /* 
+   * standardSpleenItem indicates a spleen item of size 4 with an adventure yield of 5-10. Taken from the wiki. They are all functionally equivalent.
```
docrostov, line 321: wow what an incredible comment whoever wrote this must be a really strong individual with a great spirit and a happy heart, who has never once in their life experienced depression. incredible. great stuff.

docrostov (approved):
approved as i added the comment

src/diet.ts
docrostov: now that i've removed trailing spaces this is even more beautiful

docrostov (dismissed):
crying myself to sleep at how beautiful this comment is

PR #1305, opened 2023-03-10
Add initial support for Avatar of Shadows Over Loathing

s-k-z:
Support for Pig Skinner red-ray, Cheese Wizard yellow ray and Stilton Splatter, and Jazz Agent buffs

### DIFF
--- a/src/barfTurn.ts
+++ b/src/barfTurn.ts
@@ -309,6 +309,35 @@ const turns: AdventureAction[] = [
     sobriety: Sobriety.EITHER,
   },
   {
+    name: "Cheese Wizard Fondeluge",
+    available: () =>
+      have($skill`Fondeluge`) &&
+      !have($effect`Everything Looks Yellow`) &&
+      romanticMonsterImpossible(),
+    execute: () => {
+      const usingDuplicate = SourceTerminal.have() && SourceTerminal.duplicateUsesRemaining() > 0;
+
+      const location = wanderWhere("yellow ray");
+      const familiar = freeFightFamiliar({ location, allowAttackFamiliars: !usingDuplicate });
+      useFamiliar(familiar);
+      freeFightOutfit();
+      if (usingDuplicate) {
+        SourceTerminal.educate([$skill`Extract`, $skill`Duplicate`]);
+      }
+      const macro = Macro.if_(embezzler, Macro.meatKill())
+        .familiarActions()
+        .externalIf(usingDuplicate, Macro.trySkill($skill`Duplicate`))
+        .skill($skill`Fondeluge`);
+      garboAdventureAuto(location, macro);
+      if (SourceTerminal.have()) {
+        SourceTerminal.educate([$skill`Extract`, $skill`Digitize`]);
+      }
+      return have($effect`Everything Looks Yellow`);
+    },
+    spendsTurn: false,
+    sobriety: Sobriety.SOBER,
+  },
+  {
     name: "Spit Acid",
     available: () =>
       have($item`Jurassic Parka`) &&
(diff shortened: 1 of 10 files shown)


src/wanderer/lib.ts
```diff
@@ -104,6 +104,8 @@ export function canWander(location: Location, type: DraggableFight): boolean {
   switch (type) {
     case "backup":
       return canWanderTypeBackup(location);
+    case "pigskinner":
+      return have($skill`Free-For-All`) && canWanderTypeYellowRay(location);
```
horrible-little-slime, line 108: I don't think we want this logic included in wanderer; we should just have this check in barfTurn, and have pigskinner either use the yellow ray code or the backup code.

src/barfTurn.ts
```diff
@@ -338,6 +367,34 @@ const turns: AdventureAction[] = [
...
     sobriety: Sobriety.SOBER,
   },
+  {
+    name: "Pig Skinner Free-For-All",
+    available: () =>
+      have($skill`Free-For-All`) &&
+      !have($effect`Everything Looks Red`) &&
+      romanticMonsterImpossible(),
+    execute: () => {
+      const usingDuplicate = SourceTerminal.have() && SourceTerminal.duplicateUsesRemaining() > 0;
```
horrible-little-slime, line 377: Given that FFA doesn't force drops, we probably don't want to use duplicate here

src/barfTurn.ts
```diff
@@ -338,6 +367,34 @@ const turns: AdventureAction[] = [
...
+      const location = wanderWhere("pigskinner");
+      const familiar = freeFightFamiliar({ location, allowAttackFamiliars: !usingDuplicate });
```
horrible-little-slime, line 379: I'm not sure if we currently have code to prevent the use of the crimbo tree, but we should prevent the use of the crimbo tree

s-k-z: What crimbo tree? Why?

src/wanderer/lib.ts
s-k-z: It's also used to tell whether a Guzzlr quest can be completed in the available turns.

src/barfTurn.ts
horrible-little-slime: shrubling

src/wanderer/lib.ts
horrible-little-slime: Sure, but if we're treating it as identical to backup, for example, we can just change `pigSkinnerRay` to `canWanderCache["backup"] && have($skill`Free-For-All`) ? Math.floor(estimatedGarboTurns() / 25) : 0`

which, upon reflection, we should also have yellowRay check to see if we own parka before counting those. That's out of scope of this PR, just not noticed it.

s-k-z: That seems like too low of an estimate since the two are independent.

src/barfTurn.ts
s-k-z: This looks covered already [where we select familiars](https://github.com/Loathing-Associates-Scripting-Society/garbage-collector/blob/main/src/familiar/freeFightFamiliar.ts#L53) and [how we check if we can open a red present](https://github.com/Loathing-Associates-Scripting-Society/garbage-collector/blob/37a1bffa82600acbd569a7d5ddbca7ca8846e4ff/src/familiar/lib.ts#L77)

s-k-z: Oh, but the issue is we can try to do both together?

horrible-little-slime: But any time we're eligible for Free For All, we necessarily don't have ELR, which means we're eligible to open a present; we need to add a condition to `canOpenRedPresent` that says that if we have Free For All, we can't open a present

horrible-little-slime (approved):

PR #1308, opened 2023-03-15
Add abort messages for typical error cases

s-k-z:


horrible-little-slime:
[BALLS docs on the wiki](https://kol.coldfront.net/thekolwiki/index.php/Combat_Macros#abort_.5Bpredicate.5D) do describe an `abort X` command, but only one that uses a macro predicate, not a string input; has this been tested and confirmed to work with autoattack?

s-k-z:
It works with autoattack, the result is:
```
Preference lastMacroError changed from Macro Aborted ("abort") to Macro Aborted ("abort "default error message"")
```

PR #1309, opened 2023-03-17
Delay installing workshed until end if possible

s-k-z:
For now this just saves CMC until you have 105 estimated turns remaining, that way you get the most out of DNA lab or model train set.


horrible-little-slime:
Where does 25 come from?

s-k-z:
It's an arbitrary amount that I chose in case you need to run more turns to get a pill.

horrible-little-slime:
Make it a named const (maybe with a comment that says it's semi-arbitrary) so that we don't have a magic number and then this is good to go

PR #1313, opened 2023-03-20
Backup safe list

Shiverwarp:
This allows us to wander in locations that are safe to backup in because all NCs in the area are skippable or lead to a combat.

Unfortunately I cannot test this within normal garbo, I do not have guzzlr. I have however been running essentially this with a custom garbo that does Doc Bag locations and has been working well.

I'm not 100% happy the way I'm setting the choice option properties right now (Just always setting them in main), but I have yet to come up with a better answer.

**Locations that we may also consider later for long aftercore (or if there is some way to know whether we have already encountered one-time adventures this ascension):**
- The Sleazy Back Alley
- The Haunted Pantry
- The Outskirts of Cobb's Knob 

**Locations we may consider with additional handling:**
- Madness Bakery (Only if you have enough ingredients to make popular tarts, have done the quest, and creating popular tarts is meat positive)
- The Haunted Billiards Room (Only safe if you already have a pool cue. Possibly unsafe if stolen via pvp etc)

**Locations that may still be profitable to wander in despite NCs (Demon name superlikelies that are 1/150 chance)**
- The Haunted Bathroom 
- The Spooky Forest
- Pandamonium Slums (Also has a one-time adventure)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -372,13 +372,44 @@ export function main(argString = ""): void {
       1341: 1, // Cure her poison
     });
 
+    const quartetChoice = get("lastQuartetRequest", 0) !== 0 ? get("lastQuartetRequest") : 4;
     // Non-combat options for backup wanderer locations
     propertyManager.setChoices({
-      522: 2, // Cobb's Knob Barracks, skip
-      1107: 1, // tennis ball
-      1108: bestHalloweiner,
-      1340: 1, // Accept the doctor quest
-      1341: 1, // Cure her poison
+      // Cobb's Knob Barracks
+      522: 2, // Skip
+      // The Castle in the Clouds in the Sky (Basement)
+      669: 1, // Neckbeard choice
+      670: 4, // Skip
+      671: 4, // Fitness choice
+      // The Castle in the Clouds in the Sky (Ground Floor)
+      672: 3, // Skip
+      673: 3, // Skip
+      674: 3, // Skip
+      1026: 3, // Skip
+      // The Castle in the Clouds in the Sky (Top Floor)
+      675: 4, // Steampunk choice
+      676: 4, // Punk Rock choice
+      677: 1, // Fight Steam Punk Giant
+      678: 3, // Steampunk choice
+      // The Penultimate Fantasy Airship
+      182: 1, // Fight random enemy
+      178: 2, // Skip
+      // The Haunted Gallery
+      91: 2, // Skip
+      89: 6, // Skip
+      // The Haunted Library
+      163: 4, // Skip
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -372,6 +372,46 @@ export function main(argString = ""): void {
...
+      // The Haunted Ballroom
+      106: 3, // Skip
+      90: quartetChoice, // Choose currently playing song, or skip
+      // The Haiku Dungeon
+      297: 3, // Skip
+      // A Mob of Zeppelin Protesters - Probably unecessary, these shouldn't show up after ascension is complete
+      856: 2, // Skip
+      857: 2, // Skip
+      858: 2, // Skip
+    });
```
horrible-little-slime, line 413: We actually already have a location to put choice adventures like this--it's `unsupportedChoices` in `wanderer/lib.ts`

src/wanderer/lib.ts
```diff
@@ -83,8 +83,13 @@ export function unlock(loc: Location, value: number): boolean {
 }
 
 const backupSkiplist = $locations`The Overgrown Lot, The Skeleton Store, The Mansion of Dr. Weirdeaux`;
+// These are locations where all non-combats have skips or lead to a combat.
+const backupSafelist = $locations`The Haunted Gallery, The Haunted Ballroom, The Haunted Library, The Penultimate Fantasy Airship, Cobb's Knob Barracks, The Castle in the Clouds in the Sky (Basement), The Castle in the Clouds in the Sky (Ground Floor), The Castle in the Clouds in the Sky (Top Floor), The Haiku Dungeon, Twin Peak, A Mob of Zeppelin Protesters`;
```
horrible-little-slime, line 87: It might be worth breaking up the "have skips" and the "or lead to a combat" chunks here--the former is also safe to use for YRs but the latter is not

horrible-little-slime (changes requested):

src/index.ts
Shiverwarp: Ahh, thanks! I'll work on changing it into that format there

src/wanderer/lib.ts
Shiverwarp: The YR wandering should also be safe for the ones that lead to a combat, no? From what I understand those monsters are also included in the location monsters, or am I mistaken? This is what lead to YR's being used in the ballroom for a while iirc

horrible-little-slime: For YRs, at the very least, we do need to treat them differently. We use `appearanceRates` now, which should filter out things like the ballroom guy

horrible-little-slime (approved):

PR #1319, opened 2023-03-23
Initial support for Shadow Waters and Affinity

Pantocyclus:


gausie:
I had some issues with this heading back into the payphone NC and quitting out because it didn't expect to see a new quest

src/fights.ts
```diff
@@ -351,6 +358,26 @@ function pygmyOptions(forceEquip: Item[] = []) {
...
+let _bestShadowRift: Location | null = null;
+export function bestShadowRift(): Location {
+  if (!_bestShadowRift) {
+    _bestShadowRift =
+      chooseRift({
+        canAdventure: true,
+        sortBy: (l: Location) => {
+          const drops = getMonsters(l)
+            .map((m) => Object.keys(itemDrops(m)).map((s) => toItem(s)))
+            .reduce((acc, val) => acc.concat(val), []);
```
horrible-little-slime, line 370: Why don't we want to use `flat` here? It works just fine in `autumnaton.ts`

src/fights.ts
```diff
@@ -351,6 +358,26 @@ function pygmyOptions(forceEquip: Item[] = []) {
...
+          return sum(drops, garboValue);
+        },
+      }) ?? $location.none;
+    if (_bestShadowRift === $location.none) {
+      throw new Error("Failed to find a suitable Shadow Rift to adventure in");
+    }
```
horrible-little-slime, line 376: I think we can just keep the nullish value here instead of using Location.none as a dummy value

src/fights.ts
```diff
@@ -351,6 +358,26 @@ function pygmyOptions(forceEquip: Item[] = []) {
...
+  }
+  return myInebriety() > inebrietyLimit() ? $location`Shadow Rift` : _bestShadowRift;
```
horrible-little-slime, line 378: I don't understand the intention of this

src/fights.ts
```diff
@@ -1356,6 +1383,36 @@ const freeFightSources = [
...
     }
   ),
+  new FreeFight(
+    () => {
+      if (!have($item`closed-circuit pay phone`)) return false;
+      else if (have($effect`Shadow Affinity`)) return true;
+      else if (get("_shadowAffinityToday", false)) return false;
+
+      if (rufusTarget() === null) return true;
+      else if (get("rufusQuestType", "") === "items") {
```
horrible-little-slime, line 1393: You shouldn't need a 2-argument `get` with the latest libram

src/fights.ts
```diff
@@ -1356,6 +1383,36 @@ const freeFightSources = [
...
+        return false; // We deemed it unprofitable to complete the quest in potionSetup
+      } else if (get("encountersUntilSRChoice", 0) === 0) {
```
horrible-little-slime, line 1395: You shouldn't need a 2-argument `get` with the latest libram

src/fights.ts
```diff
@@ -1356,6 +1383,36 @@ const freeFightSources = [
...
+        // Target is either an artifact or a boss
+        return true; // Get the artifact or kill the boss immediately for free
+      }
+      return false; // We have to spend turns to get the artifact or kill the boss
```
horrible-little-slime, line 1399: imo this section will read better if we strip the `else`s; early returns make them unnecessary

src/fights.ts
```diff
@@ -1356,6 +1383,36 @@ const freeFightSources = [
...
+    },
+    () => {
+      propertyManager.set({ shadowLabyrinthGoal: "effects" });
+      if (!get("_shadowAffinityToday", false) && rufusTarget() === null) {
```
horrible-little-slime, line 1403: 2-arg get, etc

src/potions.ts
```diff
@@ -35,13 +44,16 @@ import {
...
+  withChoice,
 } from "libram";
 import { acquire } from "./acquire";
 import { baseMeat, HIGHLIGHT, maxBy, pillkeeperOpportunityCost, turnsToNC } from "./lib";
 import { embezzlerCount } from "./embezzler";
 import { usingPurse } from "./outfit";
 import { estimatedGarboTurns } from "./turns";
 import { globalOptions } from "./config";
+import { bestShadowRift } from "./fights";
+import { chooseQuest, rufusTarget } from "libram/dist/resources/2023/ClosedCircuitPayphone";
```
horrible-little-slime, line 56: just import the module from libram instead of importing the individual functions from the libram module

src/potions.ts
```diff
@@ -348,6 +382,117 @@ function useAsValuable(potion: Potion, embezzlers: number, embezzlersOnly: boole
...
+      return Infinity;
+    }
+
+    // We consider the average price of the shadow items to not get gated behind an expensive one
+    const shadowItems = $items`shadow brick, shadow ice, shadow sinew, shadow glass, shadow stick, shadow skin, shadow flame, shadow fluid, shadow sausage, shadow bread, shadow venom, shadow nectar`;
+    const averagePrice =
+      sum(shadowItems, (it) =>
+        historical && historicalAge(it) < 14 ? historicalPrice(it) : mallPrice(it)
+      ) / shadowItems.length;
+
```
horrible-little-slime, line 454: `garboAverageValue` exists, and `garboAverageValue(items, true)` should use historical price

src/potions.ts
```diff
@@ -348,6 +382,117 @@ function useAsValuable(potion: Potion, embezzlers: number, embezzlersOnly: boole
...
+  },
+});
+
+const rufusPotion = new Potion($item`closed-circuit pay phone`, {
+  providesDoubleDuration: false,
+  canDouble: false,
+  effect: $effect`Shadow Waters`,
+  duration: 30,
+  price: (historical: boolean) => {
+    const canAcquireItemQuest = rufusTarget() === null;
```
horrible-little-slime, line 424: bind the target to a variable; it'll make the nullish checks and type assertions much easier

src/potions.ts
```diff
@@ -348,6 +382,117 @@ function useAsValuable(potion: Potion, embezzlers: number, embezzlersOnly: boole
...
+      ) / shadowItems.length;
+
+    return averagePrice;
+  },
+  acquire: (qty: number) => {
+    equip($slot`weapon`, $item.none);
+    equip($slot`off-hand`, $item`Drunkula's wineglass`);
+    Array(qty)
+      .fill(0)
+      .every(() => {
```
horrible-little-slime, line 462: This can just be a `for` loop

horrible-little-slime (changes requested):

src/fights.ts
Pantocyclus: image
It's to address a silly bug

src/potions.ts
Pantocyclus: `garboAverageValue` doesn't support a boolean for historical price unfortunately

src/fights.ts
Pantocyclus: I guess my yarn was also a little wonky after the flat polyfill broke previously

src/fights.ts
horrible-little-slime: Then put a comment to address that

src/potions.ts
horrible-little-slime: This should probably be sliced out but that can be future you's problem. Or future me. Or future whoever, it's not hard.

src/fights.ts
```diff
@@ -158,6 +160,11 @@ import { garboValue } from "./session";
 import { bestConsumable } from "./diet";
 import { wanderWhere } from "./wanderer";
 import { globalOptions } from "./config";
+import {
+  chooseQuest,
+  chooseRift,
+  rufusTarget,
+} from "libram/dist/resources/2023/ClosedCircuitPayphone";
```
horrible-little-slime, line 167: This also appears to import functions rather than the whole module

src/fights.ts
```diff
@@ -351,6 +358,26 @@ function pygmyOptions(forceEquip: Item[] = []) {
...
 
+let _bestShadowRift: Location | null = null;
+export function bestShadowRift(): Location {
+  if (!_bestShadowRift) {
+    _bestShadowRift =
+      chooseRift({
+        canAdventure: true,
+        sortBy: (l: Location) => {
+          const drops = getMonsters(l)
+            .map((m) => Object.keys(itemDrops(m)).map((s) => toItem(s)))
```
horrible-little-slime, line 369: Not all drops drop at the same rate, and we probably aren't capping them, so it might be worth using an `Object.entries` to weight them by drop rate

src/fights.ts
```diff
@@ -351,6 +358,26 @@ function pygmyOptions(forceEquip: Item[] = []) {
...
+            .flat();
+          return sum(drops, garboValue);
+        },
+      }) ?? $location.none;
```
horrible-little-slime, line 373:
```suggestion
      });
```
I think you just missed this part

horrible-little-slime (changes requested):

src/fights.ts
```diff
@@ -361,20 +358,28 @@ function pygmyOptions(forceEquip: Item[] = []) {
...
+        // We probably aren't capping item drops with the penalty
+        // so we don't really need to compute the actual outfit (or the dropModifier for that matter actually)
+        const dropModifier = 1 + numericModifier("Item Drop") / 100;
+        return sum(getMonsters(l), (m) => {
+          const monsterDrops = itemDrops(m);
+          const items = Object.keys(monsterDrops).map(toItem);
+          const rates = Object.values(monsterDrops);
+          return sumNumbers(
+            items.map((item, idx) => garboValue(item) * clamp(rates[idx] * dropModifier, 0, 1))
+          );
```
horrible-little-slime, line 374: it won't let me make a formal suggestion for whatever reason, which sucks, but:
```
const drops = [...Object.entries(monsterDrops)];
return sum(drops, ([drop, rate]) => garboValue(toItem(drop)) * clamp(rate * dropModifier / 100, 0, 1));
```
alternately, and this might be easier:
```
return sum(itemDropsArray(m), ({ drop, rate }) => garboValue(drop) * clamp(rate * dropModifier / 100, 0, 1));
```

okay, yeah, definitely the second one

src/fights.ts
```diff
@@ -361,20 +358,28 @@ function pygmyOptions(forceEquip: Item[] = []) {
...
+    _bestShadowRift = ClosedCircuitPayphone.chooseRift({
+      canAdventure: true,
+      sortBy: (l: Location) => {
+        setLocation(l);
+        // We probably aren't capping item drops with the penalty
+        // so we don't really need to compute the actual outfit (or the dropModifier for that matter actually)
+        const dropModifier = 1 + numericModifier("Item Drop") / 100;
+        return sum(getMonsters(l), (m) => {
+          const monsterDrops = itemDrops(m);
+          const items = Object.keys(monsterDrops).map(toItem);
```
horrible-little-slime, line 370: Just as a heads up, taking native mafia functions and dropping them directly in like that will typically cause errors due to differences in the ways js and ash handle overloads; mafia will say "there's no such thing as `to_item(string, number, string[]`"

horrible-little-slime: That will be moot due to the below comment, but it's still a good note

src/fights.ts
```diff
@@ -361,20 +358,28 @@ function pygmyOptions(forceEquip: Item[] = []) {
...
+          const rates = Object.values(monsterDrops);
```
horrible-little-slime, line 371: I'm assuming you did this because Object.entries doesn't return an array, it returns an IterableIterator--we can use either `Array.from` or spread syntax to convert that to an array. so `[...Object.entries(monsterDrops)]` would have worked there. Again, moot because I think below handles it better

src/fights.ts
horrible-little-slime: `drop` is an Item in `itemDropsArray` but a string in `itemDrops` because only strings can be the keys to an object, but itemDropsArray doesn't actually use the drop as a key

src/fights.ts
```diff
@@ -361,20 +358,28 @@ function pygmyOptions(forceEquip: Item[] = []) {
...
+            items.map((item, idx) => garboValue(item) * clamp(rates[idx] * dropModifier, 0, 1))
+          );
+        });
+      },
+    });
     if (!_bestShadowRift) {
       throw new Error("Failed to find a suitable Shadow Rift to adventure in");
     }
   }
+  // Mafia bug disallows adv1($location`Shadow Rift (<exact location>)`, -1, "") when overdrunk
```
horrible-little-slime, line 382: I'll put in a PR for this maybe

src/fights.ts
```diff
@@ -1389,7 +1394,7 @@ const freeFightSources = [
       if (have($effect`Shadow Affinity`)) return true;
       if (get("_shadowAffinityToday")) return false;
 
-      if (rufusTarget() === null) return true;
+      if (ClosedCircuitPayphone.rufusTarget() === null) return true;
```
horrible-little-slime, line 1397: I think this is more legible using `!ClosedCircuitPayphone.rufusTarget()` instead of checking against null, but that is a matter of opinion and is extremely not a blocker

horrible-little-slime (changes requested):

horrible-little-slime (approved):
Approved pending testing by the pantocyclus himself

PR #1329, opened 2023-03-29
Correctly account for spleen cleansers (and not for PYEC)

Pantocyclus:


horrible-little-slime (approved):
Approving but not immediately merging because I forgot if you had tested this

Pantocyclus:
OK this seems good to go

PR #1331, opened 2023-03-30
Check for free beachwalks available before getting effect

Shiverwarp:
Garbo will currently spend an adventure to get "Do I Know You From Somewhere?" if you used all your free beach walks before running it

### DIFF
--- a/src/mood.ts
+++ b/src/mood.ts
@@ -20,6 +20,7 @@ import {
   $items,
   $skill,
   AsdonMartin,
+  BeachComb,
   get,
   have,
   Mood,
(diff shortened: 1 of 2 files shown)


horrible-little-slime:
I think we changed this deliberately to burn a turn getting the buff because it's worth it, which is to say you will need to fight to the death with whoever wanted that

Shiverwarp:
This would heavily depend on your VoA, and whether you're yachtzeechaining, how many embezzlers you do, and if you have robort I think?

With a hobo monkey (starting at 60 weight, which I believe is low for most) and if all 50 adventures are embezzlers, it's worth ~4435 by my calculations, which might be worth it for some people, but I think 60 weight is probably not realistic for most? If we start at 150 weight, it's only worth ~3960. Obviously this is much more worth for people with robort or yachtzee.

but the adventure gets used while we're still doing free fight things, so it can be detrimental by burning buffs useful there (like item drop for drum machines)

Anyway, not particularly worried either way, this just ate my squint while I was doing garbo nobarf so figured this implementation would be safer for most people. Can close this if it's not needed

horrible-little-slime:
yeah Im going to go ahead and merge this and then whoever suggested the previous change can duke it out with you

horrible-little-slime (changes requested):
oh wait, first, switch this to use the new-ish libram BeachComb support

horrible-little-slime:
just checking for headAvailable and freeCombs

PR #1335, opened 2023-04-02
Don't throw on a single accidental non-free kill

Pantocyclus:


src/fights.ts
```diff
@@ -566,7 +567,9 @@ class FreeFight {
       safeRestore();
       const curTurncount = myTurncount();
       withMacro(Macro.basicCombat(), this.run);
-      if (myTurncount() > curTurncount) throw new Error("The last fight was not free!");
+      if (myTurncount() > curTurncount) consecutiveNonFreeFights++;
+      else consecutiveNonFreeFights = 0;
+      if (consecutiveNonFreeFights >= 5) throw new Error("The last 5 FreeRunFights were not free!");
```
horrible-little-slime, line 572: I feel like honestly even 2 would be fine, but that's not a blocker

horrible-little-slime (approved):

src/fights.ts
Pantocyclus: We can reduce this in the future, but let's keep it at 5 for now

PR #1337, opened 2023-04-07
Support paw wishes

Pantocyclus:


src/potions.ts
```diff
@@ -79,6 +82,27 @@ for (const effectGroup of mutuallyExclusiveList) {
...
+  if (!validWishString.has(effect)) {
+    const allOtherNoHookahEffectNames = allNoHookahEffects
+      .filter((e) => e !== effect)
+      .map((e) => e.name.toLowerCase());
+    const validSubstrings = Array.from(
+      effect.name.toLowerCase().match(RegExp(/[\w ]+/g)) ?? []
+    ).filter((s) => !allOtherNoHookahEffectNames.some((name) => name.includes(s)));
+    validWishString.set(
+      effect,
+      (validSubstrings.length > 0 ? validSubstrings.pop() ?? "" : "").trim()
```
horrible-little-slime, line 97:
```suggestion
      (validSubstrings[0]?.trim() ?? "")
```

src/potions.ts
```diff
@@ -460,18 +484,64 @@ const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+export const wishPotions = validWishes.map(
+  (effect) =>
+    new Potion($item`pocket wish`, {
+      effect,
+      canDouble: false,
+      duration: 20,
+      use: (quantity: number) =>
+        new Array(quantity).fill(0).every(() => {
+          const s = getValidWishString(effect);
+          if (s === "") return;
```
horrible-little-slime, line 496:
```suggestion
          if (!s) return;
```

src/potions.ts
```diff
@@ -460,18 +484,64 @@ const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+          cliExecute(`genie effect ${s}`);
+        }),
+    })
```
horrible-little-slime, line 499: This is now long enough that we can't justify using Array.fill.every, and should instead use a `for` loop in a code block

src/potions.ts
```diff
@@ -460,18 +484,64 @@ const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
         canDouble: false,
-        duration: 20,
-        use: (quantity: number) =>
-          new Array(quantity).fill(0).every(() => cliExecute(`genie effect ${effect}`)),
+        price: () =>
+          get("_pawWishes", 0) >= 5 ||
+          // eslint-disable-next-line libram/verify-constants
+          !have($item`cursed monkey's paw`) ||
+          failedWishes.includes(effect)
+            ? Infinity
```
horrible-little-slime, line 514: I love using `Infinity`, but keep in mind that if we ever `sort` the array it'll cause issues (the `reduce` used in `maxBy` is fine, though)

src/potions.ts
```diff
@@ -544,6 +614,30 @@ export function potionSetup(embezzlersOnly: boolean): void {
...
+    if (
+      !pawPotions
+        .filter((potion) => numericModifier(potion.effect(), "Meat Drop") >= 100)
+        .sort((a, b) => {
+          const aValue = a
+            .value(embezzlers)
+            .sort((l, r) => (r.quantity > 0 ? r.value : 0) - (l.quantity > 0 ? l.value : 0));
+          const bValue = b
+            .value(embezzlers)
+            .sort((l, r) => (r.quantity > 0 ? r.value : 0) - (l.quantity > 0 ? l.value : 0));
```
horrible-little-slime, line 630: this is hard to follow and probably should just use a `maxBy` given that you're ultimately checking their highest value

src/potions.ts
```diff
@@ -544,6 +614,30 @@ export function potionSetup(embezzlersOnly: boolean): void {
...
+          return (
+            (bValue.length > 0 ? bValue[0].value : 0) - (aValue.length > 0 ? aValue[0].value : 0)
+          );
+        })
+        .some((potion) => potion.use(1))
```
horrible-little-slime, line 635: This also does not seem like a great structure to end up in

src/potions.ts
horrible-little-slime: honestly, given that we always check against "is an empty string", using `null` or `undefined` instead of `""` is going to be easier to follow imo

horrible-little-slime (changes requested):
I think @pstalcup will have more to say in the coming days, and given that the tracking PR is now open in mafia (thanks Veracity!), we should just wait for that and update our mafia version.

I'll cut a new eslint-plugin-libram now

src/potions.ts
Pantocyclus: I think it's fine. We should never want to use a potion that has a cost of Infinity

src/potions.ts
```diff
@@ -79,6 +82,24 @@ for (const effectGroup of mutuallyExclusiveList) {
...
+    const allOtherNoHookahEffectNames = allNoHookahEffects
+      .filter((e) => e !== effect)
+      .map((e) => e.name.toLowerCase());
+    const validSubstrings = Array.from(
+      effect.name.toLowerCase().match(RegExp(/[\w ]+/g)) ?? []
+    ).filter((s) => !allOtherNoHookahEffectNames.some((name) => name.includes(s)));
+    validWishString.set(effect, validSubstrings[0]?.trim() ?? "");
+  }
+  return validWishString.get(effect) ?? "";
+}
```
horrible-little-slime, line 98: use `null` as your dummy value instead of the empty string

src/potions.ts
```diff
@@ -79,6 +82,24 @@ for (const effectGroup of mutuallyExclusiveList) {
...
+const validWishes = allNoHookahEffects.filter((effect) => {
+  return getValidWishString(effect)?.length > 0;
+});
```
horrible-little-slime, line 101:
```suggestion
const validWishes = allNoHookahEffects.filter((effect) => getValidWishString(effect));
```

src/potions.ts
```diff
@@ -460,18 +481,64 @@ const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+    new Potion($item`pocket wish`, {
+      effect,
+      canDouble: false,
+      duration: 20,
+      use: (quantity: number) => {
+        const s = getValidWishString(effect);
+        if (!s) return false;
+        return new Array(quantity).fill(0).every(() => {
+          cliExecute(`genie effect ${s}`);
+        });
```
horrible-little-slime, line 495: Use a `for` loop here

src/potions.ts
```diff
@@ -460,18 +481,64 @@ const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+      },
+    })
+);
+
+export const pawPotions = validWishes
   .map(
     (effect) =>
-      new Potion($item`pocket wish`, {
+      // eslint-disable-next-line libram/verify-constants
+      new Potion($item`cursed monkey's paw`, {
```
horrible-little-slime, line 504: You can bump the eslint-plugin-libram version for this

src/potions.ts
```diff
@@ -493,6 +560,20 @@ export function doublingPotions(embezzlers: number): Potion[] {
...
 }
 
+export function usePawWishes(singleUseValuation: (potion: Potion) => number): void {
+  while (get("_pawWishes", 0) < 5) {
+    // Sort the paw potions by the profits of a single wish, then use the best one
+    if (
+      !pawPotions
+        .filter((potion) => numericModifier(potion.effect(), "Meat Drop") >= 100)
+        .sort((a, b) => singleUseValuation(b) - singleUseValuation(a))
+        .some((potion) => potion.use(1))
```
horrible-little-slime, line 570: Don't put effectful code in this `if` statement

src/potions.ts
```diff
@@ -493,6 +560,20 @@ export function doublingPotions(embezzlers: number): Potion[] {
     .map((pair) => pair.potion);
 }
 
+export function usePawWishes(singleUseValuation: (potion: Potion) => number): void {
+  while (get("_pawWishes", 0) < 5) {
+    // Sort the paw potions by the profits of a single wish, then use the best one
+    if (
+      !pawPotions
+        .filter((potion) => numericModifier(potion.effect(), "Meat Drop") >= 100)
+        .sort((a, b) => singleUseValuation(b) - singleUseValuation(a))
```
horrible-little-slime, line 569: This will behave very poorly with `Infinity` in the array

src/potions.ts
```diff
@@ -493,6 +560,20 @@ export function doublingPotions(embezzlers: number): Potion[] {
     .map((pair) => pair.potion);
 }
 
+export function usePawWishes(singleUseValuation: (potion: Potion) => number): void {
+  while (get("_pawWishes", 0) < 5) {
+    // Sort the paw potions by the profits of a single wish, then use the best one
+    if (
+      !pawPotions
+        .filter((potion) => numericModifier(potion.effect(), "Meat Drop") >= 100)
```
horrible-little-slime, line 568: You already do this filter above; is this redundant?

src/potions.ts
```diff
@@ -493,6 +560,20 @@ export function doublingPotions(embezzlers: number): Potion[] {
...
+        .sort((a, b) => singleUseValuation(b) - singleUseValuation(a))
+        .some((potion) => potion.use(1))
+    ) {
+      break;
```
horrible-little-slime, line 572: use `return` instead of `break` if there's nothing after the `while` loop

horrible-little-slime (changes requested):

horrible-little-slime:
This also needs a kolmafia version bump

src/potions.ts
horrible-little-slime: Or, rather, not this far buried in the `if` statement--obviously you need to capture and evaluate the value somehow

src/potions.ts
```diff
@@ -79,6 +81,26 @@ for (const effectGroup of mutuallyExclusiveList) {
...
+    ).filter((s) => !allOtherNoHookahEffectNames.some((name) => name.includes(s)));
+    validWishString.set(effect, validSubstrings[0]?.trim());
+  }
+  return validWishString.get(effect);
+}
+const validWishes = allNoHookahEffects
+  .filter((effect) => getValidWishString(effect))
+  .filter(
+    (effect) => (globalOptions.prefs.yachtzeechain ? effect !== $effect`Eau d' Clochard` : true) // hardcoded heuristics
+  );
```
horrible-little-slime, line 102: These can/should probably be merged

horrible-little-slime:
```suggestion
  .filter((effect) => !!getValidWishString(effect) && (!globalOptions.prefs.yachtzeechain || effect !== $effect`Eau d' Clochard`));
```

you may or may not need the double bang here

src/potions.ts
```diff
@@ -460,18 +482,64 @@ const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
-export const wishPotions = Effect.all()
-  .filter((effect) => !effect.attributes.includes("nohookah"))
+export const wishPotions = validWishes.map(
+  (effect) =>
+    new Potion($item`pocket wish`, {
+      effect,
+      canDouble: false,
+      duration: 20,
+      use: (quantity: number) => {
+        const s = getValidWishString(effect);
```
horrible-little-slime, line 492: you don't need this for genie, afaik; I don't know why I didn't notice this earlier lol

src/potions.ts
```diff
@@ -460,18 +482,64 @@ const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+          }
+
+          const s = getValidWishString(effect);
+          if (!s) return false;
+
+          const currentEffectTurns = haveEffect(effect);
+          visitUrl("main.php?action=cmonk&pwd");
+          runChoice(1, `wish=${s}`);
+          visitUrl("main.php");
+          if (haveEffect(effect) <= currentEffectTurns) {
```
horrible-little-slime, line 534: imo it would be better practice to regexp the result of runChoice, but that can wait for libram

src/potions.ts
```diff
@@ -79,6 +81,26 @@ for (const effectGroup of mutuallyExclusiveList) {
   }
 }
 
+const allNoHookahEffects = Effect.all().filter((effect) => !effect.attributes.includes("nohookah"));
+const validWishString: Map<Effect, string> = new Map();
+function getValidWishString(effect: Effect): string | undefined {
```
horrible-little-slime, line 86: extremely funny thought: do we need to check these substrings against item names as well?

horrible-little-slime (changes requested):
This really feels like the last round of feedback! Great work!

src/potions.ts
horrible-little-slime: if you do need special logic for genie strings, it will be different from the special logic needed for monkey strings, because life is hell

@s-k-z knows a lot about this iirc, and may have an open libram PR that I should merge

src/potions.ts
horrible-little-slime: update:
we only care about if a substring is an exact match for an available item name. This will probably? never happen.
But!
```
const availableItems = [...new Set(Location.all().filter((l) => canAdventure(l)).map((l) => getMonsters(l)).filter((m) => m.copyable).map((m) => itemDropsArray(m)).flat())];
```
or something like that

Pantocyclus: I added this in, but we should also keep in mind that this is created at the instance of invoking garbo. We may gain (one-day tickets, transponder etc) and/or lose (content unlocker effects running out) access to locations as garbo runs, so this will then be a slightly incorrect list. 

(But I don't think we should be updating this every turn, so this is the best we can do)

horrible-little-slime (approved):

src/yachtzee/buffs.ts
```diff
@@ -67,6 +68,13 @@ export function yachtzeePotionProfits(potion: Potion, yachtzeeTurns: number): nu
   return yachtzeeValue + embezzlerValue + barfValue - potion.price(true);
 }
 
+const doublingValue = (potion: Potion, yachtzeeTurns: number) =>
+  Math.min(
+    potion.price(false),
+    yachtzeePotionProfits(potion.doubleDuration(), yachtzeeTurns) -
+      Math.max(0, yachtzeePotionProfits(potion, yachtzeeTurns))
+  );
+
```
horrible-little-slime, line 77:
```suggestion
const doublingValue = (yachtzeeTurns: number) => (potion: Potion) =>
  Math.min(
    potion.price(false),
    yachtzeePotionProfits(potion.doubleDuration(), yachtzeeTurns) -
      Math.max(0, yachtzeePotionProfits(potion, yachtzeeTurns))
  );
```

Would this look better as a curried function, letting us do `maxBy(doublingPotions, doublingValue(yachtzeeTurns))`?
This isn't actively a suggestion, just a pitch

Pantocyclus: This is the same as `partial()` in Python, so it makes sense. I just don't know whether it's a common stylistic thing in JS/TS; if so I can change it

horrible-little-slime (dismissed):

PR #1338, opened 2023-04-08
Don't wander at Professor Jacking's Huge-A-Ma-tron

s-k-z:
The Whole Kingdom combat resulting from the noncombat can't be instakilled and the noncombat is not skippable otherwise it seems.


horrible-little-slime:
This is maybe the funniest first world problem I've ever seen

PR #1341, opened 2023-04-14
abort when you're busy

horrible-little-slime:


docrostov (approved):
cute/good change, i like the language and logic looks good.

PR #1342, opened 2023-04-14
Don't run pygmies when one is sniffed

horrible-little-slime:


s-k-z (changes requested):
Mild optimization but probably want to reorder these with `!pygmySniffed()` last, since it's more values to check at once.

PR #1345, opened 2023-04-16
Stop June Cleaver overriding NEP quest acceptance

CheeseyPickle:


src/tasks/daily.ts
```diff
@@ -525,12 +525,22 @@ export const DailyTasks: Task[] = [
...
     completed: () => get("_questPartyFair") !== "unstarted",
     do: () => nepQuest(),
-    outfit: () =>
-      myInebriety() > inebrietyLimit() &&
-      have($item`Drunkula's wineglass`) &&
-      canEquip($item`Drunkula's wineglass`)
-        ? { offhand: $item`Drunkula's wineglass` }
-        : {},
+    outfit: (): OutfitSpec => {
+      const specs: OutfitSpec = { avoid: $items`` };
```
horrible-little-slime, line 529: this is equivalent to Item.all(); you can just use a regular empty array

src/tasks/daily.ts
```diff
@@ -525,12 +525,22 @@ export const DailyTasks: Task[] = [
...
+      ) {
+        specs.offhand = $item`Drunkula's wineglass`;
+      }
+
+      if (have($item`June cleaver`) && get("_juneCleaverFightsLeft") <= 0) {
+        specs.avoid?.concat($items`June cleaver`);
+      }
+
+      return specs;
+    },
```
horrible-little-slime, line 543: it would probably be easier to just add ```avoid: $items`June cleaver`, ``` to both specs in the existing code

src/tasks/daily.ts
```diff
@@ -525,12 +525,22 @@ export const DailyTasks: Task[] = [
...
+      if (
+        myInebriety() > inebrietyLimit() &&
+        have($item`Drunkula's wineglass`) &&
+        canEquip($item`Drunkula's wineglass`)
+      ) {
+        specs.offhand = $item`Drunkula's wineglass`;
+      }
+
+      if (have($item`June cleaver`) && get("_juneCleaverFightsLeft") <= 0) {
+        specs.avoid?.concat($items`June cleaver`);
```
horrible-little-slime, line 539: `concat` creates a new array rather than modifying an existing array and as such will not do what you want

horrible-little-slime (changes requested):

horrible-little-slime:
#1350

PR #1351, opened 2023-04-18
Set-location-to-barf

horrible-little-slime:
we generally shouldn't have effectful code in something like meatDrop, and this has the dangerous side effect of making it never think we're at barf mountain for things like sunglasses

### DIFF
--- a/src/outfit.ts
+++ b/src/outfit.ts
@@ -14,6 +14,7 @@ import {
   myInebriety,
   numericModifier,
   retrieveItem,
+  setLocation,
   toInt,
   toSlot,
   totalTurnsPlayed,
@@ -165,6 +166,7 @@ export function tryFillLatte(): boolean {
 
 export function meatOutfit(embezzlerUp: boolean, requirement?: Requirement, sea?: boolean): void {
   const equipMode = embezzlerUp ? "embezzler" : "barf";
+  if (equipMode === "barf") setLocation($location`Barf Mountain`);
   const bjornChoice = pickBjorn(equipMode);
 
   const parameters = [...(requirement?.maximizeParameters ?? []), "-tie"];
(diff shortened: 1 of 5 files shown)


src/potions.ts
```diff
@@ -211,7 +211,6 @@ export class Potion {
   }
 
   meatDrop(): number {
-    setLocation($location`Friar Ceremony Location`);
     return (
       getModifier("Meat Drop", this.effect()) +
       2 * (usingPurse() ? getModifier("Smithsness", this.effect()) : 0)
```
Pantocyclus, line 216: I feel like `meatDrop()` should be independent of what our previous adventured location was. We could always do a
```
const prevLocation = myLocation();
setLocation($location.none);
const computedMeatDrop = getModifier("Meat Drop", this.effect()) +
      2 * (usingPurse() ? getModifier("Smithsness", this.effect()) : 0);
setLocation(myLocation);
return computedMeatDrop;
```

We handle barf-specific buffs in mood, while potions is also used in yachtzees (including for nobarf). I also wouldn't want to start grabbing underwater meat% immediately after yachtzees.

src/outfit.ts
```diff
@@ -165,6 +166,7 @@ export function tryFillLatte(): boolean {
 
 export function meatOutfit(embezzlerUp: boolean, requirement?: Requirement, sea?: boolean): void {
   const equipMode = embezzlerUp ? "embezzler" : "barf";
+  if (equipMode === "barf") setLocation($location`Barf Mountain`);
```
Pantocyclus, line 169: We can also unset this before returning, which might be useful if we are doing any cache compute for whatever reasons.

src/outfit.ts
```diff
@@ -165,179 +165,182 @@ export function tryFillLatte(): boolean {
...
-        have($item`protonic accelerator pack`) &&
-        get("questPAGhost") === "unstarted" &&
-        get("nextParanormalActivity") <= totalTurnsPlayed() &&
-        !preventEquip.includes($item`protonic accelerator pack`)
-      ) {
-        forceEquip.push($item`protonic accelerator pack`);
-      }
-
-      if (have($item`mafia pointer finger ring`)) {
+  const location = embezzlerUp ? $location.none : $location`Barf Mountain`;
```
Pantocyclus, line 168: Should we have a default location for `sea`?

Pantocyclus:
Everything else seems fine

PR #1352, opened 2023-04-20
Try for more shadow artifact quests

CheeseyPickle:
If we're not yachtzee-chaining, we can use Clara's bell & stuff to force the Shadow realm NC to complete artifact quests!
Also, we can use free runs there once affinity runs out.

If anyone can figure out a reliable way to win against the bosses, maybe that would be better than the artifact quest, though that does burn shadow affinity.


CheeseyPickle:
I'm not quite sure which file using Bell & stuff to force the quest should go in. If anyone has insights it would be greatly appreciated :D

src/fights.ts
```diff
@@ -1428,6 +1428,62 @@ const freeFightSources = [
...
+
+      // Check NC-forcers
+      if (get("encountersUntilSRChoice") === 0) {
+        // NC is forced natively
+        return true;
+      }
+
+      if (have($item`Clara's bell`) && !get("_claraBellUsed")) {
+        return true;
+      }
```
Pantocyclus, line 1449: Note that we sometimes intentionally save the clara's bell for the volcano (if the best quest for the volcoino requires us to force an NC)

src/fights.ts
```diff
@@ -1428,6 +1428,62 @@ const freeFightSources = [
...
+      adv1(bestShadowRift(), -1, "");
+
+      if (questStep("questRufus") === 1) {
+        withChoice(1498, 1, () => use($item`closed-circuit pay phone`));
+      }
+
+      if (!have($effect`Shadow Affinity`) && get("encountersUntilSRChoice", 0) !== 0) {
+        setLocation($location`Friar Ceremony Location`); // Reset location to not affect mafia's item drop calculations
+      }
```
Pantocyclus, line 1483: Veracity recently added support for `setLocation($location.none)`. We should probably update this everywhere else in garbo - replacing everything is not necessarily within the scope of this PR, but at the very least this here should use that new feature

(and then we'll need to update index.ts to ensure that users have the latest mafia version as well)

src/fights.ts
```diff
@@ -1428,6 +1428,62 @@ const freeFightSources = [
     },
     true
   ),
+  // Use NC forcers to complete shadow quests
+  new FreeFight(
```
Pantocyclus, line 1432: Is this actually a `FreeFight`? I personally don't think this feature belongs in fights.ts - it should be handled either by potion.ts or mood.ts since it deals with obtaining buffs

Pantocyclus: Alternatively this should be handled directly in the free shadow rift fights - i.e. if the conditions below return `true`, then we should spike a free rift monster -> grab the buff -> spike a free rift monster -> grab the buff etc.

Notably, this requires that you have not obtained Shadow Affinity for the day, so it makes sense to have alternative handling of using spikes to obtain the buff (but you'd have to think through how we are going to make those spike fights free)

src/fights.ts
CheeseyPickle: The Volcano code runs in DailySetup, before free fights happen, so I think we're all good on this part.

src/fights.ts
CheeseyPickle: It's kind of weird, since we automatically obtain shadow affinity when we take the quest, we either have to do parka spikes with the buff active or after it's run out.

I'm thinking of just using spikes with shadow affinity turns, but there's probably a better solution.

src/fights.ts
```diff
@@ -1428,6 +1428,63 @@ const freeFightSources = [
...
+
+      if (have($item`Clara's bell`) && !globalOptions.clarasBellClaimed) {
+        return true;
+      }
+
+      if (have($item`Jurassic Parka`) && get("_spikolodonSpikeUses") < 5) {
+        return true;
+      }
+
+      return false; // Can't force any NCs for shadow waters
```
horrible-little-slime, line 1455: We should return a `number` here instead of a `boolean`

horrible-little-slime (commented):
I agree with Panto that I'm not sure this should be a `FreeFight`, although I disagree that it should be in potion or mood--instead, I'd add this to our existing Telephone `FreeFight`s.

src/fights.ts
CheeseyPickle: The previous shadow code returns booleans here too, so I think it's okay to keep them?

CheeseyPickle:
Probably needs a little testing, but this should all work.

src/fights.ts
```diff
@@ -1762,6 +1837,15 @@ const freeRunFightSources = [
...
   ),
+  // Try to accelerate the shadow nc, if you're able to do a quest
+  new FreeRunFight(
+    () =>
+      have($item`closed-circuit pay phone`) &&
+      get("rufusQuestType") !== "items" &&
+      !have($effect`Shadow Affinity`) &&
+      get("encountersUntilSRChoice") > 0,
+    (runSource: ActionSource) => garboAdventure(bestShadowRift(), runSource.macro)
+  ),
```
CheeseyPickle, line 1848: We tend to get shadow affinity when we grab shadow forest items earlier in dailyItems. Should I just remove this code?
```suggestion
```

CheeseyPickle, line 1848: Everything works as intended, except this code rarely runs as we grab shadow forest items in dailyItems, usually leaving us with turns of Shadow Affinity.

Not sure what I should do with this, since if you run garbo again after shadow affinity's run out, then this will trigger.
```suggestion
```

Pantocyclus:
I'm inclined to agree with Phred that, if nothing else, this should go into the existing telephone `FreeFight` instead of being its own `FreeFight`.

The existing telephone `FreeFight` already has well-defined conditions governing when we would execute it; there's little reason to create a new/additional one which only triggers on free rift fights

horrible-little-slime:
Another important element here is that free fights are executed in batches
instead of cycling through to the first available one (this will Some Day
change).

I'm going to sketch out some logic for this later, but:
 - if we really want it to, Clara's bell + cincho can go in the potion
code. This would probably be blocked by katarn's upcoming changes
 - spikolodon spikes should be, whenever possible, used against shadow
monsters for simplicity
 - we will want to check if we have a shadow lodestone at the start and at
the end of the sequence (end because we'll stop running them when we run
out of the shadow effect)
 - when we start doing bosses, probably all of this logic changes

src/fights.ts
```diff
@@ -1400,39 +1402,89 @@ const freeFightSources = [
   new FreeFight(
     () => {
       if (!have($item`closed-circuit pay phone`)) return false;
+      // Check if we have or can get Shadow Affinity
       if (have($effect`Shadow Affinity`)) return true;
-      if (get("_shadowAffinityToday")) return false;
+      if (!get("_shadowAffinityToday") && !ClosedCircuitPayphone.rufusTarget()) return true;
 
-      if (!ClosedCircuitPayphone.rufusTarget()) return true;
```
Pantocyclus, line 1406: The logic here cannot be reduced to the one that you've written. If `get("_shadowAffinityToday")` is `true`, we want to return `false` immediately.

In your combined case, `if (!get("_shadowAffinityToday") && !ClosedCircuitPayphone.rufusTarget())` does not return (`false`) when `get("_shadowAffinityToday")` is `true`, which could have us entering the `FreeFight` without the ability to obtain Shadow Affinity

src/fights.ts
```diff
@@ -1400,39 +1402,89 @@ const freeFightSources = [
...
       if (get("encountersUntilSRChoice") === 0) {
         // Target is either an artifact or a boss
         return true; // Get the artifact or kill the boss immediately for free
       }
-      return false; // We have to spend turns to get the artifact or kill the boss
+
+      // Consider forcing noncombats below:
+      if (globalOptions.prefs.yachtzeechain) return false; // NCs are better when yachtzeeing, probably
+      // TODO: With the KoL update, is there a function for checking if an NC is already forced?
+      if (have($item`Clara's bell`) && !globalOptions.clarasBellClaimed) {
```
Pantocyclus, line 1420: In line 2521 of `fights.ts` we have `if (use($item`Clara's bell`)) return true;` which sets `_claraBellUsed` to `true` without setting `globalOptions.clarasBellClaimed` to `true`. We should fix that too as we expand our reliance on this global option variable.

src/fights.ts
CheeseyPickle: That's intentional on my part. Even if we can't get shadow affinity, if we're able to force an NC we can complete a quest for free.

src/fights.ts
```diff
@@ -1400,39 +1402,89 @@ const freeFightSources = [
...
+          have($effect`Shadow Affinity`)
+        ) {
+          freeFightOutfit(new Requirement([], { forceEquip: $items`Jurassic Parka` }));
+          cliExecute("parka spikolodon");
+          const macro = Macro.skill($skill`Launch spikolodon spikes`).basicCombat();
+          // Adventure until we use spikes
+          const startingSpikes = get("_spikolodonSpikeUses");
+          do {
+            garboAdventureAuto(bestShadowRift(), macro);
+          } while (get("_spikolodonSpikeUses") === startingSpikes);
```
Pantocyclus, line 1459: I don't like this `do... while` loop. If something goes wrong (e.g. misalignment of the prefs) we could potentially burn every single turn here.

Imo, this should be a single `garboAdventureAuto()` call, and we should throw if we did not manage to spike like we intended.

src/fights.ts
```diff
@@ -1400,39 +1402,89 @@ const freeFightSources = [
...
         }
-        adv1(bestShadowRift(), -1, ""); // grab the NC
       }
 
+      // If you forced an NC, make sure you have a quest to do
+      if (ncForced && ClosedCircuitPayphone.have() && !ClosedCircuitPayphone.rufusTarget()) {
+        ClosedCircuitPayphone.chooseQuest(() => 2);
+      }
+
+      adv1(bestShadowRift(), -1, "");
```
Pantocyclus, line 1475: I'd like to rethink the structure of this.

It was previously much clearer which `adv1`s were for fighting the free rift mobs, and which `adv1`s were for grabbing the NCs. Both are now mixed into this particular `adv1` here, which feels weird. 

Especially because the rough structure I have in my mind is that we should be passing the Spikes macro here (when intended) - i.e. when we `adv1` to fight a free rift mob, we decide whether we should spike in this particular `adv1` (and then go on to handle the NC if we did decide to spike, which isn't too different from how we were previously handling the scheduled NC if it was coming up next).

Maybe Phred would have other ideas to chip in.

horrible-little-slime:
One other note is capturing the return value of use for Clara's bell
actually does nothing; it always returns true

src/fights.ts
horrible-little-slime: This is probably OOS of this PR, but we shouldn't use spikes (or any other NC forcer, for that matter) if you only have a few turns left before the NC; we should use them when your distance from it is maximal.

But spikes in particular, if you use them with 1 turn left you've achieved nothing.

src/fights.ts
horrible-little-slime: I think this will probably need to be sliced out into several functions, possibly in its own file for simplicity. 

The exact way we proceed will vary a lot by whether or not the `Adventure Modifiers` thing gets added to api.php in the near future. If we're expecting it any day now, I'd just wait until we can definitively track whether an NC is up.

src/fights.ts
CheeseyPickle: Erm... I'm pretty sure that already uses the global variable. It sets it on line 2520.

src/fights.ts
Pantocyclus: Let's do it like this.
```
() => {
      if (have($item`Rufus's shadow lodestone`)) setChoice(1500, 2);
      if (!get("_shadowAffinityToday") && !ClosedCircuitPayphone.rufusTarget()) {
        ClosedCircuitPayphone.chooseQuest(() => 2); // Choose an artifact (not supporting boss for now)
      }
      runShadowRiftTurn(); // was previously adv1(bestShadowRift(), -1, ""); 
      if (get("encountersUntilSRChoice", 0) === 0) {
        if (ClosedCircuitPayphone.have() && !ClosedCircuitPayphone.rufusTarget()) {
          ClosedCircuitPayphone.chooseQuest(() => 2);
        }
        adv1(bestShadowRift(), -1, ""); // grab the NC
      }

      if (questStep("questRufus") === 1) {
        withChoice(1498, 1, () => use($item`closed-circuit pay phone`));
      }

      if (!have($effect`Shadow Affinity`) && get("encountersUntilSRChoice") !== 0) {
        setLocation($location`Friar Ceremony Location`); // Reset location to not affect mafia's item drop calculations
      }
    },
```

We then have the function
```
function runShadowRiftTurn(): void { // we can probably have a better name
  if (get("encountersUntilSRChoice") === 0) return;
  if (globalOptions.prefs.yachtzeechain || get("rufusQuestType") === "items") {
    adv1(bestShadowRift(), -1, ""); // We shouldn't be using NC forcers
    return;
  }

  if (have($item`Clara's bell`) && !globalOptions.clarasBellClaimed) {
    globalOptions.clarasBellClaimed = true;
    use($item`Clara's bell`);
  } else if (have($item`Cincho de Mayo`) && useableCinch() >= 60) {
      equip($slot`acc3`, $item`Cincho de Mayo`);
      while (get("_cinchUsed", 0) > 40) {
        if (!freeRest()) throw new Error("We are out of free rests!");
      }
      useSkill($skill`Cincho: Fiesta Exit`);
      // we should probably re-equip whatever was previously on $slot`acc3` here
  } else if (
      have($item`Jurassic Parka`) &&
      get("_spikolodonSpikeUses") < 5 &&
      have($effect`Shadow Affinity`) &&
      get("encountersUntilSRChoice") >= 2
    ) {
      freeFightOutfit(new Requirement([], { forceEquip: $items`Jurassic Parka` }));
      cliExecute("parka spikolodon");
      const macro = Macro.skill($skill`Launch spikolodon spikes`).basicCombat();
      garboAdventureAuto(bestShadowRift(), macro);
  } else {
    adv1(bestShadowRift(), -1, ""); // We wanted to use NC forcers, but none are suitable now
  }
}
```

CheeseyPickle:
This works for now, though it might be worth waiting until we figure out how to check the new Adventure Modifiers section so we can see if an NC is forced instead of what I'm doing currently, which is just setting encountersUntilSRChoice to 0 every time we force an NC.

Pantocyclus:
I'm going to give this the preliminary thumbs up for now - it hasn't broken anything on my end, and I think it's a reasonable start. I'd still like to move some of the stuff out of FreeFights, but that can probably be done in future iterations.

I'll still leave it to phred to pull the trigger on this, especially since I'll be out of town for the next few weeks.

src/fights.ts
```diff
@@ -1387,25 +1390,41 @@ const freeFightSources = [
...
       }
       if (get("encountersUntilSRChoice") === 0) {
         // Target is either an artifact or a boss
         return true; // Get the artifact or kill the boss immediately for free
       }
-      return false; // We have to spend turns to get the artifact or kill the boss
+
+      // Consider forcing noncombats below:
+      if (globalOptions.prefs.yachtzeechain) return false; // NCs are better when yachtzeeing, probably
+      // TODO: With the KoL update, is there a function for checking if an NC is already forced?
```
docrostov, line 1407: this doesn't exist yet but in theory should exist soon, i guess? 

i don't think this needs to happen for initial adoption of this PR but [the way soolar parses the charpane for NC forces](https://github.com/Loathing-Associates-Scripting-Society/ChIT/blob/424aba089e87b7e66b4762d97a1fb2cd0246df0f/src/relay/charpane.ash#L1202) is a good example for how libram or garbo might do the same.

src/fights.ts
```diff
@@ -2525,3 +2558,41 @@ function yachtzee(): void {
     }
   }
 }
+
+function runShadowRiftTurn(): void {
+  // we can probably have a better name
+  if (get("encountersUntilSRChoice") === 0) return;
+  if (globalOptions.prefs.yachtzeechain || get("rufusQuestType") === "items") {
```
docrostov, line 2565: shouldn't it bypass the NC-forcer logic even if the rufus quest is the boss? given that garbo doesn't support the boss fights, i could see a small problem here where it would keep forcing a boss that garbo might not be able to kill if, for instance, the user came out of an ascension having accepted the boss fight quest. i doubt this came up in testing but i would just say != "artifact" instead of === items here

src/lib.ts
```diff
@@ -575,3 +577,31 @@ export function withLocation<T>(location: Location, action: () => T): T {
...
+    get("_cinchUsed", 0) +
+    sumNumbers(cinchRestored.slice(cinchRestsUsed, cinchRestsUsed + freeRestsLeft))
+  );
+}
+
+export function freeRest(): boolean {
+  if (get("timesRested") >= totalFreeRests()) return false;
+
+  if (get("chateauAvailable")) {
+    visitUrl("place.php?whichplace=chateau&action=chateau_restlabelfree");
```
docrostov, line 599: it might be good to have some logic here that evaluates if it is profitable to the user to swap to the free rest vs the adv ceiling fan in the chateau. i do not think this is needed for first implementation but maybe a comment here to note that it would be a nice-to-have would be good.

docrostov (approved):
i will approve because the logic overall looks sound and you've been testing it for a while... but my 2nd comment might actually create a frustrating break condition for garbo where a user having done something beforehand leads to garbo just getting its ass wholly kicked by shadow rift bosses, which would cause problems :tm:

src/lib.ts
horrible-little-slime: probably just anyone who owns cincho + a way to eke value from NC forcers should switch. the bigger issue is the Guy who gets mad we don't reinstall their antlers. I'll work on that later

src/fights.ts
horrible-little-slime: it's in api.php now

horrible-little-slime:
I'm gonna merge this with the knowledge that NC forcers being in api.php is about to flip our lives upside down. And then I'm going to immediately put in a PR to use new cincho support in libram.

PR #1353, opened 2023-04-21
add main.php hit before checking choice

ktpolanski:


Pantocyclus:
We should throw in a check for the designer sweatpants at the start of garbo as well (to realign the sweat if it got misaligned)
```visitUrl(`desc_item.php?whichitem=${$item`designer sweatpants`.descid}`);```

horrible-little-slime (approved):

PR #1355, opened 2023-04-22
retrieve_price() fix

Pantocyclus:
`retrieve_price()` has inconsistent behaviour with `mall_price()` and `npc_price()`.
Specifically, `retrieve_price()` returns infinity (or a very large number) when it deems the item unacquireable, while `mall_price()` and `npc_price()` return 0.

### DIFF
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -1033,7 +1033,8 @@ export function runDiet(): void {
     }
 
     MenuItem.defaultPriceFunction = (item: Item) => {
-      const itemRetrievePrice = retrievePrice(item);
+      const itemRetrievePrice =
+        retrievePrice(item) < Number.MAX_SAFE_INTEGER ? retrievePrice(item) : Infinity;
       const itemMallPrice = mallPrice(item) > 0 ? mallPrice(item) : Infinity;
       const itemNpcPrice = npcPrice(item) > 0 ? npcPrice(item) : Infinity;
       const bestPrice = Math.min(itemRetrievePrice, itemMallPrice, itemNpcPrice);
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -1034,8 +1034,10 @@ export function runDiet(): void {
 
     MenuItem.defaultPriceFunction = (item: Item) => {
       const itemRetrievePrice = retrievePrice(item);
+      const itemMallPrice = mallPrice(item) > 0 ? mallPrice(item) : Infinity;
       const itemNpcPrice = npcPrice(item) > 0 ? npcPrice(item) : Infinity;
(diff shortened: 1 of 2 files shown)


src/diet.ts
```diff
@@ -1033,7 +1034,10 @@ export function runDiet(): void {
 
     MenuItem.defaultPriceFunction = (item: Item) => {
       const itemRetrievePrice = retrievePrice(item);
-      return itemRetrievePrice > 0 ? itemRetrievePrice : item.tradeable ? mallPrice(item) : 0;
+      const itemMallPrice = mallPrice(item) > 0 ? mallPrice(item) : Infinity;
+      const itemNpcPrice = npcPrice(item) > 0 ? npcPrice(item) : Infinity;
+      const bestPrice = Math.min(itemRetrievePrice, itemMallPrice, itemNpcPrice);
+      return bestPrice === Infinity && !item.tradeable && have(item) ? 0 : bestPrice; // Handle unpurchaseable items like distention pills
```
horrible-little-slime, line 1040: I am (mildly) concerned that having unobtainable items with a price of `0` will result in bad behavior from the diet calculator; if you can point to a specific place where we prune that out I'll happily merge this now, but otherwise I'd like to wait for review from @pstalcup 

I know that right now there are situations where we have specific items with a price of 0, but I think a lot of those have bespoke limits in place (e.g., distention pill)

src/index.ts
```diff
@@ -240,6 +240,9 @@ export function main(argString = ""): void {
   const completedProperty = "_garboCompleted";
   set(completedProperty, "");
 
+  // re-align sweat (useful for diet and outfit)
+  visitUrl(`desc_item.php?whichitem=${$item`designer sweatpants`.descid}`);
```
horrible-little-slime, line 244:
```suggestion
 examine($item`designer sweatpants`);
```
and import `examine` from libram :)

src/diet.ts
```diff
@@ -1033,7 +1034,10 @@ export function runDiet(): void {
 
     MenuItem.defaultPriceFunction = (item: Item) => {
       const itemRetrievePrice = retrievePrice(item);
```
horrible-little-slime, line 1036:
```suggestion
      const itemRetrievePrice = retrievePrice(item) < Number.MAX_SAFE_INTEGER ? retrievePrice(item) : Infinity;
```

horrible-little-slime:
Alright I'm gonna go ahead and say this is fine if it's tested

Pantocyclus:
I haven't tested it specifically for Mayo, but it didn't blow up on non-mayo so I assume that's fine

PR #1359, opened 2023-04-24
`use` returns a useless value for the fishy pipe

horrible-little-slime:


docrostov (approved):
image

this bug was causing garbo to go "oh, great, fishy pipe is usable, that means i can enter the sea" and causing garbo to occasionally do sea turns without fishy, leading to -2 turns for the embezzie rather than the normal -1. hilarious.

PR #1362, opened 2023-04-26
Use known quest information to calculate rufus potion price

s-k-z:


src/potions.ts
```diff
@@ -476,10 +479,27 @@ export const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+      runChoice(6);
+      if (handlingChoice()) {
+        throw `Stuck in the closed-circuit pay phone dialogue? Please manually hang up and re-run`;
+      }
+      if (get("rufusDesiredItems") === "") {
+        throw `Could not determine the next closed-circuit payphone item quest. Try manually handing in a quest to continue.`;
+      }
+    }
+
+    const knownPrice = getPrice(Item.get(get("rufusDesiredItems")));
```
horrible-little-slime, line 499: Should I force libram to interpret this guy as an `Item | null`?

src/potions.ts
```diff
@@ -476,10 +479,27 @@ export const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+    const futureQuests = estimatedGarboTurns() / 30 - 1;
```
horrible-little-slime, line 500: does this need to account for any existing turns of shadow waters?

src/potions.ts
```diff
@@ -476,10 +479,27 @@ export const rufusPotion = new Potion($item`closed-circuit pay phone`, {
 
     // We consider the average price of the shadow items to not get gated behind an expensive one
     const shadowItems = $items`shadow brick, shadow ice, shadow sinew, shadow glass, shadow stick, shadow skin, shadow flame, shadow fluid, shadow sausage, shadow bread, shadow venom, shadow nectar`;
-    const averagePrice =
-      sum(shadowItems, (it) =>
-        historical && historicalAge(it) < 14 ? historicalPrice(it) : mallPrice(it)
-      ) / shadowItems.length;
+    const getPrice = (item: Item): number =>
+      3 * (historical && historicalAge(item) < 14 ? historicalPrice(item) : mallPrice(item));
+    const avgSpeculativePrice = sum(shadowItems, (it) => getPrice(it)) / shadowItems.length;
```
horrible-little-slime, line 484:
```suggestion
    const avgSpeculativePrice = sum(shadowItems, getPrice) / shadowItems.length;
```

src/potions.ts
```diff
@@ -476,10 +479,27 @@ export const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+
+    if (haveItemQuest) return avgSpeculativePrice;
+
+    if (get("rufusDesiredItems") === "") {
+      directlyUse($item`closed-circuit pay phone`);
```
horrible-little-slime, line 489: `ClosedCircuitPayphone.chooseQuest`  lets you dynamically choose the quest using a function that takes, among other things, the Item it's looking for as an input; do we want to use that here, or does it not work?

src/potions.ts
```diff
@@ -476,10 +479,27 @@ export const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+      runChoice(6);
```
horrible-little-slime, line 490: Is this the right choice option? 6 feels like it's right for a shadow labyrinth, but I don't think there's any rufus dialog or dialogue with that many options

src/potions.ts
horrible-little-slime: if it doesn't work for our purposes it doesn't work for our purposes, I just think it would be nice

src/potions.ts
s-k-z: Yeah, I'll add that.

src/potions.ts
```diff
@@ -476,10 +479,27 @@ export const rufusPotion = new Potion($item`closed-circuit pay phone`, {
 
     // We consider the average price of the shadow items to not get gated behind an expensive one
     const shadowItems = $items`shadow brick, shadow ice, shadow sinew, shadow glass, shadow stick, shadow skin, shadow flame, shadow fluid, shadow sausage, shadow bread, shadow venom, shadow nectar`;
-    const averagePrice =
-      sum(shadowItems, (it) =>
-        historical && historicalAge(it) < 14 ? historicalPrice(it) : mallPrice(it)
-      ) / shadowItems.length;
+    const getPrice = (item: Item): number =>
+      3 * (historical && historicalAge(item) < 14 ? historicalPrice(item) : mallPrice(item));
```
horrible-little-slime, line 483: there are few enough shadow items that we can probably get away with just using mallPrice, but that's a personal choice best left to you

src/potions.ts
s-k-z: I do not want to choose a quest for this under any circumstances.

src/potions.ts
s-k-z: > js availableChoiceOptions()

Returned: aggregate string [int]
1 => I'll fight the entity
2 => I'll find the artifact
3 => I'll collect the goods
6 => Hang up

src/potions.ts
s-k-z: Yeah, I think so.

src/potions.ts
s-k-z: Seems fine as is.

src/potions.ts
s-k-z: To elaborate on the why, it would lock us into a quest which we can otherwise reroll the next day.

src/potions.ts
horrible-little-slime: update: per discussion in discord, item properties don't exist yet so this will be a "in a long while from now" change

src/potions.ts
```diff
@@ -476,10 +479,27 @@ export const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+    const avgSpeculativePrice = sum(shadowItems, getPrice) / shadowItems.length;
+
+    if (haveItemQuest) return avgSpeculativePrice;
+
+    if (get("rufusDesiredItems") === "") {
+      directlyUse($item`closed-circuit pay phone`);
+      runChoice(6);
```
horrible-little-slime, line 490: would it make more sense to do `withChoice(1497, 6, () => use(the phone)`?

s-k-z: Oh, yeah I like that more.

src/potions.ts
```diff
@@ -476,10 +477,26 @@ export const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
-      ) / shadowItems.length;
+    const getPrice = (item: Item): number =>
+      3 * (historical && historicalAge(item) < 14 ? historicalPrice(item) : mallPrice(item));
+    const avgSpeculativePrice = sum(shadowItems, getPrice) / shadowItems.length;
+
+    if (haveItemQuest) return avgSpeculativePrice;
+
+    if (get("rufusDesiredItems") === "") {
+      withChoice(1497, 6, () => use($item`closed-circuit pay phone`));
+      if (handlingChoice()) {
```
horrible-little-slime, line 488: might be worth a visitUrl to main.php before deciding that we're stuck?
I guess wait until that comes up during testing before throwing a potentially-unnecessary pagehit in there

src/potions.ts
```diff
@@ -476,10 +477,26 @@ export const rufusPotion = new Potion($item`closed-circuit pay phone`, {
...
+        throw `Stuck in the closed-circuit pay phone dialogue? Please manually hang up and re-run`;
```
horrible-little-slime, line 489: I think we generally prefer to throw Error objects instead of strings, although I'm not entirely if there's a compelling reason for that convention.

src/potions.ts
s-k-z: Oh yeah this was preliminary testing logic. I was planning on removing it after.

s-k-z:
A real solution is more complicated, needs changing the behavior of potions in general so I'm going to close this for now.

PR #1365, opened 2023-04-30
Allow using a custom Combat Macro for Shadow Rift

eegeeZA:
Configure this if the default garbo CCS is failing to kill monsters in the Shadow Rift. TheKolWiki has details about how to write a macro: https://kol.coldfront.net/thekolwiki/index.php/Combat_Macros


horrible-little-slime:
Superceded by #1367; we try not to require that users configure their own combat settings, and so the more elegant solution here is simply to skip LTS and regular attacks when in the shadow rifts.

PR #1369, opened 2023-04-30
outfit rehaul

horrible-little-slime:


src/outfit/barf.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  const bjornChoice = chooseBjorn(BonusEquipMode.BARF, outfit.familiar, sim);
+
+  outfit.modifier.push(
+    `${valueOfMeat(BonusEquipMode.BARF)} Meat Drop`,
+    `${valueOfItem(BonusEquipMode.BARF)} Item Drop`,
+    "-tie"
+  );
+
+  if (myInebriety() > inebrietyLimit()) {
```
Kasekopf, line 110: I think this needs to account for Stooper's contribution to inebriety_limit

src/outfit/embezzler.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const underwater = target.environment === "underwater";
+  if (underwater) {
+    if (!outfit.familiar.underwater) {
+      const familiarEquip = familiarWaterBreathingEquipment.find((item) => have(item));
+      if (familiarEquip) outfit.equip(familiarEquip);
+    }
+
+    const airEquip = waterBreathingEquipment.find((item) => have(item) && canEquip(item));
+    if (!airEquip || !outfit.equip(airEquip)) outfit.modifier.push("sea");
```
Kasekopf, line 39: Since the different equipments could be in different slots (hat or back), it might be better to try equipping more than the first canEquip-able item. You could consider using `outfit.equipFirst`.

src/outfit/embezzler.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  useUPCsIfNeeded(outfit);
```
Kasekopf, line 42: This and `validateGarbageFoldable` both make changes to the character state. This is fine for now as long as every constructed outfit is followed by a dress. But in the future you may consider separating this away from the construction of the Outfit object. validateGarbageFoldable will be handled by a fix for grimoire#90; I'll think about the sticker items.

src/outfit/embezzler.ts
```diff
@@ -0,0 +1,69 @@
...
+    outfit.setBonus(bjornalike, bjornChoice.value);
+    outfit.equip(bjornalike);
+    const other = $items`Buddy Bjorn, Crown of Thrones`.filter((i) => i !== bjornalike)[0];
+    outfit.avoid.push(other);
+    switch (bjornalike) {
+      case $item`Buddy Bjorn`:
+        outfit.bjornify(bjornChoice.familiar);
+        break;
+      case $item`Crown of Thrones`:
+        outfit.bjornify(bjornChoice.familiar);
```
Kasekopf, line 57: `outfit.enthrone`, instead

src/outfit/barf.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  const gun = chooseGun(outfit);
+  if (!gun) return { available: false, items: [] };
+
+  return {
+    available: true,
+    items: {
+      back: $item`unwrapped knock-off retro superhero cape`,
+      equip: [gun],
```
Kasekopf, line 57: The wiki says that the gun needs to be in the mainhand. But I see the old garbo code also didn't specify this so maybe that is incorrect.

src/fights.ts
```diff
@@ -416,13 +425,12 @@ export function dailyFights(): void {
...
-                forceEquip: $items`Pocket Professor memory chip`,
-              })
-            : new Requirement(maximizeParameters, {});
+          const profSpec = {
+            familiar: $familiar`Pocket Professor`,
+            equip: $items`Pocket Professor memory chip`.filter((i) => have(i)),
+          };
 
-          goalMaximize(Requirement.merge([professorRequirement, ...fightSource.requirements]));
+          goalMaximize({ ...profSpec, ...fightSource.spec });
```
Kasekopf, line 433: You could potentially run into some issues merging specs this way. For example if any of the fightSource has an `equip: Item[]` set it will just override the memory chip in the professor spec.

I should add an OutfitSpec merge to grimoire...

src/outfit/barf.ts
horrible-little-slime: I think it's mostly that we are exceedingly unlikely to dual wield ranged weapons. but I'll make it a weapon

Rinn (approved):

src/outfit/free.ts
```diff
@@ -0,0 +1,60 @@
...
+      ? "Familiar Experience"
+      : "Familiar Weight"
+  );
+
+  const bjornChoice = chooseBjorn(mode, outfit.familiar);
+
+  if (get("_vampyreCloakeFormUses") < 10) outfit.equip($item`vampyric cloake`);
+  outfit.bonuses = bonusGear(mode);
+
+  if (outfit.familiar !== $familiar`Grey Goose`) outfit.setBonus($item`tiny stillsuit`, 500);
```
s-k-z, line 37: Shouldn't the pocket professor also be excluded? I think the familiar experience is more useful for it.

PR #1372, opened 2023-05-01
Upper chamber has no NCs after finishing

Shiverwarp:
The NC no longer shows up after defeating Ed, but mafia will still show the combat rate being 85.


horrible-little-slime:
Feels like this could also be a mafia PR

PR #1374, opened 2023-05-02
Check buffed Muscle before using Powerful Glove and prefer Moleman for Thesis

eegeeZA:
Idea: consider using `molehill mountain` since it's a free fight

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -26,6 +26,7 @@ import {
   Monster,
   myAdventures,
   myAscensions,
+  myBuffedstat,
   myClass,
   myFamiliar,
   myInebriety,
@@ -699,6 +700,10 @@ function bowlOfScorpionsAvailable() {
   }
 }
 
+function molemanReady() {
+  return have($item`molehill mountain`) && !get("_molehillMountainUsed");
+}
+
 const freeFightSources = [
   new FreeFight(
     () =>
@@ -716,7 +721,7 @@ const freeFightSources = [
     }
   ),
   new FreeFight(
-    () => (have($item`molehill mountain`) && !get("_molehillMountainUsed") ? 1 : 0),
+    () => (thesisReady() ? 0 : molemanReady() ? 1 : 0),
     () => withMacro(Macro.basicCombat(), () => use($item`molehill mountain`)),
     true
   ),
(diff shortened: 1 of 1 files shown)


src/fights.ts
```diff
@@ -2096,6 +2097,7 @@ export function deliverThesisIfAble(): void {
   safeRestore();
 
   if (
+    myBuffedstat($stat`Muscle`) < 1800 &&
```
eegeeZA, line 2100: Matches the muscle stat lower down. Should I extract to a local `const`?
https://github.com/Loathing-Associates-Scripting-Society/garbage-collector/blob/acf0323471e4fd54c6e028db77faa0718c839426/src/fights.ts#L2110

horrible-little-slime: yeah, good shout

horrible-little-slime:
does molehill mountain scale high enough?

we should maximize muscle before we check for triple-sized as well

src/fights.ts
eegeeZA: Cool, going to refactor a lot more. Based on the comment about maximising muscle before triple-sized.

eegeeZA:
Based on my calc's moleman does scale and required lower Muscle. Check my calc and let me know:
`(849+15) * 1.5 = 1296 = min(2 * floor(1296^0.25), 11)`

docrostov:
your calculation looks right, molehill mountain should theoretically work

eegeeZA:
The code will now prioritise fighting Moleman if available. The required muscle has also been optimised to only require the minimum, depending on which monster will be fought. And lastly, the Powerful Glove will not be used if one already has enough buffed Muscle.

horrible-little-slime:
Gotta delete that unused import

eegeeZA:
> Gotta delete that unused import

Sorted. Apologies for all the quick successive pushes.

src/fights.ts
```diff
@@ -716,7 +721,7 @@ const freeFightSources = [
     }
   ),
   new FreeFight(
-    () => (have($item`molehill mountain`) && !get("_molehillMountainUsed") ? 1 : 0),
+    () => (molemanReady() && !thesisReady() ? 1 : 0),
```
horrible-little-slime, line 724: Based on the way thesisReady works, people who don't own a prof will never fight the moleman.

eegeeZA: You're quite right. I've adjusted the logic.

eegeeZA: Hang on, isn't `(thesisReady() ? 0 : molemanReady() ? 1 : 0)` the same as `(molemanReady() && !thesisReady() ? 1 : 0)`? How will people without a Prof be blocked from the fight?
| Thesis | Moleman  | desired |
|--------|-------|---------|
| TRUE   | TRUE  | 0       |
| TRUE   | FALSE | 0       |
| FALSE  | TRUE  | 1       |
| FALSE  | FALSE | 0       |

horrible-little-slime (approved):

src/fights.ts
horrible-little-slime: oh, I just straight-up misread that then. You're right!

eegeeZA: Hehe cool. The refactored version probably reads a bit easier. Thanks for the approve.

PR #1375, opened 2023-05-02
Set value of workshed items to 0 for session valuing

s-k-z:


horrible-little-slime (approved):
Long-term, I think it would be nice to add `getCampground()` or something analogous to libram's sesssion, but this is fine for now

PR #1381, opened 2023-05-03
Handle acquiring a lodestone on the final shadow rift fight.

horrible-little-slime:


src/fights.ts
```diff
@@ -1406,7 +1406,7 @@ const freeFightSources = [
       return false; // We have to spend turns to get the artifact or kill the boss
     },
     () => {
-      if (have($item`Rufus's shadow lodestone`)) setChoice(1500, 2);
+      setChoice(1500, 2);
```
docrostov, line 1409: i think this looks fine-ish but i'm not sure i understand why exactly the have needs to be removed here? choice 1500 is the lodestone choice, there shouldn't possibly be a time where the user enters a shadow rift and encounters 1500 without the have being true

src/fights.ts
```diff
@@ -1423,6 +1423,10 @@ const freeFightSources = [
         withChoice(1498, 1, () => use($item`closed-circuit pay phone`));
       }
 
+      if (have($item`Rufus's shadow lodestone`)) {
+        adv1(bestShadowRift(), -1, "");
```
docrostov, line 1427: are we just assuming the user never does the forest items choice? imo it probably would be worth doing once a day but it's fine if we say "nah" since mafia doesn't do a good job handling "3 if present otherwise 2"

docrostov (approved):
this looks fine but please reply to my comments ty friend

src/fights.ts
horrible-little-slime: we handle the forest choice elsewhere

src/fights.ts
horrible-little-slime: It's because we have multiple adv1s in this single codeblock; you might not start with a lodestone, but that doesn't mean you won't have one later

horrible-little-slime:
I will probably let the utility of this get absorbed into #1352

PR #1387, opened 2023-05-08
only abandon quests we can abandon; consider the price of guzzlr booze

horrible-little-slime:


src/wanderer/guzzlr.ts
```diff
@@ -78,24 +79,31 @@ export function guzzlrFactory(
...
-          if (!guzzlrBooze) {
-            // this is an error state - accepted a guzzlr quest but mafia doesn't know the booze
-            return false;
-          }
+      return guzzlrBooze
+        ? [
+            new WandererTarget(
+              "Guzzlr",
+              location,
+              guzzlrValue(Guzzlr.getTier()) - mallPrice(guzzlrBooze),
```
docrostov, line 87: does garbo's mall price consider price of components vs price of the full one? i assume it does i am just curious

src/wanderer/guzzlr.ts
```diff
@@ -21,11 +21,12 @@ function considerAbandon(locationSkiplist: Location[]) {
...
   if (
+    Guzzlr.canAbandon() &&
     // consider abandoning
-    !location || // if mafia faled to track the location correctly
-    locationSkiplist.includes(location) ||
-    !canAdventureOrUnlock(location) || // or the zone is marked as "generally cannot adv"
-    (globalOptions.ascend && wandererTurnsAvailableToday(location) < remaningTurns) // or ascending and not enough turns to finish
+    (!location || // if mafia failed to track the location correctly
+      locationSkiplist.includes(location) ||
+      !canAdventureOrUnlock(location) || // or the zone is marked as "generally cannot adv"
```
docrostov, line 28: i do not think this is necessarily needed in this PR, but -- regarding the abandonment of quests, i think there is a case to also calculate, via # of turns to complete the quest, how many of the content unlockers the user would have to buy in order to do the premium quest and using it to help solve this one. my recommendation would be to store this value and subtract it in to the wanderer value. canAdventureOrUnlock (in theory?) should make absinthe and transponder zones as plausibly available but the cost of those should be incorporated in garbo's decision of which wanderer source to use.

docrostov (approved):
looks basically fine for what you want to do in this PR. i added one small "hey, maybe we should incorporate this too" thing but you can also just unilaterally say "nah that's too annoying" and that is OK too

horrible-little-slime:
Unfortunately that's much harder than it seems; backups get used densely
for the most part but are sometimes used to reset extro. Digitized are used
at set intervals, as are voters and kramco, but the confluence of all 3
makes math hard. Throw in free for all and a few other sources and you've
got yourself a nightmare stew.

src/wanderer/guzzlr.ts
horrible-little-slime: Unfortunately that's much harder than it seems; backups get used densely for the most part but are sometimes used to reset extro. Digitizes are used at set intervals, as are voters and kramco, but the confluence of all 3 makes math hard. Throw in free for all and a few other sources and you've got yourself a nightmare stew.

PR #1389, opened 2023-05-10
don't wander in other wormwood zones either

Rinn:
#1385 ++


docrostov (approved):
thank you for adding the rest here, it is good to enforce skip conditions on these. approved/merged.

PR #1392, opened 2023-05-15
1st phase Chibibuddy implementation

Shiverwarp:
This just picks up the ChibiChanged buff (5 fam weight) during our daily tasks if it's available.

It does not do any caretaking of the Chibibuddy, so it'll only last a few days, but even if it just died it still gives you 5 turns of the buff.

There isn't any mafia support of chibibuddy currently, so this probably isn't very performant. It's been working good for me during eternal aftercore though.

Thanks to phred for the names ([Discord post](https://discord.com/channels/466605739838930955/864070665615179786/1042932371902693476)) but if it's too "cute" we can use some generic name? Or possibly just use the auto-generated name with an empty field.

This won't pick up the buff if you already had it from manually doing it pre-rollover, but I couldn't come up with a better `completed:` without mafia support

### DIFF
--- a/src/tasks/daily.ts
+++ b/src/tasks/daily.ts
@@ -192,8 +192,6 @@ function chibiSetup(): void {
       runChoice(7);
     }
   }
-  // Mafia does not currently see the item change if our buddy died
-  cliExecute("refresh inventory");
   // If we only have an (off) chibibuddy (whether from the beginning, or we just turned it off)
   // Turn it back on and confirm availability of the buff
   if (have($item`ChibiBuddy™ (off)`) && !have($item`ChibiBuddy™ (on)`)) {
@@ -400,8 +398,10 @@ export function configureSnojo(): void {
 export const DailyTasks: Task[] = [
   {
     name: "Chibi Buddy",
-    ready: () => have($item`ChibiBuddy™ (on)`) || have($item`ChibiBuddy™ (off)`),
-    completed: () => have($effect`ChibiChanged™`),
+    ready: () =>
+      get("_chibiChanged", true) &&
+      (have($item`ChibiBuddy™ (on)`) || have($item`ChibiBuddy™ (off)`)),
+    completed: () => get("_chibiChanged", true),
     do: chibiSetup,
     limit: { soft: 1 },
   },
--- a/src/tasks/daily.ts
+++ b/src/tasks/daily.ts
@@ -185,6 +185,7 @@ function chibiSetup(): void {
       // This is the choice option for chibi chat
       if (availableChoiceOptions()[5]) {
         runChoice(5);
+        runChoice(7);
         return;
       }
       // Exit the choice, if our buddy died, this will give us back a buddy (off)
@@ -206,7 +207,6 @@ function chibiSetup(): void {
     if (handlingChoice()) {
       if (availableChoiceOptions()[5]) {
         runChoice(5);
-        return;
       }
       runChoice(7);
     }
--- a/src/tasks/daily.ts
+++ b/src/tasks/daily.ts
@@ -1,7 +1,6 @@
 import { Task } from "grimoire-kolmafia";
 import {
   adv1,
-  availableChoiceOptions,
   canadiaAvailable,
   canAdventure,
   canEquip,
@@ -44,7 +43,6 @@ import {
   $monster,
   $slot,
   BeachComb,
-  directlyUse,
   findLeprechaunMultiplier,
   get,
   getModifier,
--- a/src/tasks/daily.ts
+++ b/src/tasks/daily.ts
@@ -170,37 +170,12 @@ function chibiBuffAvailable(): boolean {
   if (!have($item`ChibiBuddy™ (on)`) && !have($item`ChibiBuddy™ (off)`)) return false;
   // We need to name our buddy when we turn it on!
   const chibiNames = [
-    "Oliver",
-    "Neil",
-    "William",
-    "Guillermo",
-    "Aaron",
-    "Isobel",
-    "Pope Gregory",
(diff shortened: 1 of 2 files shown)


src/tasks/daily.ts
```diff
@@ -164,6 +166,53 @@ function voterSetup(): void {
...
 
+function chibiSetup(): void {
+  if (!have($item`ChibiBuddy™ (on)`) && !have($item`ChibiBuddy™ (off)`)) return;
+  // We need to name our buddy when we turn it on!
+  const chibiNames = [
+    "SSBBHax",
+    "Frasier Crane",
+    "William",
+    "Gamuel Sauce",
+    "Axis Shadowbaenimus",
```
horrible-little-slime, line 177: Interesting pick

src/tasks/daily.ts
```diff
@@ -164,6 +166,53 @@ function voterSetup(): void {
...
+    "Dale Cooper",
+  ];
+  // It is possible our Buddy died after rollover, and will turn to (off) after using it
+  // It is also possible we ascended with a Buddy(on) in which case it will reset immediately without the choice adventure on use
+  if (have($item`ChibiBuddy™ (on)`)) {
+    directlyUse($item`ChibiBuddy™ (on)`);
+    if (handlingChoice()) {
+      // This is the choice option for chibi chat
+      if (availableChoiceOptions()[5]) {
```
horrible-little-slime, line 186: this may(?) read nicer with the `in` operator. That or you can check `Object.values(availableChoiceOptions()).includes(` if that reads nicer

src/tasks/daily.ts
```diff
@@ -349,6 +398,12 @@ export function configureSnojo(): void {
 }
 
 export const DailyTasks: Task[] = [
+  {
+    name: "Chibi Buddy",
+    ready: () => have($item`ChibiBuddy™ (on)`) || have($item`ChibiBuddy™ (off)`),
+    completed: () => have($effect`ChibiChanged™`),
+    do: () => chibiSetup(),
```
horrible-little-slime, line 405:
```suggestion
    do: chibiSetup,
```

src/tasks/daily.ts
```diff
@@ -349,6 +398,12 @@ export function configureSnojo(): void {
 }
 
 export const DailyTasks: Task[] = [
+  {
+    name: "Chibi Buddy",
+    ready: () => have($item`ChibiBuddy™ (on)`) || have($item`ChibiBuddy™ (off)`),
+    completed: () => have($effect`ChibiChanged™`),
```
horrible-little-slime, line 404: I would use a `let` variable like you can see for `refreshedLatte` and a few other tasks; I don't think relying on the automatically-imposed limit feels particularly good. We want to be able to change or modify that limit without going through every task and deciding which one we want

Alternately, you could define a `limit` in this task itself

src/tasks/daily.ts
Shiverwarp: Should I change others to use this format instead as well? For example:

`  {
    name: "Check Barf Mountain Quest",
    ready: () => get("stenchAirportAlways") || get("_stenchAirportToday"),
    completed: () => !attemptCompletingBarfQuest,
    do: () => checkBarfQuest(),
  },`

Would instead just be `do: checkBarfQuest,`?

src/tasks/daily.ts
Shiverwarp: I like adding a specific limit, so did that.

src/tasks/daily.ts
Shiverwarp: I'll admit my picks were almost completely random. Very open to suggestions ahaha

src/tasks/daily.ts
Shiverwarp: I personally liked the way this read, we're seeing the choice options, and whether 5 exists, so I feel it's both succinct and clear?

I could certainly be wrong about that though

src/tasks/daily.ts
horrible-little-slime: ye

Shiverwarp: Alright, I just did these updates in this PR as well, hopefully that is fine and not too distasteful

Shiverwarp:
Okay, using the new mafia pref now. Using `get `with a default to true until libram changes so we don't keep trying

I'll be able to test tomorrow, but it'll be a day or two after that where I can check whether things still work fine if buddy dies after rollover

src/tasks/daily.ts
```diff
@@ -349,6 +396,15 @@ export function configureSnojo(): void {
 }
 
 export const DailyTasks: Task[] = [
+  {
+    name: "Chibi Buddy",
+    ready: () =>
```
s-k-z, line 401:
```suggestion
    ready: () => have($item`ChibiBuddy™ (on)`) || have($item`ChibiBuddy™ (off)`),
```
Don't need to check the completion criteria twice.

src/tasks/daily.ts
```diff
@@ -349,6 +396,15 @@ export function configureSnojo(): void {
...
+      !get("_chibiChanged", true) &&
+      (have($item`ChibiBuddy™ (on)`) || have($item`ChibiBuddy™ (off)`)),
```
s-k-z, line 403:
```suggestion
    ready: () =>have($item`ChibiBuddy™ (on)`) || have($item`ChibiBuddy™ (off)`),
```
Don't need to double up on the completion criteria

s-k-z (changes requested):
Need to increase sinceKolmafiaRevision to 27360. Other than that I think it's all set

s-k-z (approved):

PR #1393, opened 2023-05-19
don't spam up the CLI with checkpoints

horrible-little-slime:


docrostov (approved):
moving this after the return is a good idea :tm:

PR #1401, opened 2023-05-20
run yarn upgrade

Rinn:
been a bit since we upgraded dependencies


horrible-little-slime (approved):
If this builds and runs, seems like a good choice

PR #1404, opened 2023-05-21
Don't value ChibiBuddy™

s-k-z:


horrible-little-slime (commented):
We can cut the workshed code by upgrading libram, and then either give chibibuddy (off) the same valuation as chibibuddy (on) or set up an `exclusions` set anyways

s-k-z:
Would be better to do that in libram session handling on a later date.

s-k-z:
Based on the checksum, it looks like libram was already on 0.8.1 just not actually mandated.

PR #1405, opened 2023-05-21
Change pizza and calzone of legend into regular diet options & don't prevent pilsners and pizza until ascending

s-k-z:
Benefits: generally it's better to consume these items immediately coming out of an ascension rather than at a later date.

Drawbacks: users whose preferences are out of sync will try and fail to eat the pizzas resulting in at least one incorrect diet calculation.

### DIFF
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -84,6 +84,7 @@ import { Potion, PotionTier } from "./potions";
 import { garboValue } from "./session";
 import synthesize from "./synthesis";
 import { estimatedGarboTurns } from "./turns";
+import { BooleanProperty } from "libram/dist/propertyTypes";
 
 const MPA = get("valueOfAdventure");
 print(`Using adventure value ${MPA}.`, HIGHLIGHT);
@@ -333,6 +334,13 @@ function menu(): MenuItem<Note>[] {
     (item) => new MenuItem<Note>(item, { maximum: availableAmount(item) })
   );
 
+  const legendaryPizzas = [
+    { pizza: $item`Pizza of Legend`, pref: "pizzaOfLegendEaten" },
+    { pizza: $item`Calzone of Legend`, pref: "calzoneOfLegendEaten" },
+  ]
+    .filter(({ pref }) => !get(pref))
+    .map(({ pizza }) => new MenuItem(pizza));
+
   return [
     // FOOD
     new MenuItem($item`Dreadsylvanian cold pocket`),
@@ -348,6 +356,7 @@ function menu(): MenuItem<Note>[] {
     new MenuItem(mallMin(lasagnas)),
     new MenuItem(mallMin(smallEpics)),
     new MenuItem($item`green hamhock`),
+    ...legendaryPizzas,
 
     // BOOZE
     new MenuItem($item`elemental caipiroska`),
@@ -589,9 +598,8 @@ export function potionMenu(
     ? potion($item`Boris's bread`, { price: 2 * ingredientCost($item`Yeast of Boris`) })
     : [];
 
-  // Replace string with BooleanProperty later
-  const ofLegendPotion = (item: Item, prefName: string) => {
-    if (get(prefName, true)) return [];
+  const ofLegendPotion = (item: Item, pref: BooleanProperty) => {
+    if (get(pref)) return [];
 
     const recipes = [
       item,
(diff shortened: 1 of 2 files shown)


src/diet.ts
```diff
@@ -84,6 +84,7 @@ import { Potion, PotionTier } from "./potions";
 import { garboValue } from "./session";
 import synthesize from "./synthesis";
 import { estimatedGarboTurns } from "./turns";
+import { BooleanProperty } from "libram/dist/propertyTypes";
```
horrible-little-slime, line 87: in libram 8.1 you should be able to import this directly; be sure to `yarn install` if you can't

src/diet.ts
```diff
@@ -631,7 +631,7 @@ export function potionMenu(
     ...campfireHotdog,
     ...foodCone,
     ...borisBread,
-    ...ofLegendMenuItems,
+    ...ofLegendPotion($item`Deep Dish of Legend`, "deepDishOfLegendEaten"),
```
horrible-little-slime, line 634: probably not worth having a bespoke function that's called in only one spot

src/diet.ts
```diff
@@ -328,11 +329,18 @@ function menu(): MenuItem<Note>[] {
...
+  const pilsners = $items`astral pilsner`.filter((item) => have(item));
   const limitedItems = [...boxingDayCareItems, ...pilsners].map(
     (item) => new MenuItem<Note>(item, { maximum: availableAmount(item) })
   );
 
+  const legendaryPizzas = [
+    { pizza: $item`Pizza of Legend`, pref: "pizzaOfLegendEaten" },
+    { pizza: $item`Calzone of Legend`, pref: "calzoneOfLegendEaten" },
+  ]
+    .filter(({ pref }) => !get(pref))
```
horrible-little-slime, line 341: we want to check whether or not you have all the recipes, much like we do in the `ofLegendPotion` function

src/diet.ts
```diff
@@ -589,31 +608,18 @@ export function potionMenu(
...
-    ? [
-        ...ofLegendPotion($item`Calzone of Legend`, "calzoneOfLegendEaten"),
-        ...ofLegendPotion($item`Pizza of Legend`, "pizzaOfLegendEaten"),
-        ...ofLegendPotion($item`Deep Dish of Legend`, "deepDishOfLegendEaten"),
-      ]
-    : [];
+  const deepDish = get("deepDishOfLegendEaten")
+    ? []
+    : !canCookLegendaryPizza($item`Deep Dish of Legend`)
+    ? []
```
horrible-little-slime, line 614: combine these conditions

src/diet.ts
```diff
@@ -328,11 +336,21 @@ function menu(): MenuItem<Note>[] {
...
 
+  const legendaryPizzas = [
+    { pizza: $item`Pizza of Legend`, pref: "pizzaOfLegendEaten" },
+    { pizza: $item`Calzone of Legend`, pref: "calzoneOfLegendEaten" },
+  ]
+    .filter(({ pizza, pref }) => {
+      if (get(pref)) return false;
+      return canCookLegendaryPizza(pizza);
+    })
+    .map(({ pizza }) => new MenuItem(pizza));
```
horrible-little-slime, line 352: we need the price override for these as well

src/diet.ts
```diff
@@ -328,11 +354,19 @@ function menu(): MenuItem<Note>[] {
...
   const limitedItems = [...boxingDayCareItems, ...pilsners].map(
     (item) => new MenuItem<Note>(item, { maximum: availableAmount(item) })
   );
 
+  const legendaryPizzas = legendaryPizzaToMenu(
+    [
+      { item: $item`Calzone of Legend`, pref: "calzoneOfLegendEaten" },
+      { item: $item`Pizza of Legend`, pref: "pizzaOfLegendEaten" },
+    ],
+    (out) => new MenuItem<Note>(out.item, { priceOverride: out.price })
```
horrible-little-slime, line 367: we need a `maximum: 1` in the options here

src/diet.ts
```diff
@@ -589,31 +624,10 @@ export function potionMenu(
...
-  const ofLegendMenuItems = globalOptions.ascend
-    ? [
-        ...ofLegendPotion($item`Calzone of Legend`, "calzoneOfLegendEaten"),
-        ...ofLegendPotion($item`Pizza of Legend`, "pizzaOfLegendEaten"),
-        ...ofLegendPotion($item`Deep Dish of Legend`, "deepDishOfLegendEaten"),
-      ]
-    : [];
+  const deepDish = legendaryPizzaToMenu(
+    [{ item: $item`Deep Dish of Legend`, pref: "deepDishOfLegendEaten" }],
+    (out: { item: Item; price: number }) => limitedPotion(out.item, 1, { price: out.price })[0]
```
horrible-little-slime, line 629: will this ever be nullish? the `[0]` has me 👀ing

s-k-z: No, but it's very annoying since it could fall apart in the future.

horrible-little-slime (approved):
a little surprised you didn't put `flat` on the end of the `map` in `legendaryPizzaToMenu` but this has been cooking long enough, it's time to let it ride

PR #1409, opened 2023-05-23
Burn mp to rest if we need to

Pantocyclus:


src/yachtzee/lib.ts
```diff
@@ -124,6 +138,16 @@ export function useSpikolodonSpikes(): void {
 export function freeRest(): boolean {
   if (get("timesRested") >= totalFreeRests()) return false;
 
+  if (myHp() >= myMaxhp() && myMp() >= myMaxmp()) {
+    // burn some mp so that we can rest
+    const bestSkill = maxBy(
+      $skills``.filter((sk) => have(sk) && mpCost(sk) >= 1),
```
horrible-little-slime, line 144:
```suggestion
      $skills.all().filter((sk) => have(sk) && mpCost(sk) >= 1),
```

horrible-little-slime: This strikes me as potentially dangerous; there are several MP-costing skills that we actively don't want active. Alternately, this could do Weird Thrall stuff.

Would it be easier to do a mix of meatMood and burnLibrams?

horrible-little-slime: We should still prioritize product mp use over unproductive mp use by using meatMood and/or burnLibrams

as Katarn suggested, Awful Poetry Journal is pretty good as an emergency option.

Players may be a WoL or SOL class; do they have 1-mp class skills?

horrible-little-slime (approved):
I think an optimal implementation of this calls a `mood` function and then uses poetry journal as a fallback, but great is the enemy of good, and the enemy of your friend is your enemy.

PR #1413, opened 2023-06-01
Skip sloppy seconds NC when spiking

gausie:


horrible-little-slime (changes requested):
Instead of using a withChoices wrapper, use the propertiesmanager the way unsupportedChoices does in wanderer.ts

Also, go ahead and export unsupportedChoices through the index of wanderer, so we can import it from the module and not from the lib

PR #1414, opened 2023-06-01
value circumstantial bonuses when running barf turns

horrible-little-slime:


ajcoppa (approved):
Not a trained garbologist, but this makes sense

PR #1416, opened 2023-06-05
Initial support for Comma Chameleon

Pantocyclus:


src/fights.ts
```diff
@@ -415,18 +416,42 @@ export function dailyFights(): void {
...
           }
 
-          const profSpec: OutfitSpec = { familiar: $familiar`Pocket Professor` };
-          if (have(chip)) {
+          const profSpec: OutfitSpec = {
+            familiar: have($familiar`Pocket Professor`)
+              ? $familiar`Pocket Professor`
+              : $familiar`Comma Chameleon`,
+          };
+          if (have(chip) && myFamiliar() === $familiar`Pocket Professor`) {
```
horrible-little-slime, line 454: chameleon can't equip chip, so that's not going to work

src/fights.ts
```diff
@@ -415,18 +416,42 @@ export function dailyFights(): void {
...
+              if (!have(chip)) {
+                acquire(
+                  1,
+                  chip,
+                  EMBEZZLER_MULTIPLIER() *
+                    get("valueOfAdventure") *
+                    Math.min(40, pocketProfessorLectures() - get("_pocketProfessorLectures"))
+                );
+              }
+              visitUrl(`inv_equip.php?which=2&action=equip&whichitem=${toInt(chip)}&pwd`);
```
horrible-little-slime, line 444: need to make the chameleon your familiar before calling this, or equip straight to the terrarium

horrible-little-slime (commented):
I worry that this will make affected codeblocks harder to maintain. At the very least, we should create libram support for comma chameleon before starting on this.

s-k-z:
Amulet coin isn't valued currently, if this pushes up the market price on those, it should be important to check if it's worth using it on them.

docrostov:
i am not going to close this issue but i am going to raise a point here -- the economic ramifications of this PR are... not great. in a 600 turn day, famjacks will be used 13-14 times by a comma owner trying to copy a robort. that's a hell of a lot of jacks per user. we already saw jacks prices go completely wild during cognac-mania; if garbo does this, we wouldn't -just- need an amulet coin check, we'd also need to add ways for garbo to only convert comma-to-robort for embezzler fights and/or additional major jacks handling. this also relies on users to be using keeping-tabs or philter to convert their jacks to robort equipment, AND relies on people not trying to capture that market. overall, my lean is this would cease to be profitable a week or two after implementation and leave the market for famjacks in flames. is this really worth it?

Pantocyclus:
The idea here is that it should equilibriate to a point where jacks are too expensive to be used every turn, but one can still eke value out of the first 5 jelly extracts and whatever other limited daily fam drops.

E.g. with a single jack one should expect to be able to grab the first 5 stench jelly drops, putting the value of that jack at >=150k (far above the value of a robort jack)

PR #1417, opened 2023-06-06
Pickpocket GOTOs while using mayflies

Shiverwarp:


horrible-little-slime (approved):

Rinn (commented):
My only thought would be to check maxPassiveDamage() to make sure we won't accidently kill the monster. I don't remember if we bjorn anything or what we do with pasta thralls, those would be the most likely sources

Shiverwarp:
> My only thought would be to check maxPassiveDamage() to make sure we won't accidently kill the monster. I don't remember if we bjorn anything or what we do with pasta thralls for free run/fights, those would be the most likely sources

Good shout, I made some changes

Shiverwarp:
Alright I think this needs some changes.

1. Should be moved before Hipster fights. This only needs two banishes to guarantee the Basic elemental every time. Hipster fights eat up all remaining non-familiar fights.
2. Maybe shouldn't specify a familiar? I did so to avoid attacking familiars, but this means Bander won't be used here
3. Olfact/Turtlesex the basic elemental (only olfact if we haven't used any olfacts at all today, and maybe not even that? Would need input whether to specify that)
4. Try to favour doing different banishes on each bad monster. Probably I'll just specify hatred/snoke for now but any suggestions how to handle it better very welcome

src/fights.ts
```diff
@@ -1737,6 +1737,59 @@ const freeRunFightSources = [
       },
     }
   ),
+  // Try for an ultra-rare with mayfly runs ;)
+  new FreeRunFight(
+    () =>
+      have($item`mayfly bait necklace`) &&
+      canAdventure($location`Cobb's Knob Menagerie, Level 1`) &&
+      // Max combat length is 4 rounds, Basic elementals have 40 HP base
+      4 * maxPassiveDamage() < 39 + numericModifier("Monster Level") &&
```
Rinn, line 1746: down side here is if you do have some passive damage and could still free run but not do all the rounds this will stop it from free running with mayfly bait

Shiverwarp: Is the best way to do that just to make another copy of this FreeRunFight where if we do have passive damage, we only do the free runs and no pickpocketing?

The other option is I make extra checks for each step whether the monsters HP is high enough, but that would require monster manuel

Rinn (changes requested):
one small thing and this should be good

src/fights.ts
```diff
@@ -1737,6 +1744,59 @@ const freeRunFightSources = [
...
+              .trySkill($skill`Summon Mayfly Swarm`)
+          )
+          .if_($monster`Fruit Golem`, Macro.trySkill($skill`Feel Hatred`))
+          .if_($monster`Knob Goblin Mutant`, Macro.trySkill($skill`Snokebomb`))
+          .step(runSource.macro)
+      );
+    },
+    {
+      spec: () => {
+        const canPickPocket = myPrimestat() === $stat`Moxie`;
```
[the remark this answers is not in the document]
Shiverwarp: Good call

```
> js Class.all().filter((i) => i.primestat == "Moxie")

Returned: aggregate class [6]
0 => Disco Bandit
1 => Snake Oiler
2 => Accordion Thief
3 => Avatar of Sneaky Pete
4 => Jazz Agent
5 => Gelatinous Noob
```

Also fixed the other place we were using this to determine our pickpocket skill

Shiverwarp:
Oops I accidentally clicked the re-request review button, hope that isn't too annoying for whoever received it

Rinn (approved):

PR #1420, opened 2023-06-15
Cincho Projectile Pinata

Shiverwarp:
Took a crack at this.

I've swapped around how the Feliz Value is handled so that I can export and use it

This will require the user to have manuel so that we can safely use pinata while doing stasis before critting.

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -387,9 +387,8 @@ export class Macro extends StrictMacro {
 
     const checkGet = (i: Item) => have(i) && (itemAmount(i) > 0 || retrieveItem(i));
     const stasisItem = $items`facsimile dictionary, dictionary, seal tooth`.find(checkGet);
-    const canPinata =
-      CinchoDeMayo.have() && CinchoDeMayo.currentCinch() >= 5 && monsterManuelAvailable();
     const pinataCastsAvailable = Math.floor(CinchoDeMayo.currentCinch() / 5);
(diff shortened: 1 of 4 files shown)


horrible-little-slime (commented):
My only real concern here is I am not convinced that the average user who owns cincho is successfully getting out a projectile without sacrificing crits against tourists, and so I'm a little skeptical about that value estimate.

I can be convinced otherwise! but I will need convincing

Shiverwarp:
> My only real concern here is I am not convinced that the average user who owns cincho is successfully getting out a projectile without sacrificing crits against tourists, and so I'm a little skeptical about that value estimate.
> 
> I can be convinced otherwise! but I will need convincing

The `maxPassiveDamage() >= 50` check for the bonusEquip should cover that, as well as requiring manuel. Perhaps I'm misunderstanding something there though

horrible-little-slime:
> > My only real concern here is I am not convinced that the average user who owns cincho is successfully getting out a projectile without sacrificing crits against tourists, and so I'm a little skeptical about that value estimate.
> > I can be convinced otherwise! but I will need convincing
> 
> The `maxPassiveDamage() >= 50` check for the bonusEquip should cover that, as well as requiring manuel. Perhaps I'm misunderstanding something there though

This seems fine?
My biggest core concern is that I think people really like seeing numbers hit 100 or 0 or whatever, but doing so isn't always actually optimal; if we end up failing to crit-force and sing along and extract jelly because we're cinchoing, that can easily be a net loss of meat by using this resource.

I don't think we're actually in that situation, though. I'll just merge this.

PR #1422, opened 2023-06-22
Use retrocape robo-kiss on the %monster% if available

s-k-z:
The macro will do a minimum 9 round stun assuming you own all skills/items as an arbitrary class.
Extra leeway is available for Pastamancers and Saucerors with entangling noodles or soul bubble for more rounds of stuns.
You need at least 5 rounds of stuns to ensure victory.

### DIFF
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -818,6 +818,40 @@ const freeFightSources = [
   new FreeFight(
     () =>
       have($item`[glitch season reward name]`) &&
+      have($item`unwrapped knock-off retro superhero cape`) &&
+      !get("_glitchMonsterFights") &&
+      get("garbo_fightGlitch", false),
+    () =>
+      withMacro(
+        Macro.trySkill($skill`Shadow Noodles`)
+          .trySkill($skill`Entangling Noodles`)
+          .trySkill($skill`Summon Love Gnats`)
+          .trySkill($skill`Frost Bite`)
+          .trySkill($skill`Soul Bubble`)
+          .tryItem($item`Rain-Doh blue balls`)
+          .skill($skill`Blow a Robo-Kiss`)
+          .repeat(),
+        () => {
+          restoreHp(myMaxhp());
+          if (have($skill`Blood Bubble`)) ensureEffect($effect`Blood Bubble`);
+          retrieveItem($item`[glitch season reward name]`);
+          visitUrl("inv_eat.php?pwd&whichitem=10207");
+          runCombat();
+        }
+      ),
+    true,
+    {
+      spec: {
+        back: $items`unwrapped knock-off retro superhero cape`,
+        modes: { retrocape: ["robot", "kiss"] },
+      },
+      macroAllowsFamiliarActions: false,
+    }
+  ),
+
+  new FreeFight(
+    () =>
+      have($item`[glitch season reward name]`) &&
       !get("_glitchMonsterFights") &&
       get("garbo_fightGlitch", false),
     () =>
@@ -1652,11 +1686,12 @@ const freeRunFightSources = [
       get("_gingerbreadCityTurns") + (get("_gingerbreadClockAdvanced") ? 5 : 0) === 19 &&
       (availableAmount($item`sprinkles`) > 5 || haveOutfit("gingerbread best")),
     () => {
+      const choice = itemAmount($item`sprinkles`) > 300 ? 2 : 1;
       propertyManager.setChoices({
         1203: 4, // Gingerbread Civic Center 5 gingerbread cigarettes
         1215: 1, // Gingerbread Civic Center advance clock
         1209: 2, // enter the gallery at Upscale Midnight
-        1214: 1, // get High-End ginger wine
+        1214: choice, // get fancy chocolate sculpture OR High-End ginger wine
       });
       const best = bestConsumable("booze", true, $items`high-end ginger wine, astral pilsner`);
       const gingerWineValue =
@@ -1664,6 +1699,7 @@ const freeRunFightSources = [
           getAverageAdventures($item`high-end ginger wine`) * get("valueOfAdventure")) /
         2;
       const valueDif = gingerWineValue - best.value;
+      print(`Gingerbread wine value = ${valueDif}`);
       if (
         haveOutfit("gingerbread best") &&
         (availableAmount($item`sprinkles`) < 5 ||
--- a/src/fights.ts
+++ b/src/fights.ts
@@ -823,7 +823,8 @@ const freeFightSources = [
       get("garbo_fightGlitch", false),
     () =>
       withMacro(
-        Macro.trySkill($skill`Shadow Noodles`)
+        Macro.trySkill($skill`Curse of Marinara`)
+          .trySkill($skill`Shadow Noodles`)
           .trySkill($skill`Entangling Noodles`)
           .trySkill($skill`Summon Love Gnats`)
           .trySkill($skill`Frost Bite`)


Shiverwarp (dismissed):
Today I learned stun duration stacks.

Because this is already behind a pref I think it's fine without ensuring the user has the minimum amount of stuns

s-k-z:
Added a minimum stun durations, so they can be summed up to ensure there's enough to do this.

src/fights.ts
```diff
@@ -701,6 +702,16 @@ function molemanReady() {
...
 }
 
+const stunDurations = new Map<Skill | Item, Delayed<number>>([
+  [$skill`Blood Bubble`, 1],
+  [$skill`Entangling Noodles`, () => (myClass() === $class`Pastamancer` ? 2 : 0)],
+  [$skill`Frost Bite`, 2],
+  [$skill`Shadow Noodles`, 3],
+  [$skill`Soul Bubble`, () => (mySoulsauce() >= 5 ? 3 : 0)],
+  [$skill`Summon Love Gnats`, 2],
+  [$item`Rain-Doh blue balls`, 2],
```
Shiverwarp, line 712:
```suggestion
  [$skill`Blood Bubble`, 1],
  [$skill`Entangling Noodles`, () => (myClass() === $class`Pastamancer` ? 1 : 0)],
  [$skill`Frost Bite`, 1],
  [$skill`Shadow Noodles`, 2],
  [$skill`Soul Bubble`, () => (mySoulsauce() >= 5 ? 2 : 0)],
  [$skill`Summon Love Gnats`, 1],
  [$item`Rain-Doh blue balls`, 1],
```

As per discussion and spading on discord I think these should be the actual durations because it takes a combat round to cast these?

src/fights.ts
```diff
@@ -701,6 +702,16 @@ function molemanReady() {
   return have($item`molehill mountain`) && !get("_molehillMountainUsed");
 }
 
+const stunDurations = new Map<Skill | Item, Delayed<number>>([
+  [$skill`Blood Bubble`, 1],
+  [$skill`Entangling Noodles`, () => (myClass() === $class`Pastamancer` ? 1 : 0)],
+  [$skill`Frost Bite`, 1],
+  [$skill`Shadow Noodles`, 2],
```
horrible-little-slime, line 709: Shadow Noodles cannot be used in the same fight as Entangling Noodles; I don't think(?) this handles that(?)

src/fights.ts
```diff
@@ -815,6 +826,45 @@ const freeFightSources = [
...
 
+  new FreeFight(
+    () =>
+      have($item`[glitch season reward name]`) &&
+      have($item`unwrapped knock-off retro superhero cape`) &&
+      !get("_glitchMonsterFights") &&
+      get("garbo_fightGlitch", false) &&
+      Array.from(stunDurations).reduce(
+        (sum, [thing, duration]) => sum + (have(thing) ? undelay(duration) : 0),
+        0
```
horrible-little-slime, line 837:
```suggestion
     sum([...stunDurations], ([thing, duration]) => have(thing) ? undelay(duration) : 0) >= 5
```

horrible-little-slime (commented):
I think this looks good?

Is blowing a kiss always more desirable than conventional methods?

s-k-z:
The advantage is that it works at any level for any quantity of implementations and doesn't use any resources aside from MP. Disadvantage is you lose the back slot, but I think that's pretty minor.

src/fights.ts
s-k-z: I think this suggestion doesn't cover enough lines.

src/fights.ts
```diff
@@ -821,6 +835,45 @@ const freeFightSources = [
...
+      ) >= 5,
```
horrible-little-slime, line 847:
```suggestion
     sum([...stunDurations], ([thing, duration]) => have(thing) ? undelay(duration) : 0) >= 5
```
Does this cover enough lines? more news as this develops

src/fights.ts
```diff
@@ -821,6 +835,42 @@ const freeFightSources = [
...
+    () =>
+      withMacro(
+        Macro.trySkill($skill`Curse of Marinara`)
+          .trySkill($skill`Shadow Noodles`)
+          .trySkill($skill`Entangling Noodles`)
+          .trySkill($skill`Summon Love Gnats`)
+          .trySkill($skill`Frost Bite`)
+          .trySkill($skill`Soul Bubble`)
+          .tryItem($item`Rain-Doh blue balls`)
+          .skill($skill`Blow a Robo-Kiss`)
```
Malibu-Stacey, line 854: Uh this is a Yellow Ray and a not free one at that.
I think you meant to use Heck General Tendrils which just requires configuring and wearing the retrocape.
https://kol.coldfront.net/thekolwiki/index.php/Unwrapped_knock-off_retro_superhero_cape

src/fights.ts
```diff
@@ -707,6 +708,19 @@ function molemanReady() {
...
+  [$skill`Blood Bubble`, 1],
+  [
+    $skill`Entangling Noodles`,
+    () => (myClass() === $class`Pastamancer` && !have($skill`Shadow Noodles`) ? 1 : 0),
+  ],
+  [$skill`Frost Bite`, 1],
+  [$skill`Shadow Noodles`, 2],
+  [$skill`Soul Bubble`, () => (mySoulsauce() >= 5 ? 2 : 0)],
+  [$skill`Summon Love Gnats`, 1],
+  [$item`Rain-Doh blue balls`, 1],
```
Malibu-Stacey, line 721: I would also conditionally add

- Club Foot: requires SC, 2+ Fury & wielding a club to Stun. Otherwise just Staggers
- Shell Up: requires TT and Blessing of the Storm Tortoise to Stun. Otherwise just Staggers
- Accordion Bash: requires AT & wielding an accordion to Stun. Otherwise just Staggers

Lost Comb is also a re-usable 3 round Stun combat item. Trivial to acquire for anyone with the VIP room key.

Malibu-Stacey (changes requested):

src/fights.ts
horrible-little-slime: Sauceshell too, which has no reqs and does some chip damage

Rinn: The Lost Comb is used up after 3 uses just btw.

src/fights.ts
Shiverwarp: Robo kiss is the correct one, no?

[](https://kol.coldfront.net/thekolwiki/index.php/Blow_a_Robo-Kiss)

src/fights.ts
Shiverwarp: > Sauceshell too, which has no reqs and does some chip damage

I think the intention of this is to kill specifically with Robo Kiss, therefore any extra chip damage is superfluous and doesn't actually help us.

If the user doesn't have enough stuns, we should use the normal glitch kill anyway.

src/fights.ts
s-k-z: Unleash the Devil's Kiss is the yellow ray.

src/fights.ts
s-k-z: I'll add shell up, the rest are too conditional.

src/fights.ts
Malibu-Stacey: Robo-kiss does 20% HP as damage but doesn't stun. The multi-round stun is Heck General/Hold Me.

Shiverwarp: The plan is to stack up 5 rounds of consecutive stuns, and then use robo-kiss while the monster is stunned to kill it

Malibu-Stacey (approved):

Rinn (approved):

PR #1423, opened 2023-06-26
Greedy dog for diet

Shiverwarp:
This is simple to add for general garbo, but yachtzee is a bit more arcane for me. Hopefully I did it correctly, I just left the max price of them at 60k, same as boris bread and radishes

(Max price for embezzlers would be something under 40k, and something under 80k for yachtzee depending on user VoA)

I don't know if the supply of these can meet demand currently, but people might consider zapping race car reds and lambadas. So essentially people with the beer garden are making one per week if I understand that correctly.

Perhaps that was a reason these weren't implemented before though. to not mess with the supply of those.

### DIFF
--- a/src/yachtzee/diet.ts
+++ b/src/yachtzee/diet.ts
@@ -142,6 +142,10 @@ class YachtzeeDietUtils {
         ensureConsumable("Boris's bread", n, 1, 0, 0);
         eat(n, $item`Boris's bread`);
       }),
+      new YachtzeeDietEntry("bottle of Greedy Dog", 0, 0, 3, 0, (n: number) => {
+        ensureConsumable("bottle of Greedy Dog", n, 0, 3, 0);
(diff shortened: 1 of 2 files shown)


horrible-little-slime (approved):

src/yachtzee/diet.ts
```diff
@@ -142,6 +142,10 @@ class YachtzeeDietUtils {
         ensureConsumable("Boris's bread", n, 1, 0, 0);
         eat(n, $item`Boris's bread`);
       }),
+      new YachtzeeDietEntry("bottle of Greedy Dog", 0, 0, 3, 0, (n: number) => {
+        ensureConsumable("bottle of Greedy Dog", n, 0, 3, 0);
+        eat(n, $item`bottle of Greedy Dog`);
```
gausie, line 147: Should this not be drink?

PR #1438, opened 2023-07-06
reduce the glitch implentation count to start with gas cans.

Zasqdddas:


Zasqdddas:
Created a pull request to lower the start at which gas cans are used to kill the glitch.
I was dying to glitch fights with a full base skill and Dread skill perm set at ~500 implementations.  
(Hope I didn't mess this up too badly - GARFJALEN was a lot of help if I didn't.)

s-k-z (changes requested):
Line 860 also needs to be changed, otherwise gas cans won't be retrieved.

s-k-z (approved):

PR #1442, opened 2023-07-09
fix: use my forks of Ezandora's scripts to fix modifier issues

midgleyc:
Recent modifier changes in https://github.com/kolmafia/kolmafia/pull/1815 broke Bastille and Gain; point to my forks instead until the PRs are merged so that users don't have to pull scripts manually.


docrostov (approved):
given that these changes are semi-breaking, approved/merged

PR #1454, opened 2023-07-16
Cincho projectile pinata (should be ready?)

Shiverwarp:


Rinn:
might be a few days before I can look into this

Rinn:
I've almost got all the existing circular imports resolved, this is on my todo list after I finish those.

Rinn:
Ah sorry forgot about this and I'm about to go on holiday. What I would do if you want some idea on what was happening is update your branch then run `yarn madge` on main, then run it on your branch and see what additional circular dependencies are exposed and that should help you track down what caused the issue last time.

Rinn:
@Shiverwarp Okay I did a quick rebase of this locally to figure out the circular imports, here's what you'll need to do I think:

The merge issue is you'll need to move cinchoDeMayo() to dropgearAccessories.ts

The circular dependencies are the following:
- felizValue() from tasks/dailyFamiliars.ts
- maxPassiveDamage() & monsterManuelAvailable() from combat.ts

If those were resolved it should fix the build issues.

I'd recommend for felizValue() just using `get("garbo_felizValue", 0)` and call felizValue() during the daily familiar tasks to init that property if you have Cincho de Mayo but not robortender.

For the other two functions it looks like they and all their dependent functions could be move to a separate file no problem as they don't appear to rely on other garbo ts files.

ajcoppa:
I've run off this branch for 2 days and it works great. Uses all cinch on projectiles, including free resting between combats to refill and use it again.

src/combat.ts
```diff
@@ -466,6 +473,13 @@ export class Macro extends StrictMacro {
           haveEquipped($item`vampyric cloake`) && get("_vampyreCloakeFormUses") < 10,
           Macro.if_(`${hpCheck}`, Macro.tryHaveSkill($skill`Become a Wolf`)),
         )
+        .externalIf(
+          haveEquipped($item`Cincho de Mayo`) && canPinata,
+          Macro.while_(
+            `${hpCheckCincho} && hasskill 7442`,
```
Rinn, line 479: please get the id from the skill object rather than hard coding it

src/combat.ts
```diff
@@ -528,11 +542,20 @@ export class Macro extends StrictMacro {
...
+    const canPinata = haveEquipped($item`Cincho de Mayo`) && CinchoDeMayo.currentCinch() >= 5;
     return (
       this.externalIf(
         myClass() === $class`Sauceror` && have($skill`Curse of Weaksauce`),
         Macro.trySkill($skill`Curse of Weaksauce`),
       )
+        .externalIf(
+          !doingYachtzee && canPinata,
+          Macro.while_(
+            `hasskill 7442 && !pastround 24 && !hppercentbelow 25`,
```
Rinn, line 555: please get the id from the skill object rather than hard coding it

Rinn (changes requested):

src/combat.ts
```diff
@@ -668,41 +693,41 @@ export class Macro extends StrictMacro {
...
-          Macro.tryCopier($skill`Digitize`),
+          Macro.tryCopier($skill`Digitize`)
         )
         .tryCopier($item`Spooky Putty sheet`)
         .tryCopier($item`Rain-Doh black box`)
         .tryCopier($item`4-d camera`)
         .tryCopier($item`unfinished ice sculpture`)
         .externalIf(get("_enamorangs") === 0, Macro.tryCopier($item`LOV Enamorang`))
-        .meatKill(),
+        .meatKill()
```
horrible-little-slime, line 730: It seems like we may have a linting rule that is being inconsistently applied across different installs

Rinn (approved):

src/combat.ts
s-k-z: This is due to [prettier 3.0](https://prettier.io/blog/2023/07/05/3.0.0.html)

> Change the default value for `trailingComma` to `all`

PR #1463, opened 2023-07-18
Don't try to equip a moveable feast between each adventure when...

tyrion-the-imp:
there is no access to one.


Rinn (changes requested):
this will prevent a user that owns a moveable feast that has a stashClan set to none from using it, you would also need to check if one is available

tyrion-the-imp:
OK...think it's right now?

horrible-little-slime:
Feels like a more elegant fix would just be to track whether we've tried to moveable feast with a pref or a variable

tyrion-the-imp:
> Feels like a more elegant fix would just be to track whether we've tried to moveable feast with a pref or a variable

I wasn't even sure how the feast worked.  I just wanted to stop my character pane from flashing red between each yachtzee turn. (This does fix that.)

PR #1475, opened 2023-07-25
create WandererManager for grimoirization

horrible-little-slime:
the wanderer manager uses a caching mechanism that'll make it easier to expose wanderer information to different parts of a Grimoire Task; it does, however, separate the NC-setting code from the wanderer-choosing code, which will be awkward until then.

Once rinn's mafia PR is live we can maybe get rid of the `clear` method, or at least not rely on it as much


horrible-little-slime:
Re-drafted because we might as well wait for https://github.com/kolmafia/kolmafia/pull/1854 for better cache invalidation

Rinn (approved):

PR #1497, opened 2023-08-02
Grimoire free fight work in progress

Rinn:
prob not going to have time to continue this any time soon if anyone wants to run with it

todo:
- actually hook up the tasks to an engine
- check costs (do we really even need the cost property i added? just make a condition of available instead?)
- move the rest of the free fight tasks (maybe split these up logically to separate quests to make it cleaner)
- wanderer choices aren't returned in a structure that can be passed right to grimoire task choices
- free runs
- free kills (sandworms)


horrible-little-slime:
We have some really messy code right now for getting robortender ingredients from free kills. We don't need to tackle that here, but it does make me think that maybe locket shouldn't always force robort, given that current code doesn't value robort drops

PR #1499, opened 2023-08-06
Scepter

Tokoeka:
Add in August Scepter. Currently only calculates the best 5 skills (ignoring the skill of the day) and doesnt actually do anything with them.

### DIFF
--- a/src/tasks/daily.ts
+++ b/src/tasks/daily.ts
@@ -60,7 +60,7 @@ import { globalOptions } from "../config";
 import { embezzlerCount } from "../embezzler";
 import { meatFamiliar } from "../familiar";
 import { estimatedTentacles } from "../fights";
-import { averageEmbezzlerNet, baseMeat, HIGHLIGHT } from "../lib";
+import { baseMeat, HIGHLIGHT } from "../lib";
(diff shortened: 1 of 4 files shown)


src/tasks/daily.ts
```diff
@@ -350,6 +356,46 @@ export function configureSnojo(): void {
   }
 }
 
+const MPA = get("valueOfAdventure");
+
+const oysters = $items`brilliant oyster egg, gleaming oyster egg, glistening oyster egg, lustrous oyster egg, magnificent oyster egg, pearlescent oyster egg, scintillating oyster egg`;
+
+export function bestScepterSkills(): Skill[] {
+  const date = toInt(todayToString().slice(-2)); // TODO: this may be based on local time despite what mafiawiki says
```
horrible-little-slime, line 364: use `Number(todayToString()) % 100`

horrible-little-slime:
In the immortal words of A Tribe Called Quest, We got it from Here... Thank You 4 Your service

(#1532)

I did pilfer a lot of this code and it was extremely helpful but I think 1532 better matches the shape of what this'll need to be.

Issue #1490, opened 2023-07-30
Use volcano disco for sobriety instead of free volcoino, when MPA is very high


sundayLOOKplants:
when cognac-ing, etc, the turns from a splendid martini (etc) are worth a fair bit more than 1 volcoino.

docrostov:
garbo already shuts off at VOA > 10k; if you want this functionality, it's relatively easy to do a quick ash wrapper that `drink 1 splendid martini; [insert whatever command is needed to then clear that with volcano]` before you run garbo. closing this.

PR #1500, opened 2023-08-06
use floor 4 if VOA warrants it

Tokoeka:
Considers whether to reduce drunkeness at Disco Inferno rather than getting the volcoino, if your VOA warrants it: #1490

### DIFF
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -28,6 +28,7 @@ import {
   npcPrice,
   print,
   retrievePrice,
+  runChoice,
   sellsItem,
   setProperty,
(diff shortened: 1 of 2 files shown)


Tokoeka:
It has been pointed out this is a cleaner, not an expander, further implementation required

docrostov:
garbo already shuts off at VOA > 10k; if we want this functionality, it's relatively easy to do a quick ash wrapper that `drink 1 splendid martini; [insert whatever command is needed to then clear that with volcano]` before one runs garbo. closing this.

PR #1501, opened 2023-08-06
Perpetrate mild evil

CheeseyPickle:
Works the same as extinguisher, so we can just stick it in there, I think.


burningbman:
Doesn't package.json need to be updated too? That's what dependabot is doing in the opened PRs.

burningbman (approved):

PR #1509, opened 2023-08-09
August Scepter support for yachtzee diet and fishy source

Pantocyclus:


src/yachtzee/diet.ts
```diff
@@ -267,6 +267,13 @@ export function executeNextDietStep(stopBeforeJellies?: boolean): void {
         const entry = dietUtil.dietArray.find((entry) => entry.name === name);
         if (entry) {
           if (entry.fullness > 0) {
+            if (
+              have($skill`Aug. 16th: Roller Coaster Day!`) &&
+              !get("_aug16Cast") &&
+              myFullness() > 0
+            ) {
+              useSkill($skill`Aug. 16th: Roller Coaster Day!`);
+            }
```
horrible-little-slime, line 276: I don't know if we want to blindly cast this one

horrible-little-slime: Same with the other one. Unless it happens to be that day, we don't know that we definitely want to cast it.

Pantocyclus: I'm working off the assumption that this is probably the last thing we're going to be doing - we would have grabbed items in dailies and buffs prior to yachtzeeing, so if we deemed those less profitable than a stomach space/Lucky! charge, then we shouldn't have any issues using those here

Pantocyclus: I suppose we might want to save 1 charge of the scepter skills for offhand remarkable if we aren't ascending

horrible-little-slime: If you add that I'll merge this in right away. I have an image for what I want support to look like now, so we're all good

Pantocyclus:
There's a fun little issue where if the scepter is equipped on the lefty, you can't actually cast any of the August skills.

Not sure if this is a KoL bug, or whether mafia or garbo should be the one responsible for unequipping the lefty whenever we want to cast an August skill

docrostov:
> There's a fun little issue where if the scepter is equipped on the lefty, you can't actually cast any of the August skills.
> 
> Not sure if this is a KoL bug, or whether mafia or garbo should be the one responsible for unequipping the lefty whenever we want to cast an August skill

this is a persistent kol bug. it’s rare but i believe this also happens with lefty kramco. effectively, for things that require ownership to do a special action, the game doesn’t think you own most things equipped on a familiar in your terrarium (with key exceptions being crystal ball ponder and stillsuit drinking). it’s probably a good idea for garbo to unequip whatever lefty is holding by default

horrible-little-slime:
Will delete or heavily modify this as I roll out full support after grimoirization of free fights

Issue #673, opened 2021-12-31
[resource] Equip guzzlr pants on last turn to get more bux


Rinn:

horrible-little-slime:
One thing that makes this more challenging is that our outfit and wanderer code don't currently talk to eachother. Adding it as a bonusEquip would be trivial were it not for that.

pstalcup:
Do we know when we build our outfit if our next turn will be a wanderer? 
This will be complicated to solve and require restructuring of lots of code

docrostov:
wanderer manager code refactor makes this easier for outfits to access; phred thinks this is a medium difficulty add (which probably means it's med-to-high) where you'd override the embezzler outfit in the event it's the last guzzlr turn. it is worth roughly 8-9k per quest if we do it. good luck!

PR #1521, opened 2023-08-23
Add bonus value for guzzlr pants

horrible-little-slime:
#673 ya burnt


docrostov (approved):
looks fine to me but i did not test it yolo

PR #1525, opened 2023-08-25
Support for HR path; clean up support for future free kill wanderers

pstalcup:


src/barfTurn.ts
```diff
@@ -286,6 +288,19 @@ const turns: AdventureAction[] = [
     spendsTurn: false,
     sobriety: Sobriety.EITHER,
   },
+  {
+    name: "Heavy Rains Rain Main",
+    available: () => have($skill`Rain Man`) && myRain() >= 50,
```
horrible-little-slime, line 293: This doesn't check whether you have embezzler as rainmannable--what are the conditions for that? I know that having it in your manuel is helpful, but I assume that rain man isn't wholly gated behind manuel?

src/barfTurn.ts
```diff
@@ -380,6 +395,26 @@ const turns: AdventureAction[] = [
...
+      const location = wanderer.getTarget("backup");
+      freeFightOutfit({}, { location }).dress();
+      const macro = Macro.if_(embezzler, Macro.meatKill())
+        .familiarActions()
+        .skill($skill`Lightning Strike`);
+      garboAdventureAuto(location, macro);
+      if (SourceTerminal.have()) {
+        SourceTerminal.educate([$skill`Extract`, $skill`Digitize`]);
+      }
+      return have($effect`Everything Looks Red`);
```
horrible-little-slime, line 413: this seems wrong

src/barfTurn.ts
Tokoeka: have it in your manuel, or have fought it in the current HR ascension as I understand it

Rinn: yes that is correct, if you don't have manuel you can hit the choice adventure once to get the full list for free if you want to know if you can summon something. Or just only support manuel havers.

Rinn: this is the check from autoscend if you wanted to verify first
```
boolean canRainManSummon(monster target)
{
	if(!have_skill($skill[Rain Man]) || my_rain() < 50)
	{
		return false;
	}

	// Can only rain man summon copyable monsters
	if(!target.copyable || target.id < 0)
	{
		return false;
	}

	// Can summon any monster with available factoids
	if (target.monster_factoids_available(false) > 0)
	{
		return true;
	}

	// Check the page text
	auto_log_info(target + " factoids unavailable, checking Rain Man if summon is possible", "blue");
	buffer page = visit_url("runskillz.php?pwd&action=Skillz&whichskill=16011&quantity=1");
	// Escape
	run_choice(2);

	return page.contains_text("<option value=" + target.id + ">");
}
```

horrible-little-slime:
Update this to use the new libram functions and I'll merge 'er in

src/barfTurn.ts
horrible-little-slime: Also fix this

horrible-little-slime:
Needs to be updated for the new shit, but I'm into it

horrible-little-slime:
@pstalcup this is good to close, right?

PR #1534, opened 2023-08-29
Another Scepter PR

xn:
I've got circular deps here that I need help to sort out. I've got it running locally, just by editing the built code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,8 +18,8 @@
   "dependencies": {
     "core-js": "^3.32.1",
     "grimoire-kolmafia": "^0.3.22",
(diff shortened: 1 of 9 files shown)


xn:
I can pull everything from value.ts into lib.ts, but I will either need to move garboValue out of value.ts or find a suitable replacement. Thoughts?

The circular dependency is barfing on EMBEZZLER_MULTIPLIER() anonymous function.

xn:
@horrible-little-slime feel free to scavenge anything you like in here.

xn:
I've also got this PR hanging out on libram: https://github.com/loathers/libram/pull/513

PR #1535, opened 2023-08-30
Break circular dependencies to help with August Scepter

xn:
Extracted embezzlerCount() to its own file and destroyed a loop there, but I didn't remove the refactor  `new EmbezzlerFight ` calls to depend on the the new count function, I want a go ahead on all of this so I don't waste time doing something that is unwanted.

Was able to extract calls to estimatedGarboAdventures() in Macro.meatkill() and Macro.embezzler() by turning it into a parameter. updated calls throughout the app.

Also extracted Potion classes to allow for valuation without cycles in the August Scepter PR.


src/potions/variable_meat_potion.ts
```diff
@@ -0,0 +1,147 @@
...
+  retrievePrice,
+  use,
+} from "kolmafia";
+import { maxBy } from "libram";
+import { acquire } from "../acquire";
+import { baseMeat, retrieveUntradeablePrice, turnsToNC } from "../lib";
+import { estimatedGarboTurns } from "../turns";
+import { globalOptions } from "../config";
+
+function triangleNumber(b: number, a = 0) {
```
horrible-little-slime, line 22: @Pantocyclus I think we should have more informative variable names here; what is the `a` parameter used for? why is `b` the lead variable?

src/potions/variable_meat_potion.ts
```diff
@@ -0,0 +1,147 @@
...
+    this.softcap = softcap;
+    this.meatBonusPerTurn = meatBonusPerTurn;
+    this.cappedMeatBonus = softcap * meatBonusPerTurn;
+  }
+
+  use(quantity: number): boolean {
+    acquire(
+      quantity,
+      this.potion,
+      (1.2 * retrievePrice(this.potion, quantity)) / quantity,
```
horrible-little-slime, line 53: @Pantocyclus where does the 1.2 magic number come from? can we make this a named screaming snake case const somewhere?

src/combat.ts
```diff
@@ -378,8 +378,8 @@ export class Macro extends StrictMacro {
       .kill();
   }
 
-  static meatKill(): Macro {
-    return new Macro().meatKill();
+  static meatKill(turnCount: number): Macro {
```
horrible-little-slime, line 381: This doesn't have to be in scope but it is probably worth separating out our embezzler macro from our barf macro (romance wanderers mean we do still need embezzler code in the meatkill code, but still), because it seems tedious to be passing a turncount to `meatKill` every dingdang time when it almost never matters.

src/counts/embezzler.ts
```diff
@@ -0,0 +1,216 @@
...
+  return sum(embezzlerCounts, (countEntry: NameCountPair) => countEntry[1]());
+}
+
+interface CountFunction {
+  (): number;
+}
+
+type NameCountPair = [string, CountFunction];
+
+const embezzlerCounts = new Array<NameCountPair>(
```
horrible-little-slime, line 33: Why are you using the array constructor instead of an array literal here?

src/counts/embezzler.ts
```diff
@@ -0,0 +1,216 @@
...
+
+export function embezzlerCount(): number {
+  return sum(embezzlerCounts, (countEntry: NameCountPair) => countEntry[1]());
+}
+
+interface CountFunction {
+  (): number;
+}
+
+type NameCountPair = [string, CountFunction];
```
horrible-little-slime, line 31:
```suggestion
type NameCountPair = [string, () => number];
```

that being said, I think this would be easier and more legible as an object type instead of as a tuple, so
```
type EmbezzlerCount = { name: string, quantity: () => number };
```

src/counts/embezzler.ts
```diff
@@ -0,0 +1,216 @@
...
+  get,
+  have,
+  property,
+  SourceTerminal,
+  sum,
+} from "libram";
+
+const embezzler = $monster`Knob Goblin Embezzler`;
+
+export function embezzlerCount(): number {
```
horrible-little-slime, line 23: I am hesitant to separate the count from the EmbezzlerFights themselves; I'll accept it as a necessary evil, but it would be nice if we could use some dumb TypeScript bullshit to force us to update them in parallel. It sucks to have to update code in multiple disparate places without an obvious link betwixt.

That being said, this solution is good enough that it can be polished up later

src/combat.ts
horrible-little-slime: yeah okay it absolutely sucks ass to have to pass `turnCount` into a billion different functions because of this change; make it an optional parameter and make it being defined part of the condition used to redigitize, I think.

horrible-little-slime (changes requested):
We've been traditionally using `camelCase` for files, so rename `variable_meat_potion` to `variableMeatPotion`. Also, add an `index.ts` to the `potions` and `count` subfolders, and make imports go through the index when possible (I acknowledge it is not always possible, that's the whole point of this refactor)

This is great stuff. I have some comments that are not related to this restructure that just came up from being forced to look at all this code, and I'll do my best to ping the relevant people for those. I also have some comments that are related to this refactor. Feel free to talk about them here or on discord.

src/counts/embezzler.ts
pstalcup: I think to do this you'd need to decouple the combat part of the embezzler fight from the count part of the embezzler fight (to not introduce a circular dependency where combat depends on the counts of embezzlers which itself depends on combats) 

maybe we should make 

currently we have 

```ts
export const chainStarters: EmbezzlerFight[] = []; 
export const copySources: EmbezzlerFight[] = []
```
Where each `EmbezzlerFight` knows about `Macro`

Could we instead do something like this?

```ts
export const chainStarters: (macro: MacroClass) => EmbezzlerFight[] = []; 
export const copySources: (macro: MacroClass) => EmbezzlerFight[] = []
```

so as to remove the dependency on Macro and instead have some interface that Macro satisfies?

src/combat.ts
docrostov: (note: when sweaty bill noted that this sucks ass, he did not mean the way you did this inherently sucks ass, just that it sucked that you had to do this due to the interconnectivity of garbo's systems lol)

src/counts/embezzler.ts
horrible-little-slime: it satisfies the base Macro class, that's for dingdang sure. But I'm not sure if we want to mix those

src/counts/embezzler.ts
xn: Typescript gonna typescript. I was leaving a little something to launch off of in case I needed to refactor EmbezzlerFight in this PR.

src/combat.ts
xn: I didn't take it negatively. No worries. 

There's a a few functions that we call multiple times and I wonder if we can call it once and store the result. `estimatedGarboTurns` is one of them.

src/counts/embezzler.ts
xn: I have to pop out for a few hours.

File renamed.

I'll get to the index files when I get back.

FYI, if it wasn't clear:
IGN: Chartreusenator
Disc: Cinchreusenator

src/counts/embezzler.ts
horrible-little-slime: you can do `const embezzlerCounts: NameCountPair[] = [` if you want it to force you to pay attention to the types

xn: done!

src/combat.ts
xn: done!

xn:
Also, to fully squash the cycles, yachtzee/buffs.ts and yachtzee/diet.ts should be merged into one file. I can't figure out a way to break that cycle as written.

src/potions/variable_meat_potion.ts
Pantocyclus: This function returns the sum of integers from `a` to `b` inclusive, with `a=0` by default
(It seems less intuitive to sum from `b` to `a`, setting `b=0` by default)

src/potions/variable_meat_potion.ts
Pantocyclus: We are averaging the price/value of an item over `quantity`. If the items have different prices over this range, then there exists individual items whose price is above the average.
The `1.2` is there to ensure that we can still buy the item above its average value (since we already bought quantities below its average value, and also determined that the entire bulk purchase makes sense [we don't want to make only partial purchases of variable meat potions])

Pantocyclus: To add on, we also need to be wary of price cliffs in case something goes terribly wrong (e.g. someone buys out the potions after we've done the averaging), so we definitely do not want a factor of, say, `10`.
I can't tell you whether `1.2` is better than `1.3` though

horrible-little-slime:
Just merged scepter. Still very interested in this. I have a bit of a busy weekend but I will likely re-refactor this to remove the turncount arg from meatkill. Again, stellar work, just great stuff

xn:
No worries and thanks! Please ping me when you do, I want to see your refactor. Doing this PR forced me to learn a ton about TS KOL development. Totally worth it.

xn:
@horrible-little-slime do you want me to rebase this?

horrible-little-slime:
I did not do the thing I thought I would do, but this is superceded by the libgarbo stuff

Still, this is great work and thank you so much for your contributions. It really helped Rever and I look at the code in question and plan out where we needed to put stuff!

PR #1542, opened 2023-09-03
do postCombatActions _after_ processing yachtzee information

horrible-little-slime:
interrupting yachtzee  with the "interrupt garbo" problem aborts before we update _stenchJellyUsed or _stenchJellyChargeTarget, which sucks!


docrostov (approved):
seems relatively straightforward and reasonable, don't really see a major issue here. feel free to wait for panto to merge but it looks like this very simple change will address the issue.

PR #1551, opened 2023-09-08
Use free skills for embezzler banish

Shiverwarp:
Right now this is mostly bespoke, but is using skills/items that are easily replenishable and available to everyone.

This could be more extensive, but I'd just like to get something started that works.

If you have greg fights, and they aren't more than our bespoke free handlers cover (2x snoke and 1x middle finger) it will banish bunnies with them instead of the very expensive ones like human musk etc


src/embezzler.ts
```diff
@@ -605,10 +605,13 @@ export const gregFights = (
       () => ((get(monsterProp) === embezzler && get(fightsProp) > 0) || totalCharges() > 0 ? 1 : 0),
       (options: EmbezzlerFightRunOptions) => {
         const run = ltbRun();
+        const runMacro = get(`_garboUsingFreeEmbezzlerBanish`)
```
horrible-little-slime, line 608:
```suggestion
        const runMacro = get("_garboUsingFreeEmbezzlerBanish", false)
```

src/extrovermectin.ts
```diff
@@ -334,6 +334,21 @@ const shortBanishes = [
 ];
 
 function banishBunny(): void {
+  if (get("_garboUsingFreeEmbezzlerBanish")) {
```
horrible-little-slime:
```suggestion
  if (get("_garboUsingFreeEmbezzlerBanish, false")) {
```

src/tasks/daily.ts
```diff
@@ -365,6 +367,21 @@ export function configureSnojo(): void {
...
+    gregLikeFightCount() > 0 &&
+    have($item`mafia middle finger ring`) &&
+    !get("_mafiaMiddleFingerRingUsed") &&
+    have($skill`Snokebomb`) &&
+    get(`_snokebombUsed`) === 0
+  ) {
+    set("_garboUsingFreeEmbezzlerBanish", true);
+    return;
+  }
+  set("_garboUsingFreeEmbezzlerBanish", false);
```
horrible-little-slime: capture the value of the condition in the `if` in a variable, and `set` the pref to that variable, instead of using the early return

horrible-little-slime: 120 is a magic number; either name it as a variable or comment as to how it came to be

horrible-little-slime (changes requested):
We should have some existing code that involves special casing snokebomb and/or hatred--can you double check if we do, and if so make all special casing around snokebomb account for eachother

also just some misc stuff

src/lib.ts
```diff
@@ -477,17 +477,44 @@ export function userConfirmDialog(msg: string, defaultValue: boolean, timeOut?:
...
+      Item | Familiar | Skill | (Item | Familiar | Skill)[],
+      () => boolean // function that returns true if we should disallow usage of the source while we're reserving embezzler banishers
+    >([
+      [$skill`Snokebomb`, () => get(`_snokebombUsed`) > 0], // We intend to save at least 2 uses for embezzlers, so if we've already used one, disallow usage.
+      [$item`mafia middle finger ring`, () => true],
+    ]);
+    const disallowUsage = props.get(action.source);
+    return !(disallowUsage?.() && get("_garboUsingFreeEmbezzlerBanish", false));
+  },
+};
```
horrible-little-slime, line 516: Don't duplicate this logic; find a way to build the latte constraints from the regular ones

src/lib.ts
```diff
@@ -477,17 +477,44 @@ export function userConfirmDialog(msg: string, defaultValue: boolean, timeOut?:
...
+      sum(forceEquipsOtherThanLatte, weaponHands) < 2 &&
+      !(disallowUsage?.() && get("_garboUsingFreeEmbezzlerBanish", false))
     );
   },
 };
 
+export const freeRunConstraints = {
+  allowedAction: (action: ActionSource) => {
+    const props = new Map<
+      Item | Familiar | Skill | (Item | Familiar | Skill)[],
```
horrible-little-slime, line 507:
```suggestion
      ActionSource["source"]
```

I think this is less distracting?

src/tasks/daily.ts
```diff
@@ -365,6 +368,22 @@ export function configureSnojo(): void {
...
+  if (freeBanishesChecked) return;
+  const useFreeBanishes =
+    getBanishedMonsters().get($item`ice house`) !== $monster`fluffy bunny` &&
+    gregLikeFightCount() < 120 && // 60 turns of banish from mafia middle finger ring, and 30 x 2 from two snokebombs
+    gregLikeFightCount() > 0 &&
+    have($item`mafia middle finger ring`) &&
+    !get("_mafiaMiddleFingerRingUsed") &&
+    have($skill`Snokebomb`) &&
+    get(`_snokebombUsed`) === 0
+      ? true
```
horrible-little-slime, line 381: It feels kind of silly to use ternary to decide a boolean; why would you do `condition ? true : false` when you can just do `condition`?

src/tasks/daily.ts
```diff
@@ -365,6 +368,22 @@ export function configureSnojo(): void {
   }
 }
 
+function determineFreeEmbezzlerBanish(): void {
```
horrible-little-slime, line 371: We aren't banishing embezzlers, we're banishing bunnies!

src/fights.ts
```diff
@@ -606,6 +607,17 @@ class FreeRunFight extends FreeFight {
...
+          const props = new Map<
+            Item | Familiar | Skill | (Item | Familiar | Skill)[],
+            () => boolean // function that returns true if we should disallow usage of the source while we're reserving embezzler banishers
+          >([
+            [$skill`Snokebomb`, () => get(`_snokebombUsed`) > 0], // We intend to save at least 2 uses for embezzlers, so if we've already used one, disallow usage.
+            [$item`mafia middle finger ring`, () => true],
+          ]);
+          const disallowUsage = props.get(action.source);
+          return !(disallowUsage?.() && get("_garboUsingFreeEmbezzlerBanish", false));
+        },
```
horrible-little-slime, line 620: You have written this code thrice now; it would be very col if we could get it down to once

src/fights.ts
```diff
@@ -665,8 +677,11 @@ const pygmySniffed = () =>
   sniffSources.some((source) => pygmyBanishHandlers.some(({ pygmy }) => pygmy === get(source)));
 
 const pygmyMacro = Macro.step(
-  ...pygmyBanishHandlers.map(({ pygmy, skill, item }) =>
-    Macro.if_(pygmy, skill ? Macro.trySkill(skill).item(item) : Macro.item(item)),
+  ...pygmyBanishHandlers.map(({ pygmy, skill, item, check, limit }) =>
+    Macro.if_(
+      pygmy,
+      skill && get?.(check) < limit ? Macro.trySkill(skill).item(item) : Macro.item(item),
```
horrible-little-slime, line 683:
```suggestion
      skill && get?.(check) < limit ? Macro.trySkill(skill).item(item) : Macro.item(item),
```
I think I don't quite understand what's happening here. `get` is a function that's always defined, right? why are we using optional chaning on it?

horrible-little-slime: In general, this seems like you're trying to do the 3-arg version of `Macro.externalIf`; I recognize that it doesn't behave super nicely with possibly-nullish parameters, but I think that would still be better

src/fights.ts
```diff
@@ -1021,7 +1037,12 @@ const freeFightSources = [
               $item`Louder Than Bomb`,
             ),
           )
-            .if_($monster`cat-alien`, Macro.trySkill($skill`Snokebomb`).tryItem($item`tennis ball`))
+            .if_(
+              $monster`cat-alien`,
+              get("_snokebombUsed") < snokeLimit
+                ? Macro.trySkill($skill`Snokebomb`).item($item`tennis ball`)
+                : Macro.item($item`tennis ball`),
```
horrible-little-slime, line 1044: This seems like a piece of logic we're using a lot; consider creating a bespoke Macro method for it. Or something.

src/fights.ts
```diff
@@ -1861,8 +1882,6 @@ const freeRunFightSources = [
               )
               .trySkill($skill`Summon Mayfly Swarm`),
           )
-          .if_($monster`Fruit Golem`, Macro.trySkill($skill`Feel Hatred`))
-          .if_($monster`Knob Goblin Mutant`, Macro.trySkill($skill`Snokebomb`))
```
horrible-little-slime, line 1865: where did our bespoke handling for these beloved friends go?

src/tasks/daily.ts
```diff
@@ -365,6 +368,22 @@ export function configureSnojo(): void {
...
+  if (freeBanishesChecked) return;
+  const useFreeBanishes =
+    getBanishedMonsters().get($item`ice house`) !== $monster`fluffy bunny` &&
+    gregLikeFightCount() < 120 && // 60 turns of banish from mafia middle finger ring, and 30 x 2 from two snokebombs
```
horrible-little-slime, line 375: So one issue here is that we don't _just_ care about the greglike embezzlers, we care about all embezzlers we do before our final gregory. That includes profchain, for example, as well as our chain starter and probably a few interpolated digitizes

horrible-little-slime (changes requested):

src/tasks/daily.ts
```diff
@@ -643,6 +669,11 @@ const DailyTasks: GarboTask[] = [
     completed: () => snojoConfigured,
     do: configureSnojo,
   },
+  {
+    name: "Determine whether to use free banishers for embezzlers",
+    completed: () => freeBanishesChecked,
+    do: determineFreeBunnyBanish,
```
s-k-z, line 675: Does this need to be a grimoire task? It doesn't actually make any network queries as far as I can tell nor does it read or change the state of the game, only the state of kolmafia.

horrible-little-slime: So I'm not sure how viable it is, but I would basically have it get calculated the first time we ask about this information, instead of calculating it separately. Like a getter

src/embezzler.ts
```diff
@@ -633,6 +635,30 @@ export const gregFights = (
...
+    "Habitats Monster",
+    () => have($skill`Just the Facts`),
+    "_monsterHabitatsMonster",
+    "_monsterHabitatsFightsLeft",
+    () =>
+      have($skill`Just the Facts`)
+        ? (3 - get("_monsterHabitatsRecalled")) * 5 + get("_monsterHabitatsFightsLeft")
+        : 0,
+  ),
+];
```
horrible-little-slime, line 656: This looks like duplicated code, isn't it?

If you're going to make this its own source for counting, you should then plop this into the conditionalSources array, instead of us generating copies of the same fights there

src/extrovermectin.ts
```diff
@@ -334,22 +334,38 @@ const shortBanishes = [
...
+        preventEquip: $items`carnivorous potted plant`,
+        forceEquip: $items`mafia middle finger ring`,
+      }).maximize();
+      garboAdventure(
+        $location`The Dire Warren`,
+        Macro.if_($monster`fluffy bunny`, Macro.skill($skill`Show them your ring`)).embezzler(),
+      );
+    } while (
+      "fluffy bunny" !== get("lastEncounter") &&
+      !get("banishedMonsters").includes("fluffy bunny")
```
horrible-little-slime, line 349: I'm not sure how I feel about separating this from the main body of the code instead of formatting the free banish like the other banishes and forcing it to be chosen if available (which will already happen, because it's free)

src/lib.ts
```diff
@@ -485,18 +488,65 @@ export function userConfirmDialog(msg: string, defaultValue: boolean, timeOut?:
...
-    );
-  },
-};
+function determineFreeBunnyBanish(): boolean {
+  const expectedPocketProfFights = !have($familiar`Pocket Professor`)
+    ? 0
+    : (!get("_garbo_meatChain", false) ? Math.max(10 - get("_pocketProfessorLectures"), 0) : 0) +
+      (!get("_garbo_weightChain", false) ? Math.min(15 - get("_pocketProfessorLectures"), 5) : 0);
+  const expectedDigitizesDuringGregs =
+    SourceTerminal.have() && get("_sourceTerminalDigitizeUses") < 3 ? 3 : 0; // To encounter 3 digitize monsters it takes 91 adventures. Just estimate we fight all 3 to be safe.
```
horrible-little-slime, line 497: Yeah that is fine, we can do a more precise estimate but it's better to err on the side of caution

src/lib.ts
```diff
@@ -81,12 +82,14 @@ import {
...
   set,
   SongBoom,
+  SourceTerminal,
   sum,
   uneffect,
 } from "libram";
 import { acquire } from "./acquire";
 import { globalOptions } from "./config";
 import { garboValue } from "./value";
+import { gregLikeFightCount } from "./embezzler";
```
horrible-little-slime, line 92: We do not want to do this. I'm not sure where the best place to put the bunny banish functions given this, but we don't want to import embezzler to lib

src/fights.ts
```diff
@@ -667,8 +669,12 @@ const pygmySniffed = () =>
   sniffSources.some((source) => pygmyBanishHandlers.some(({ pygmy }) => pygmy === get(source)));
 
 const pygmyMacro = Macro.step(
-  ...pygmyBanishHandlers.map(({ pygmy, skill, item }) =>
-    Macro.if_(pygmy, skill ? Macro.trySkill(skill).item(item) : Macro.item(item)),
+  ...pygmyBanishHandlers.map(({ pygmy, skill, item, check, limit }) =>
```
horrible-little-slime, line 672: god this is so many vars this isn't a review comment just jesus christ i forgot how terrible pygmies are

horrible-little-slime (approved):
This looks good but I think I would like one other person to take a look at this just because I have been driven to madness

src/lib.ts
```diff
@@ -485,18 +489,82 @@ export function userConfirmDialog(msg: string, defaultValue: boolean, timeOut?:
...
-      sum(forceEquipsOtherThanLatte, weaponHands) < 2
-    );
-  },
-};
+function determineFreeBunnyBanish(): boolean {
+  const extraOrbFights = have($item`miniature crystal ball`) ? 1 : 0;
+  const possibleGregsFromSpleen =
+    Math.floor((spleenLimit() - mySpleenUse()) / 2) * (3 + extraOrbFights);
+  const currentAvailableGregs = get("beGregariousCharges") * (3 + extraOrbFights);
+  const habitatFights = (3 - get("_monsterHabitatsRecalled")) * (5 + extraOrbFights);
```
s-k-z, line 497: I think it'd be good to clamp these values between 0 and 3 in case a user sets them to values outside the normal range.

s-k-z (dismissed):

horrible-little-slime (approved):

PR #1552, opened 2023-09-09
Ask for Ode to Booze buff when without skill

eegeeZA:
Allow those without Ode to Booze to request it from the generous Buffy bot.

### DIFF
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -114,8 +114,8 @@ function eatSafe(qty: number, item: Item) {
 
 function drinkSafe(qty: number, item: Item) {
   const prevDrunk = myInebriety();
+  const odeTurns = qty * item.inebriety;
   if (have($skill`The Ode to Booze`)) {
-    const odeTurns = qty * item.inebriety;
     const castTurns = odeTurns - haveEffect($effect`Ode to Booze`);
     if (castTurns > 0) {
       useSkill(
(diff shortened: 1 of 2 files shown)


s-k-z (changes requested):
Please switch it to one of the other buffbots. Buffy is an altruistic service provided free of charge which is best left for newer players and those not automating the game.

Malibu-Stacey:
I am vehemently against this.
I suggest you try playing the game for a change instead of spending 17 years meatfarming.

eegeeZA:
@s-k-z I'll do some research and then revert with the changes.

@Malibu-Stacey I can understand your stance, but I do believe different people play the game differently. The intent behind this change is not break the way you play the game but to introduce an alternative style for others.

eegeeZA:
@s-k-z will `Flesh Puppet` be a reasonable alternative?

gausie:
I think you need to make a proactive case for the changes you've made to mitigate the negative effect your change would have on the community.

eegeeZA:
Okay sure. My account is 17 years old, and I've never ascended. I started playing more actively again since Crimbo last year, after quite a long hiatus. My class is a Seal Clubber, and I don't have access to Ode to Booze by any means, besides requesting buffs. Ascending now would mean having to lose 92 skills after perming one. I really enjoy using garbo for some meat farming.

If I understand the concern correctly, adding this change might discourage others from ascending? Or is it that making use of a buff bot as part of automation is frowned upon?

s-k-z:
Personally I don't have strong opinions about paying for buffs through automation. Using a free service that consumes the resources of the provider is not acceptable, similar to how we don't automatically grab a carpe from the VIP Floundry. I believe Flesh Puppet is another free service, just like Buffy.

Discouraging others from ascending is another big issue, and we really don't want to promote that. Players who go deeper into that hole will essentially be trapped, as you may find yourself. Basically, we should always send the message that it is better to ascend sooner rather than later.

PR #1558, opened 2023-09-13
stop garbo from infinite loop meat loss

docrostov:
garbo has trouble handling misaligned gregs and the way we have handled this has led to a very bad nigh-infinite meat loss problem where the system, if your greg count gets off, just continually buys and throws LTBs. this has led to an unfortunate -70 mil meat day for kas and -30 for another user. 

this is not a great solution, but it is a simple one -- check garbovalue pre-greg and abort if garbo is more than 3 million meat in the red. even for some theoretical garbo user who is at t0 and hasn't made a single bit of meat, diet never costs more than 1.5-2 mil, so if garbo is at -3 something has gone horribly wrong and it is continually cycling some bit of code. realistically this may even be worth abstracting out of embezzies into a general purpose function called after freefights have been run but for now this feels like a very simple fix for very bad behavior

### DIFF
--- a/src/embezzler.ts
+++ b/src/embezzler.ts
@@ -68,6 +68,7 @@ import { waterBreathingEquipment } from "./outfit";
 import wanderer, { DraggableFight } from "./wanderer";
 import { MonsterProperty, NumericProperty } from "libram/dist/propertyTypes";
 import { shouldAugustCast } from "./resources";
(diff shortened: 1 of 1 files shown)


docrostov:
closing because phred is going to solve this and i misread garbovalue

PR #1564, opened 2023-09-14
Sometimes stomach goes over max and that is fine

gausie:


src/lib.ts
```diff
@@ -386,12 +401,7 @@ export function safeRestore(): void {
   const mpTarget = safeRestoreMpTarget();
   const shouldRestoreMp = () => myMp() < mpTarget;
 
-  if (
-    shouldRestoreMp() &&
```
s-k-z, line 390: This needs to stay

PR #1567, opened 2023-09-14
Create preference to control how many times we source an embezzler

libraryaddict:


mafia/data/garbo_settings.json
```diff
@@ -51,6 +51,11 @@
     "type": "boolean",
     "description": "**WARNING: Experimental** Don't show user confirm dialogs, instead automatically select yes/no in a way that will allow garbo to continue executing. Useful for scripting/headless. Risky and potentially destructive."
   },
+  {
+    "name": "garbo_autoUserConfirm_embezzlerInvocationsCount",
```
s-k-z, line 55: I think this should be changed to `garbo_autoUserConfirm_embezzlerInvocationsThreshold`.

mafia/data/garbo_settings.json
```diff
@@ -51,6 +51,11 @@
...
+    "defaultValue": 1,
+    "description": "This is used in conjunction with autoUserConfirm, will automatically use resources this many times to invoke an embezzler before requesting user interference."
```
s-k-z, line 57: I think the description should be more explicit that it only functions if autoUserConfirm is enabled, and that invoking means wishing, using a clover, etc.

s-k-z: Also, it should specify that it is up to that many times, not guaranteed that many times.

s-k-z (approved):

PR #1568, opened 2023-09-14
Fix css

libraryaddict:


gausie (approved):
Thank you, this was totally untested

Issue #1543, opened 2023-09-04
Diet Loops When Fullness Exceeds Limit


Alium58:
Post small run my stomach was 20/15. Diet looped on this:
![image](https://github.com/loathers/garbage-collector/assets/4781473/8e79ce15-c9eb-4ffc-b322-c264ab14f27f)
[Garbo Fullness Loop - Trimmed.txt](https://github.com/loathers/garbage-collector/files/12514839/Garbo.Fullness.Loop.-.Trimmed.txt)

Impact is large session file. Yesterday mine was 9GB. Perhaps an abort is appropriate if fullness > limit with some description to avoid large session log and let the user know why it failed. Or maybe it is an easy fix in the diet code, I'm not familiar with it.

PR #1570, opened 2023-09-15
Safely diet even if fullness exceeds capacity after a Smol run

danielwhite:
This targets #1543 where it's possible to have 20 / 15 fullness after breaking the prism on a [Smol](https://kol.coldfront.net/thekolwiki/index.php/A_Shrunken_Adventurer_Am_I).

This aims to fix to the capacity calculation when evaluating diet entries so that negative capacity isn't possible.

As a means of avoiding surprise loops, `consumeSafe` will now barf on negative quantities.

### DIFF
--- a/src/diet.ts
+++ b/src/diet.ts
@@ -136,6 +136,9 @@ function chewSafe(qty: number, item: Item) {
 }
 
 function consumeSafe(qty: number, item: Item, additionalValue?: number, skipAcquire?: boolean) {
+  if (qty <= 0) {
+    throw `Trying to consume ${qty} of ${item.name}`
+  }
   const spleenCleaned = spleenCleaners.get(item);
   if (spleenCleaned && mySpleenUse() < spleenCleaned) {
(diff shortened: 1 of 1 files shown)


danielwhite:
Turned of Yachtzee to test this. #1564 seems to have a handle on it (although it didn't quite work for me).

On the one test, I ran into an error with it planning to eat more than I could. Toasted brie for some reason. Error isn't in my session log, so I'll have to try again after the end of my next run.

gausie:
> Turned off Yachtzee to test this. #1564 seems to have a handle on it (although it didn't quite work for me).
> 
> On the one test, I ran into an error with it planning to eat more than I could. Toasted brie for some reason. Error isn't in my session log, so I'll have to try again after the end of my next run.

That is a libram issue that #1569 will fix. I'll close this but thank you for writing some code all the same, we appreciate it!

danielwhite:
Whoops. Totally glossed over that dot point. 🤦

PR #1578, opened 2023-09-19
support snapper as a familiar

pstalcup:


src/familiar/freeFightFamiliar.ts
```diff
@@ -69,6 +81,46 @@ export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
...
       });
     }
+    if (
+      location === $location`Barf Mountain` &&
+      Snapper.have() &&
+      Snapper.getTrackedPhylum() === $phylum`dude`
+    ) {
+      /*
+      # E stands for olfacted Garbage Tourist, A is angry toursit, F is horrible tourist family
+      import itertools
```
gausie, line 91: Write pseudocode in typescript (or not at all!) If possible

pstalcup: this isn't psuedo code - its a literal python function

horrible-little-slime: imo this is fine

horrible-little-slime (dismissed):
Tentative approval. I'm concerned there may be other queue manipulators not being accounted for in your python script--it doesn't look like it checks for both the turtlesexy and the turtlesexless. But also I think we can just assume people have turtlesex; it's 300k and an autoperm, apparently

src/lib.ts
```diff
@@ -683,3 +687,80 @@ export function printEventLog(): void {
...
+export function barfEncounterRate(options: {
+  snapper?: boolean;
+  snapperPhylum?: Phylum;
+  olfact?: Monster;
+  longCon?: Monster;
+  motif?: Monster;
+  turtle?: Monster;
+  monkeyPoint?: Monster;
+  humanity?: boolean;
+}): Map<Monster, number> {
```
horrible-little-slime, line 702: just make this a `Partial`

src/lib.ts
```diff
@@ -683,3 +687,80 @@ export function printEventLog(): void {
...
+  const olfact = options.olfact ?? get("olfactedMonster");
+  const longCon = options.longCon ?? get("longConMonster");
+  const motif = options.motif ?? get("motifMonster");
+  const turtle = options.turtle ?? get("_gallapagosMonster");
+  const monkeyPoint = options.monkeyPoint ?? get("monkeyPointMonster");
+  const snapper = options.snapper ?? myFamiliar() === $familiar`Red-Nosed Snapper`;
+  const snapperPhylum = options.snapperPhylum ?? Snapper.getTrackedPhylum();
+  const humanity = options.humanity ?? have($effect`Ew, The Humanity`);
```
horrible-little-slime, line 710: imo i would refactor this as something like
```suggestion
  const DEFAULT_OPTIONS = {
    olfact: get("olfactedMonster"),
    longCon: get("longConMonster"),
    etc,
   };
   const { olfact, longCon, motif, etc } = { ...DEFAULT_OPTIONS, ...options };
```
but if that looks a lot don't do that

src/lib.ts
```diff
@@ -683,3 +687,80 @@ export function printEventLog(): void {
...
+    .map((v) => `${v}`)
+    .join(":");
+
+  const cachedValue = rateCache.get(cacheKey);
+  if (cachedValue) {
+    return cachedValue;
+  }
+
+  const copies = (target: Monster | null, n: number): Monster[] =>
+    n === 0 ? [] : [...zoneMonsters.filter((m) => m === target), ...copies(target, n - 1)];
```
horrible-little-slime, line 737: I feel like a `.fill` solution will be more performant and legible than the recursive one

src/lib.ts
```diff
@@ -683,3 +687,80 @@ export function printEventLog(): void {
...
+
+  const monsterQueue = [
```
horrible-little-slime, line 739: call this the CSV, not the queue. Or call it some other thing. But don't call it the queue

horrible-little-slime: `monsterPool`?

src/lib.ts
```diff
@@ -683,3 +687,80 @@ export function printEventLog(): void {
...
+    humanity,
+    ...encounterQueue,
+  ]
+    .map((v) => `${v}`)
+    .join(":");
+
+  const cachedValue = rateCache.get(cacheKey);
+  if (cachedValue) {
+    return cachedValue;
+  }
```
horrible-little-slime, line 734:
```suggestion
  if (cachedValue) return cachedValue;
```

src/lib.ts
```diff
@@ -683,3 +687,80 @@ export function printEventLog(): void {
...
+  const monsterQueue = [
+    ...zoneMonsters,
+    ...copies(olfact, 3),
+    ...copies(longCon, 3),
+    ...copies(motif, 2),
+    ...copies(turtle, 1),
+    ...copies(monkeyPoint, 2),
+    ...zoneMonsters
+      .filter((m) => snapper && m.phylum === snapperPhylum)
+      .flatMap((m) => copies(m, 2)),
```
horrible-little-slime, line 748: I have had trouble with polyfilling `flatMap` before; definitely make sure this works the way you want it to

src/lib.ts
```diff
@@ -683,3 +687,80 @@ export function printEventLog(): void {
...
+    ...zoneMonsters
+      .filter((m) => humanity && m.phylum === $phylum`dude`)
+      .flatMap((m) => copies(m, 2)),
+  ];
+
+  const encounters = monsterQueue.flatMap((m) =>
+    monsterQueue.map((n) =>
+      // olfaction, longcon, and motif caus that monster to ignore queue rejection
+      olfact === m || longCon === m || motif === m || !encounterQueue.includes(m) ? m : n,
```
horrible-little-slime, line 757:
```suggestion
      [olfact, longCon, motif].includes(m) && !encounterQueue.includes(m) ? m : n,
```

src/lib.ts
```diff
@@ -683,3 +687,80 @@ export function printEventLog(): void {
...
+  const motif = options.motif ?? get("motifMonster");
+  const turtle = options.turtle ?? get("_gallapagosMonster");
+  const monkeyPoint = options.monkeyPoint ?? get("monkeyPointMonster");
+  const snapper = options.snapper ?? myFamiliar() === $familiar`Red-Nosed Snapper`;
+  const snapperPhylum = options.snapperPhylum ?? Snapper.getTrackedPhylum();
+  const humanity = options.humanity ?? have($effect`Ew, The Humanity`);
+
+  const zoneMonsters = $monsters`garbage tourist, angry tourist, horrible tourist family`;
+  const encounterQueue = zoneMonsters.filter((m) =>
+    $location`Barf Mountain`.combatQueue.includes(`${m}`),
```
horrible-little-slime, line 714: I think I prefer `m.name` here but I think using the template literal below makes sense so maybe being consistent within this area is better

src/lib.ts
```diff
@@ -683,3 +687,80 @@ export function printEventLog(): void {
...
+    ...copies(monkeyPoint, 2),
+    ...zoneMonsters
+      .filter((m) => snapper && m.phylum === snapperPhylum)
+      .flatMap((m) => copies(m, 2)),
+    ...zoneMonsters
+      .filter((m) => humanity && m.phylum === $phylum`dude`)
+      .flatMap((m) => copies(m, 2)),
+  ];
+
+  const encounters = monsterQueue.flatMap((m) =>
```
horrible-little-slime, line 754: I don't understand what this variable is meant to represent

horrible-little-slime: update: I think I might understand it, but if I do then it's not quite right: we only reroll 75% of the time, but even a reroll can reroll.

This math is really hard to do--slaw wrote a big ol' thesis about it. I think we have to do it numerically instead of analytically.

src/lib.ts
horrible-little-slime: a lot worse*

horrible-little-slime:
I know that you're working on a whole new thing for this, but just as a reminder one decision we made in discord is that alongside this we should _always_ print marginal familiar printouts in barf, not just when we pick marginal over it, and for snapper specifically we should add a `based on queue-state` message or something that helps users understand why garbo keeps "changing its mind"

src/familiar/freeFightFamiliar.ts
```diff
@@ -69,6 +80,25 @@ export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
...
+      location === $location`Barf Mountain` &&
+      Snapper.have() &&
+      Snapper.getTrackedPhylum() === $phylum`dude`
+    ) {
+      const encounterRate = barfEncounterRate({ snapper: true, snapperPhylum: $phylum`dude` });
+      const dudeRate = [...encounterRate.entries()].reduce(
+        (acc: number, entry: [Monster, number]) =>
+          entry[0].phylum === $phylum`dude` ? entry[1] + acc : acc,
+        0,
+      );
```
horrible-little-slime, line 93:
```suggestion
      const dudeRate = sum([...encounterRate.entries()], ([monster, weight]) => monster.phylum === $phylum`dude` ? weight : 0);
```

PR #1580, opened 2023-09-22
Add Scepter Free Fights When Today

Tokoeka:
additional free fight(s) 2 days out of the month feels worth it


src/resources/scepter.ts
```diff
@@ -19,10 +19,12 @@ import { Potion } from "../potions";
 import { garboAverageValue, garboValue } from "../value";
 import { canAdventure, canEquip, Item, myLevel, myMeat, Skill, toSlot, useSkill } from "kolmafia";
 
+const FFV = globalOptions.prefs.valueOfFreeFight;
```
horrible-little-slime, line 22: We actually shouldn't fix this as a const right up top here. The way grimoire args works, this will be defined before args are filled, and we'll be referencing the default (or possibly just a null value?) value of the variable rather than the correct value

src/resources/scepter.ts
```diff
@@ -140,7 +152,7 @@ const SKILL_OPTIONS: ScepterSkill[] = [
 let bestScepterSkills: ScepterSkill[] | null = null;
 function getBestScepterSkills(): ScepterSkill[] {
   return (bestScepterSkills ??= SKILL_OPTIONS.filter(
-    ({ skill }) => AugustScepter.todaysSkill() !== skill,
+    ({ skill, type }) => AugustScepter.todaysSkill() !== skill && type !== "fight",
```
horrible-little-slime, line 155: I understand what you're doing here, I'm just not sure if I agree. If the best skill to cast is a fight, let them fight!

src/fights.ts
```diff
@@ -750,6 +751,16 @@ const freeFightSources = [
     () => withMacro(Macro.basicCombat(), () => use($item`molehill mountain`)),
     true,
   ),
+  new FreeFight(
```
horrible-little-slime, line 754: I personally would have done a like `...getBestScepterSkills().map(({ skill }) => ` whatever, but given that we're only supporting 2 fights this is fine

horrible-little-slime (changes requested):
I was personally holding off until after we grimoirize free fights to do this but also I didn't realize how short and sweet it would be, so absolutely no reason to hold off.

But, one change requested, and it's a pretty important (but extremely simple) one

src/resources/scepter.ts
```diff
@@ -19,10 +19,12 @@ import { Potion } from "../potions";
 import { garboAverageValue, garboValue } from "../garboValue";
 import { canAdventure, canEquip, Item, myLevel, myMeat, Skill, toSlot, useSkill } from "kolmafia";
 
+const FFV = globalOptions.prefs.valueOfFreeFight;
```
Rinn, line 22:
```suggestion
```

src/resources/scepter.ts
```diff
@@ -62,6 +64,11 @@ const SKILL_OPTIONS: ScepterSkill[] = [
       }).gross(embezzlerCount()), // TODO: Yachtzee
     type: "buff",
   },
+  {
+    skill: $skill`Aug. 8th: Cat Day!`,
+    value: () => FFV,
```
Rinn, line 69:
```suggestion
    value: () => globalOptions.prefs.valueOfFreeFight,
```

src/resources/scepter.ts
```diff
@@ -90,6 +97,11 @@ const SKILL_OPTIONS: ScepterSkill[] = [
     value: () => 3000, // Dummy value; we should some day calculate this based on free fight count, careful to avoid circular imports
     type: "buff",
   },
+  {
+    skill: $skill`Aug. 22nd: Tooth Fairy Day!`,
+    value: () => FFV,
```
Rinn, line 102:
```suggestion
    value: () => globalOptions.prefs.valueOfFreeFight,
```

Rinn:
I'll resolve the conflicts

Rinn (approved):

Rinn:
I jacked up my value of free fight and it worked, merging

PR #1588, opened 2023-09-28
Add a turns spent print out

pstalcup:


horrible-little-slime (dismissed):
o heck yeah he generalized it

PR #1592, opened 2023-09-30
bofa value

horrible-little-slime:
while I can use the `factType` proxy record in the switch--and doing so protects me from having to count pocket wish and tatter drops--wanting to check quantities of items and effects forces me to use the `itemFact`, `numericFact`, and `effectFact` functions instead of proxy records. Which is fine, but a little awkward given that I have to pass in the seed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "dependencies": {
     "core-js": "^3.32.2",
     "grimoire-kolmafia": "^0.3.22",
-    "kolmafia": "^5.27598.0",
+    "kolmafia": "^5.27625.0",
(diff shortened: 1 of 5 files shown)


gausie:
Maybe I'll just add those function overrides. Fwiw you can also Parse the .fact for duration and quantity and such. But ok I'll make the mafia change when I have time. Or someone else do it!

horrible-little-slime:
I think parsing it is more trouble than it's worth, because of turncounts
for effects and item quantities for items. Not an impossible regexp, but
this feels easier for now

pstalcup (approved):

PR #1597, opened 2023-10-03
Get continuum transfunctioner before Shadow Forest

Shiverwarp:


spaghetti-squash (commented):
this is probably fine, but it does also raise the question of why things other than `Ascending` and `Volcano` are their own quests. The thematic grouping is nice for chunking stuff up but doesn't seem to provide any tangible benefit otherwise.

horrible-little-slime:
There is a grimoire system in place for this, I think it is called `after`.
None of my scripts use it but it seems wise here

PR #1603, opened 2023-10-04
Grimoirize Yachtzee

pstalcup:


src/yachtzee/index.ts
```diff
@@ -1,163 +1,116 @@
...
+        postCombatActions();
+      },
+    },
+    {
+      name: "Clara's Bell",
+      ready: () => have($item`Clara's bell`) && !get("_claraBellUsed"),
+      completed: () => get("noncombatForcerActive"),
+      do: () => use($item`Clara's bell`),
+    },
+  ];
```
horrible-little-slime, line 102: We seem to both be saving cinch-forces when SBB is open _and_ not using them here. We should steamline that logic--I think not using cinch makes sense (I'm not sure if it's right, but it makes sense), but not the way things are currently set up

src/yachtzee/index.ts
```diff
@@ -1,163 +1,116 @@
...
-  if (get("_yachtzeeChainClosetedMeat", 0)) {
-    cliExecute(`closet take ${get("_yachtzeeChainClosetedMeat")} meat`);
-  }
-  set("_yachtzeeChainClosetedMeat", 0);
-  if (haveEffect($effect`Beaten Up`)) {
-    uneffect($effect`Beaten Up`);
-  }
-  meatMood(false, 2000).execute(Math.min(jellyTurns, fishyTurns));
-  safeRestore();
+  return [
```
horrible-little-slime, line 44: This needs handling for digitized embezzlers, which will ruin our parade. We dont' want to free run from them and we don't want to hit them in the sunken party yacht.

The smartest and coolest thing to do would be to add a check to the embezzler underwater code that says that if we have SBB and NC forcers, we _don't_ get fishy during that stint, and we instead assume that we'll do the envyfish later in the day.

It would be nice if we started saving an envyfish egg for the next day. Is that a thing we can do?

horrible-little-slime: Also kramco interpolation

horrible-little-slime (changes requested):
Rever, I'm making this review as a record so we can all work on this--I don't expect all of these changes to be shouldered by you, and I plan to hit several of them myself this weekend.

src/yachtzee/index.ts
```diff
@@ -1,163 +1,116 @@
...
+      ready: () => have($item`fishy pipe`) && !have($effect`Fishy`),
+      completed: () => get("_fishyPipeUsed"),
+      do: () => use($item`fishy pipe`),
+    },
+    {
+      name: "Spikolodon Spikes",
+      ready: () => have($item`Jurassic Parka`) && get("_spikolodonSpikeUses") < 5,
+      completed: () => get("noncombatForcerActive"),
+      do: () => {
+        const run = tryFindFreeRun(freeRunConstraints(false)) ?? ltbRun();
```
horrible-little-slime, line 78: imo we should make this a `let` variable outside, initialize it as part of the start of the task and dump it at the end. that way we can go through traditional outfit and combat code for grimoire. freecandy does this [here](https://github.com/loathers/freecandydotexe/blob/ef62f543e34330a9467ce4d9a614f0db85def5ee/src/regularTasks.ts#L74-L89), and it gets used [here](https://github.com/loathers/freecandydotexe/blob/ef62f543e34330a9467ce4d9a614f0db85def5ee/src/regularTasks.ts#L316-L343)

Rinn:
Should prob move the umdLastObtained logic from fights.ts into this

horrible-little-slime:
(Waiting on #1609 before continuing)

src/yachtzee/index.ts
horrible-little-slime: Also romanced embezzlers :-/

PR #1607, opened 2023-10-05
fix: check whether bunny is still banished instead of was ever banished

midgleyc:
The banished monsters preference isn't updated until you banish another monster. Because of this, the fluffy bunny can be present in the list of banished monsters after a used banish has expired.

### DIFF
--- a/src/embezzler/fights.ts
+++ b/src/embezzler/fights.ts
@@ -5,6 +5,7 @@ import {
   cliExecute,
   getClanLounge,
   haveEquipped,
+  isBanished,
   itemAmount,
   Location,
   mallPrice,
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
dumb little question: is the banished monsters pref ordered? should we try to find the rightmost entry instead of the leftmost, to ensure we're talking about the most recent banish?

midgleyc:
it is ordered by recency, but I think you want the first banish. The latest banish will be the one you did in the fight you just had (unless I precalculate `bunnyBanish`), which might make the error message look a bit weird.

Also, there's not much difference between the first and latest banishes? The error message could feasibly show all the banishes. Or none, because the useful bit of information is "try setting this preference to fix it".

horrible-little-slime:
> I feel like a `string[] banished_by(monster)` function might not go amiss in Mafia for this, too. Especially with phyla banishes existing.

Honestly I'd personally prefer something that returns the Items and Skills themselves, but because ash doesn't have real union types it would either need to be a record (feels silly) or be a set of functions like bofa or cargo shorts (also feels silly), which might mean that the idea is silly on the whole. We can always wrap it, in libram.

would `banished_by` appropriately account for banishes expiring?

midgleyc:
yeah, otherwise it would be pointless.

BanishManager itself doesn't track the items and skills -- that's purely a libram construction.

midgleyc:
Like, forget _ASH_ not being able to do "item or skill" -- _Java_ can't do that. So without an `ItemOrSkill` class or something other ridiculous construction, it's not happening in Mafia ;)

PR #1611, opened 2023-10-08
Add habitats to replacer embezzler fight potential

Shiverwarp:


src/embezzler/fights.ts
```diff
@@ -617,7 +617,9 @@ export const conditionalSources = [
       proceedWithOrb(),
     () =>
       ((get("beGregariousMonster") === embezzler && get("beGregariousFightsLeft") > 0) ||
-        get("beGregariousCharges") > 0) &&
+        get("beGregariousCharges") > 0 ||
+        (get("_monsterHabitatsMonster") === embezzler && get("_monsterHabitatsFightsLeft") > 0) ||
+        (have($skill`Just the Facts`) && get("_monsterHabitatsRecalled") < 3)) &&
```
horrible-little-slime, line 622: use `doingGregFight()` instead of recreating that logic

horrible-little-slime: (and the same below)

horrible-little-slime:
I have a mild concern that this will make counts less accurate for some people, but I think on the aggregate it makes things more accurate. Still, it would be nice to just fix the whole thing in one swoop instead of inching our way there.

pstalcup (approved):

PR #1621, opened 2023-10-10
garbo-lib step 3

gausie:
This step is all about taking advantage of monorepos

- Move libgarbo to garbo-lib
- Move relay to garbo-relay
- Simplify build file for garbo
- Add formatting across all


horrible-little-slime:
@gausie do we want to add `.yarn` to `gitignore`? Seems like there's a huge number of lil binary files clogging things up

horrible-little-slime (approved):

PR #1626, opened 2023-10-11
Pre-commit hooks and some other tidying

gausie:
- Remove unused pre-commit config (lets use husky), add interactive tools for yarn, remove unused babel transforms
- Remove cache stuff for removed babel transforms
- Add husky, lint-staged
- Remove user-specific vscode settings


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
Rinn, line 1: is this usable on windows?

packages/garbo-lib/package.json
```diff
@@ -7,7 +7,14 @@
   "scripts": {
     "build": "echo \"Nothing to do\"",
     "format": "eslint src --fix && prettier --write .",
-    "lint": "eslint src && prettier --check ."
+    "lint": "eslint src && prettier --check .",
+    "pre-commit": "lint-staged"
```
Rinn, line 11: Should probably still have madge to check for circular dependencies (also in CI?)

gausie: Yeah sure. I'll do it later though.

.husky/pre-commit
gausie: Yes

PR #1628, opened 2023-10-11
feat: split the quick flag into quickcombat and quickgear

midgleyc:
Sometimes (okay, all the time) I would like garbo to take half the time to run in return for losing about 3k profit. That means I want to lower the number of stasis turns, but not any other "quick" stuff. Rever suggested splitting the flag out, which this PR does.

Remove `quick`'s influence on historical prices due to #1617. If this is wrong, let me know, I'll add a `quickprices` flag.

Mark `quick` as no longer experimental. I can flip this back but I figure it's been out and used for over a year, and it works pretty well.


horrible-little-slime:
Is it worth adding some messaging if someone tries to pass the old `quick` arg? we can remove it in a few weeks, just to get the word out

packages/garbo/src/config.ts
```diff
@@ -229,3 +239,17 @@ You can use multiple options in conjunction, e.g. "garbo nobarf ascend"',
...
+  return opts.quick || opts.quickgear || opts.quickcombat;
+}
+
+export function isQuickGear(opts: GlobalOptions): boolean {
+  return opts.quick || opts.quickgear;
+}
+
+export function isQuickCombat(opts: GlobalOptions): boolean {
+  return opts.quick || opts.quickcombat;
+}
```
horrible-little-slime, line 255: We're only ever going to have one `globalOptions` object--why pass this as an arg instead of just making these functions reference the thing to begin with?

packages/garbo/src/session.ts
```diff
@@ -309,7 +309,7 @@ export function endSession(printLog = true): void {
     message("So far today", totalTurns, totalMeat, totalItems);
 
     printMarginalSession();
-    if (globalOptions.quick) {
+    if (isAnyQuick(globalOptions)) {
```
horrible-little-slime, line 312: I'm not sure this is still true, because of the historical price thing. we should just remove this messaging

horrible-little-slime (changes requested):
Okay upon reading the PR I take it back, no need to add messaging around `quick` if it's still a flag.

PR #1631, opened 2023-10-12
Grimoirize post.ts

horrible-little-slime:
Stillsuit should be handled on the engine imo

I'm making task-returning functions rather than raw tasks because I think we'll for a while be copying this quest a few places, and that feels better?

Updating session items should similarly be handled on the engine probably.


Rinn (dismissed):

packages/garbo/src/tasks/post/post.ts
```diff
@@ -0,0 +1,272 @@
...
+    spendsTurn: false,
+  };
+}
+
+const BARF_PLANTS = [
+  FloristFriar.StealingMagnolia,
+  FloristFriar.AloeGuvnor,
+  FloristFriar.PitcherPlant,
+];
+function floristFriars(): GarboTask {
```
pstalcup, line 66: Since we're going to use functions to return these, I would prefer to fully exclude post tasks that will never be applicable. 

I'd recommend something like making all these return 
```ts
GarboTask & { allowed: () => boolean } 
```
For Florist Plants
```
{
allowed: () => FloristFriar.have()
}
```

and then later on, just doing a filter
```ts
tasks: [
// ...
].filter((t) => t.allowed())
```
This would let our "post combat" engine be lighter for people with fewer resources.

pstalcup (changes requested):

packages/garbo/src/tasks/post/post.ts
horrible-little-slime: Done. Got to do some neat horrible type bullshit, too

horrible-little-slime:
This needs to do one last thing: we can't use the `limit` parameter on free fights if we're running the post combat quest.

pstalcup (approved):

Issue #218, opened 2021-07-09
[resource] zatara clan consults


docrostov:
it is perhaps worth using cheesefax's consult engine to get your 3 consults a day for a chance at a skillbook, which is an extra 500k to 1 mil meat on the rare event it procs. 

one callout here is that this code can be a little finicky. in my personal farming script there were 1 or 2 days where the script just stopped because it was waiting for the response. worth including something that has the script keep going if it fails here, because this is super minor stuff.

PR #1641, opened 2023-10-13
Add Clan Fortune Consults task

MrFizzyBubbs:
Addresses #218

### DIFF
--- a/packages/garbo/src/tasks/daily.ts
+++ b/packages/garbo/src/tasks/daily.ts
@@ -14,6 +14,7 @@ import {
   handlingChoice,
   holiday,
   inebrietyLimit,
+  isOnline,
   Item,
   itemAmount,
   mallPrice,
@@ -33,6 +34,7 @@ import {
   use,
   visitUrl,
   votingBoothInitiatives,
+  wait,
 } from "kolmafia";
 import {
   $effect,
@@ -43,6 +45,7 @@ import {
   $monster,
   $slot,
   BeachComb,
+  Clan,
   findLeprechaunMultiplier,
   get,
   getModifier,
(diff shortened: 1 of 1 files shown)


packages/garbo/src/tasks/daily.ts
```diff
@@ -611,6 +614,25 @@ const DailyTasks: GarboTask[] = [
...
+      have($item`Clan VIP Lounge key`) &&
+      getClanLounge()["Clan Carnival Game"] !== undefined &&
+      isOnline("CheeseFax") &&
+      Clan.getWhitelisted().find(
+        (c) => c.name === "Bonus Adventures from Hell",
+      ) !== undefined,
+    completed: () => get("_clanFortuneConsultUses") >= 3,
+    do: (): void => {
+      Clan.with("Bonus Adventures from Hell", () =>
+        cliExecute("fortune 3038166"),
```
horrible-little-slime, line 629: either do like ```
cliExecute(`fortune ${getPlayerId("cheesefax")}`)
```
or just add a comment saying it's cheesefax.

Also, isn't cheesefax dead?

MrFizzyBubbs: I thought its death was only temporary but who knows.

horrible-little-slime (approved):
LGTM

atbell1:
This uses Bonus Adventures from Hell if whitelisted, should it not use the pref for VIP clan instead of this hardcoded value? Will cause issues if whitelisted to BAFH but unable to leave current clan as leader, even if current clan is VIP preference

Malibu-Stacey:
> This uses Bonus Adventures from Hell if whitelisted, should it not use the pref for VIP clan instead of this hardcoded value? Will cause issues if whitelisted to BAFH but unable to leave current clan as leader, even if current clan is VIP preference

You can open an issue if you have a problem with the script or propose your own PR if you have a better solution.
This is not appropriate use of comments on a PR that was merged 4 months ago.
Consider this a friendly warning as a first offence. Future warnings will be less friendly (and less warning like).

PR #1643, opened 2023-10-14
Update ci to run madge on all

Rinn:


horrible-little-slime (approved):
running `yarn workspaces foreach -v run madge` locally worked for me, which is about as much review as I'm prepared to give this

Issue #1645, opened 2023-10-14
[BUG] TypeError: Reduce of empty array with no initial value


zweq:
When running garbo on a low-IOTM account, the following happens:

JavaScript exception: TypeError: Reduce of empty array with no initial value (file:/garbo.js#39952)
at file:/garbo.js:39952 (f)
at file:/garbo.js:41652 (freeFights)
at file:/garbo.js:45843
at file:/garbo.js:38055 (withClan)
at file:/garbo.js:38049 (withVIPClan)
at file:/garbo.js:45824
at file:/garbo.js:38034 (withStash)
at file:/garbo.js:45823 (main)

This is during Executing Daily Items/Clear Existing Rufus Quest, evidently when starting to do free fights. I would assume that at least the Neverending Party would provide some.

gausie:
We should get eslint to warn for any reducer without an initial value full stop

horrible-little-slime:
Please read our guidelines for submitting bug reports. We need a session log!

zweq:
Session log:

[zweq_20231014.txt](https://github.com/loathers/garbage-collector/files/12917952/zweq_20231014.txt)

PR #1648, opened 2023-10-15
Enforce initialValue for reducers to avoid unexpected behaviour

gausie:
Not sure what actually happened to the user in #1645 but this is good practice anyway. If we want to flag selecting no familiar as an error we should do so explicitly, not via a TypeError.

- Add eslint rule to enforce initial value for reducers
- Fix resulting linting errors

### DIFF
--- a/packages/eslint-config-garbo/index.js
+++ b/packages/eslint-config-garbo/index.js
@@ -31,6 +31,14 @@ module.exports = {
 
     // eslint-plugin-libram
     "libram/verify-constants": "error",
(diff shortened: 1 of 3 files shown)


horrible-little-slime (approved):
Approving, with a quick note:
1. This won't apply to reduces that come from external dependencies, like libram and grimoire
2. libram's `maxBy` is set up to not return nullish values, because I found that was way too annoying
3. It _does_ have an informative error message when it specifically is the source of a crash, though, so I think that's fine enough.

PR #1650, opened 2023-10-15
Prevent faxbox abort if network is down

Rinn:
Try and get a photocopied monster before embezzlers are started rather than during execution
Use faxbot rather than messaging cheesefax directly (this is the same logic as autoscend)

### DIFF
--- a/packages/garbo/src/resources/fax.ts
+++ b/packages/garbo/src/resources/fax.ts
@@ -23,14 +23,12 @@ export function faxMonster(monster: Monster): boolean {
   if (getClanLounge()["deluxe fax machine"] === undefined) return false;
   if (get("_photocopyUsed")) return false;
 
-  if (!get("_photocopyUsed")) {
+  if (checkFax(monster)) return true;
+  if (!canFaxbot(monster)) return false;
+  faxbot(monster);
+  for (let i = 0; i < 3; i++) {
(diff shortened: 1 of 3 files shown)


packages/garbo/src/resources/fax.ts
```diff
@@ -0,0 +1,34 @@
...
+  getClanLounge,
+  Monster,
+  wait,
+} from "kolmafia";
+import { $item, get, have, property } from "libram";
+
+function checkFax(monster: Monster): boolean {
+  if (!have($item`photocopied monster`)) cliExecute("fax receive");
+  if (!have($item`photocopied monster`)) return false;
+  if (property.get("photocopyMonster") === monster) {
```
horrible-little-slime, line 14: Is it worth adding an `examine(photocopied monster)` here?

horrible-little-slime: Worth noting: Even if mafia always does it as part of the command, it may not examine the monster on login and so if you start your session with a fax it may still break

horrible-little-slime (approved):
Approved, so long as the answer to my question in the comments is "yeah man it's fine"

Rinn:
> Approved, so long as the answer to my question in the comments is "yeah man it's fine"

yeah man it's fine

PR #1651, opened 2023-10-15
MPA Session Calculation

pstalcup:


horrible-little-slime (dismissed):

packages/garbo/src/outfit/barf.ts
```diff
@@ -160,5 +164,5 @@ export function barfOutfit(spec: OutfitSpec = {}, sim = false): Outfit {
     parka: "kachungasaur",
   });
 
-  return outfit;
+  return { outfit, extraValue };
```
horrible-little-slime, line 167: I don't love doing this for barf and nothing else, seems counterintuitive & inconsistent. But I get it

Might be worth exporting two functions, like the `freeFightFamiliarData` and `freeFightFamiliar` thing?

packages/garbo/src/familiar/barfFamiliar.ts
```diff
@@ -235,7 +245,14 @@ export function barfFamiliar(): Familiar {
...
   );
 
-  return best.familiar;
+  const jellyfish = fullMenu.find(
+    ({ familiar }) => familiar === $familiar`Space Jellyfish`,
+  );
+
+  return {
+    familiar: best.familiar,
+    extraValue: extraValue(best, meatFamiliarEntry, jellyfish),
```
horrible-little-slime, line 254: I think I need some documentation about what `extraValue` means here. extra compared to what?

horrible-little-slime (commented):
Okay I think I see what's going on here; you're letting the `extraValue` be a marginal familiar mpa thing, and you're trickling it down through barfFamiliar and barfOutfit, and then extracting it in session calcs.

Neat!

horrible-little-slime (approved):
I don't love the way we're passing this information around but I can't think of a nicer way; fix the merge conflicts and make it so

PR #1662, opened 2023-10-18
Add free candy support

Ignose:
Separated into it's own file as requested; similar to scepter.


packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,124 @@
...
+    const outfitCandyValue = treatValue(getTreatOutfit());
+    const totOutfitCandyMultiplier = have($familiar`Trick-or-Treating Tot`) ? 1.6 : 1;
+    const bowlValue = (1 / 5) * getSaleValue($item`huge bowl of candy`);
+    const prunetsValue = have($familiar`Trick-or-Treating Tot`)
+      ? 4 * 0.2 * getSaleValue($item`Prunets`)
+      : 0;
+
+    const outfitCandyTotal = 3 * outfitCandyValue * totOutfitCandyMultiplier;
+    let _baseAdventureValue = (1 / 5) * (outfitCandyTotal + bowlValue + prunetsValue);
+    return _baseAdventureValue;
```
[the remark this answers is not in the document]
horrible-little-slime: That being said, it's also worth noting that we get 5 lit houses out of this, so I'm not 100% sure it makes sense to apply that 1/5 multiplier?

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (!outfit.equip(piece))
+      print(`Could not equip all pieces of treat outfit: aborted on ${piece}`);
+  }
+
+  outfit.equip($item`lucky Crimbo tiki necklace`);
+  outfit.equip($familiar`Trick-or-Treating Tot`);
+  return outfit;
+}
+
+export function freeCandyValue(): number {
```
horrible-little-slime, line 69:
```suggestion
export function candyRickBlockValue(): number {
```

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,124 @@
...
+  else
+    return false;
+}
+
+function useCandyMapTask(): GarboTask {
+  return {
+  name: "Acquire Candy Map",
+  completed: () => shouldAcquireCandyMap(),
+  do: (): void => {
+    AcquireItem($item`map to a candy-rich block`)
```
horrible-little-slime, line 94: I'm not sure this does what you think it does. Also, I would make this task both acquire _and_ use the map

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,124 @@
...
+  },
+};
+}
+
+function doCandyTreat(): GarboTask {
+  return {
+    name: "Treat",
+    ready: () => !treated,
```
horrible-little-slime, line 102: we can delete this in favor of using a `limit: { skip: 1 }`, which happens for free if we're running this in a `SafeGarboEngine`, which we do use in `dailyItems`

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,124 @@
...
+    completed: () => !have($item`map to a candy-rich block`) || get("_mapToACandyRichBlockUsed") 
+    || !getBlockHtml().match(/whichhouse=\d*>[^>]*?house_l/),
```
horrible-little-slime, line 104: remove the item check, because we're now using the item as part of the other task; move the pref check to the `ready` condition. I think we can eschew all of the specialized getBlockHtml() architecture because we aren't going to more than one block.

horrible-little-slime (changes requested):

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,124 @@
...
+
+
+    if (!availableOutfits.length) {
+      print("You don't seem to actually have any outfits available, my friend!");
+    }
+
+  }
+
+  return get("freecandy_treatOutfit");
+}
```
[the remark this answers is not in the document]
Ignose: How would I go about doing this?

getOutfits() will tell me all the outfits I have; but I'm not sure how to then append the candy value of each outfit and pick the best one.

horrible-little-slime, line 54:
```suggestion
export function getTreatOutfit(): string {
  const availableOutfits = getOutfits().filter((name) =>
    outfitPieces(name).every((piece) => canEquip(piece))
  );

  if (!availableOutfits.length) {
    print("You don't seem to actually have any outfits available, my friend!");
  }
  return maxBy(availableOutfits, treatValue);
}
```

packages/garbo/src/resources/candyMap.ts
Ignose: How will we determine we've hit every house in the block?

packages/garbo/src/resources/candyMap.ts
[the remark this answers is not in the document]
Ignose: I tried a thing.

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,123 @@
...
+    relativeValue: treatValue(outfit),
+}));
+
+const outfitWithHighestValue = outfitsWithValues.reduce((prev, current) => {
+  return prev.relativeValue > current.relativeValue ? prev : current;
+});
+
+  return outfitWithHighestValue.outfit;
+ //ChatGPT made this, I have no idea if it works!
+}
```
horrible-little-slime, line 55:
```suggestion
export function getTreatOutfit(): string {
  const availableOutfits = getOutfits().filter((name) =>
    outfitPieces(name).every((piece) => canEquip(piece))
  );
  if (!availableOutfits.length) {
    print("You don't seem to actually have any outfits available, my friend!");
  }
  return maxBy(availableOutfits, treatValue);
}
```

horrible-little-slime: `maxBy` is more or less equivalent to this reduce, but a little cleaner to read here. We need to check the array length because reducing an empty array can be dangerous. We need to filter the result of `getOutfits` to only include outfits we can equip

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,124 @@
...
+
+function getBlockHtml(): string {
+  blockHtml ||= visitUrl("place.php?whichplace=town&action=town_trickortreat");
+  return blockHtml;
+}
+
+function treatValue(outfit: string): number {
+  return sum(
+    Object.entries(outfitTreats(outfit)),
+    ([candyName, probability]) => probability * getSaleValue(toItem(candyName))
```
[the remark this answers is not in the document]
Ignose: Done

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,124 @@
...
+  if(freeCandyValue() < getSaleValue($item`map to a candy-rich block`))
+    return true;
+  else
+    return false;
+}
+
+function useCandyMapTask(): GarboTask {
+  return {
+  name: "Acquire Candy Map",
+  completed: () => shouldAcquireCandyMap(),
```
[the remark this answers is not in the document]
Ignose: Done!

packages/garbo/src/resources/candyMap.ts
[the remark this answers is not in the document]
Ignose: This should also be done; I'd welcome taking a look.

packages/garbo/src/resources/candyMap.ts
Ignose: Updated as requested. ready is... new to me but I'm understanding it better!

packages/garbo/src/resources/candyMap.ts
Ignose: Done!

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export function treatOutfit(): Outfit {
+  const outfit = new Outfit();
+  const pieces = outfitPieces(getTreatOutfit());
+  for (const piece of pieces) {
+    if (!outfit.equip(piece))
+      print(`Could not equip all pieces of treat outfit: aborted on ${piece}`);
+  }
+
+  outfit.equip($item`lucky Crimbo tiki necklace`);
```
horrible-little-slime, line 60: We as a society have decided that this does nothing. I'll remove it from freecandy if relevant

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    const outfitCandyTotal = 3 * outfitCandyValue * totOutfitCandyMultiplier;
+    return (outfitCandyTotal + bowlValue + prunetsValue);
+}
+
+function shouldAcquireCandyMap(): boolean {
+  if(candyRichBlockValue() < mallPrice($item`map to a candy-rich block`))
+    return true;
+  else
+    return false;
```
horrible-little-slime, line 81:
```suggestion
  return candyRichBlockValue() < mallPrice($item`map to a candy-rich block`)
```

packages/garbo/src/resources/candyMap.ts
horrible-little-slime: also, AcquireItem is an object type used for the `acquire` field on grimoire tasks, not a function. we should instead use garbo's `acquire` function, imported from `../acquire`

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,117 @@
...
+    completed: () => get("_mapToACandyRichBlockUsed"),
+    outfit: treatOutfit,
+    do: (): void => {
+      // We do all treat houses in a row as one task for speed reasons
+      for (const house of HOUSE_NUMBERS) {
+        if (getBlockHtml().match(RegExp(`whichhouse=${house}>[^>]*?house_l`))) {
+          visitUrl(`choice.php?whichchoice=804&option=3&whichhouse=${house}&pwd`);
+        } else if (getBlockHtml().match(RegExp(`whichhouse=${house}>[^>]*?starhouse`))) {
+          visitUrl(`choice.php?whichchoice=804&option=3&whichhouse=${house}&pwd`);
+          runChoice(2);
```
horrible-little-slime, line 108: We need another visitUrl to `place.php?whichplace=town&action=town_trickortreat` after the `runChoice`, I assume that this has to do with the fact that the star house is an actual choice adventure inside the trick-or-treating choice (while the other light houses are non-choice noncombats)

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+      }
+    },
+  };
+}
+
+export function freeCandyTasks(): GarboTask[] {
+  return useCandyMapTask() + doCandyTreat();
```
horrible-little-slime, line 116:
```suggestion
  return [useCandyMapTask(), doCandyTreat()];
```

horrible-little-slime (changes requested):
There are a few syntactical errors here that I'm surprised your IDE didn't complain about--are you using vscode?

packages/garbo/src/resources/candyMap.ts
Ignose: Removed, then.

packages/garbo/src/resources/candyMap.ts
Ignose: If we're going to do acquire, I'm going to do it a bit differently. If expected profit of 1 meat is too little meat, we can change that

yarn.lock
```diff
@@ -4309,7 +4309,7 @@ __metadata:
     garbo-lib: ^0.0.1
     grimoire-kolmafia: ^0.3.23
     kolmafia: ^5.27628.0
-    libram: ^0.8.13
+    libram: latest
```
s-k-z, line 4312: If you need to bump the version then it should specify the version number. There are now two versions in this lock file due to the way it was added.

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,120 @@
...
+      use($item`map to a candy-rich block`);
+    },
+    spendsTurn: false,
+  };
+}
+
+function doCandyTreat(): GarboTask {
+  return {
+    name: "Treat",
+    completed: () => get("_mapToACandyRichBlockUsed"),
```
horrible-little-slime, line 93: t
```suggestion
    completed: () => !get("_mapToACandyRichBlockUsed") || holiday().includes("Halloween"),
```

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,120 @@
...
+  const prunetsValue = have($familiar`Trick-or-Treating Tot`)
+    ? 4 * 0.2 * getSaleValue($item`Prunets`)
+    : 0;
+
+  const outfitCandyTotal = 3 * outfitCandyValue * totOutfitCandyMultiplier;
+  return outfitCandyTotal + bowlValue + prunetsValue;
+}
+
+function shouldAcquireCandyMap(): boolean {
+  return candyRichBlockValue() < mallPrice($item`map to a candy-rich block`);
```
horrible-little-slime, line 74:
```suggestion
  return !holiday().includes("Halloween") && (candyRichBlockValue() < mallPrice($item`map to a candy-rich block`));
```

horrible-little-slime (changes requested):

packages/garbo/src/resources/candyMap.ts
horrible-little-slime: we're relying on the `limit: { skip: 1 }` from `runSafeGarboTasks` to stop this from repeating, which was admittedly my idea but also doesn't feel great. so I'll ruminate on that

yarn.lock
Ignose: Fixed!

packages/garbo/package.json
```diff
@@ -28,7 +28,7 @@
     "garbo-lib": "^0.0.1",
     "grimoire-kolmafia": "^0.3.23",
     "kolmafia": "^5.27628.0",
-    "libram": "^0.8.13"
+    "libram": "latest"
```
s-k-z, line 31:
```suggestion
    "libram": "^0.8.14"
```

yarn.lock
```diff
@@ -4928,6 +4928,17 @@ __metadata:
...
+  version: 0.8.14
+  resolution: "libram@npm:0.8.14"
+  dependencies:
+    html-entities: ^2.4.0
+  peerDependencies:
+    kolmafia: ^5.26781.0
+  checksum: 342672a9a5f2b798c3493006b79c3cd9224a7cb4425960e218bc09244b7a48b4656b6535093cdbd37e93df88bbc4d6af3edb5395bc3a5fe624ef19714f50810a
+  languageName: node
+  linkType: hard
+
```
s-k-z, line 4941: The libram package directly above this should be updated and this one should be removed.

s-k-z (changes requested):

yarn.lock
Ignose: Also fixed!

s-k-z: It looks like the yarn.lock file is no longer modified. While it's good at there aren't duplicate packages, you probably want 0.8.14 to be the version specified not 0.8.13.

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,123 @@
...
+    ([candyName, probability]) => probability * garboValue(toItem(candyName)),
+  );
+}
+
+export function getTreatOutfit(): string {
+  const availableOutfits = getOutfits().filter((name) =>
+    outfitPieces(name).every((piece) => canEquip(piece)),
+  );
+  if (!availableOutfits.length) {
+    print("You don't seem to actually have any outfits available, my friend!");
```
s-k-z, line 42:
```suggestion
    print("You don't seem to actually have any outfits available, my friend!");
```
This message should mention trick-or-treat, otherwise users won't know what it refers to.

yarn.lock
Ignose: What's the best way to correct this without yarn upgrade kolmafia@latest overriding to latest again?

s-k-z:
```
yarn add libram@0.8.14
yarn dedupe libram
```
Should be sufficient here. If it also needs to be upgraded in garbo-lib or another workspace then you can do:
```
yarn workspace garbo-lib add --peer libram@0.8.14
```

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,123 @@
...
+        ) {
+          visitUrl(
+            `choice.php?whichchoice=804&option=3&whichhouse=${house}&pwd`,
+          );
+          runChoice(2);
+          visitUrl(`place.php?whichplace=town&action=town_trickortreat`);
+        }
+      }
+    },
+    limit: { tries: 1 },
```
horrible-little-slime, line 115:
```suggestion
    limit: { skip: 1 },
```

s-k-z:
It may not be required to update garbo-lib to libram 0.8.13, but why not keep both in sync where possible?

packages/garbo/src/resources/candyMap.ts
```diff
@@ -0,0 +1,140 @@
...
+    ? 4 * 0.2 * getSaleValue($item`Prunets`)
+    : 0;
+
+  const outfitCandyTotal = 3 * outfitCandyValue * totOutfitCandyMultiplier;
+  return outfitCandyTotal + bowlValue + prunetsValue;
+}
+
+function shouldAcquireCandyMap(): boolean {
+  return (
+    !holiday().includes("Halloween") &&
```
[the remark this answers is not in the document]
horrible-little-slime: I'm very down for a sanity check, but I think it should be pretty big.

horrible-little-slime (approved):

horrible-little-slime (approved):

PR #1667, opened 2023-10-21
Ween

pstalcup:
todo:
Don't meat buff
Don't use embezzler outfit (BonusEquipMode.EMBEZZLER)
Don't calculate profit based on embezzler for copies when there is untapped sources
Diet changes
rename a bunch of variables \ functions and comments
what if the target isn't free?


Rinn:
todo:
Don't meat buff
Don't use embezzler outfit (BonusEquipMode.EMBEZZLER)
Diet changes
rename a bunch of variables \ functions and comments

packages/garbo/src/fights.ts
```diff
@@ -232,7 +231,7 @@ function embezzlerSetup() {
   setLocation($location`Friar Ceremony Location`);
```
gausie, line 231: Should this function be renamed?

packages/garbo/src/fights.ts
```diff
@@ -232,7 +231,7 @@ function embezzlerSetup() {
...
   potionSetup(false);
   maximize("MP", false);
-  meatMood(true, 750 + baseMeat).execute(embezzlerCount());
+  meatMood(true, 750 + baseMeat).execute(copyTargetCount());
```
gausie, line 234: Do we still want a meatMood here?

packages/garbo/src/fights.ts
```diff
@@ -255,7 +254,7 @@ function embezzlerSetup() {
     }
   }
 
-  bathroomFinance(embezzlerCount());
+  bathroomFinance(copyTargetCount());
```
gausie, line 257: Do we want to greenspan if not a 'bezzler

packages/garbo/src/garboWanderer.ts
```diff
@@ -18,13 +18,13 @@ export function wanderer(): WandererManager {
...
       effectValue: (effect: Effect, duration: number) =>
-        new Potion($item.none, { effect, duration }).gross(embezzlerCount()),
+        new Potion($item.none, { effect, duration }).gross(copyTargetCount()),
       prioritizeCappingGuzzlr: get("garbo_prioritizeCappingGuzzlr", false),
       freeFightExtraValue: (location: Location) =>
         freeFightFamiliarData({ location }).expectedValue,
       digitzesRemaining: digitizedMonstersRemainingForTurns,
       plentifulMonsters: [
-        $monster`Knob Goblin Embezzler`,
+        globalOptions.target,
```
gausie, line 27: Will this behave weird if sausage goblin is selected given the following lines

packages/garbo/src/outfit/lib.ts
```diff
@@ -94,7 +94,7 @@ export function useUPCsIfNeeded({ familiar }: Outfit): void {
   const currentWeapon =
     25 * (familiar ? findLeprechaunMultiplier(familiar) : 0);
   const embezzlers = globalOptions.ascend
```
gausie, line 96: Rename var?

horrible-little-slime:
We will have a lot more trouble using orb during gregs if the copy target is free.

One general concern I have with this is that I don't want it to become a lot harder to add more embezzler-related stuff, our bread and butter, because of this. We should be careful as we write this out to not make that a tougher area to work on

packages/garbo/src/embezzler/fights.ts
```diff
@@ -411,29 +419,29 @@ export const copySources = [
...
       withMacro(options.macro, () => use($item`envyfish egg`)), options.useAuto;
     },
   ),
-  new EmbezzlerFight(
+  new CopyTargetFight(
     "Screencapped Monster",
     () =>
       have($item`screencapped monster`) &&
-      property.getString("screencappedMonster") === "Knob Goblin Embezzler",
+      property.get("screencappedMonster") === globalOptions.target,
```
horrible-little-slime, line 442: most of the `getString`s are prefs that get set to weird, inappropriate values when not set (e.g., olfaction used to get set to "wabbit"). If you're switching them to `get`s, be certain that it's safe to do so, and also just import `get` directly instead of doing `propert.get`

gausie: Is it? Ugh. We should maybe normalise that on the libram side.

horrible-little-slime: I cannot confirm that this is still the case, but it was the case when this was written.

horrible-little-slime:
> > We will have a lot more trouble using orb during gregs if the copy target is free.
> > One general concern I have with this is that I don't want it to become a lot harder to add more embezzler-related stuff, our bread and butter, because of this. We should be careful as we write this out to not make that a tougher area to work on
> 
> I think the best option is to just not use orb if copier target is free. It can be done if you have two zones banished (as someone who used to do witchess knights back when they were more profitable, the best zones I found for this was to go between toxic teacups and warren, equipping antler in warren to banish) - This way you reset orb by actually encountering a predicted monster by having orb equipped the entire time. You can also do noob cave, but we also want a zone like that for replacers
> 
>     1. First adventure in toxic teacups (after already banishing) predicts greg monster
> 
>     2. Fight predicted greg monster (Any previous prediction is reset - in this case our bad prediction from warren)
> 
>     3. Fight in warren (final greg fight - predicts target monster)
> 
>     4. Fight orb prediction (predicts bunny)
> 
> 
> This is for extros with only 3 gregs though

Yeah, given that crates will be sniffed we would need another zone, and ideally one that everyone always has access to. Which means that I too fall on the side of disabling it.

gausie:
I think this should be merged to main without being released as a feature, so we don't have to deal with ongoing merge conflict resolution. I ran it today with no issues.

Future PRs to get this feature itself over the line are encouraged.

horrible-little-slime:
Go for it

PR #1671, opened 2023-10-22
Use lockpicks in DD

ktpolanski:
```
Took choice 692/3: pick lock with lockpicks, no turn spent
```


ktpolanski:
Wait why is this two commits in here

PR #1679, opened 2023-10-22
free goofballs

Rinn:


horrible-little-slime:
Does this work in CS aftercore? is there a more precise check we can apply here than "is the woods unlocked"?

s-k-z:
I think checking if `questL03Rat` is not unstarted is a decent way to check. That should guarantee you have access to the tavern.

Rinn:
I'll check today

Rinn:
Yeah you're right, no tavern access in cs aftercore this needs to check questL03Rat

PR #1680, opened 2023-10-22
Acquire sea jelly

Rinn:
Unlock the sea and get sea jelly daily with space jellyfish


horrible-little-slime:
This is part of breakfast; is there a reason to add this given that?

Malibu-Stacey:
> This is part of breakfast; is there a reason to add this given that?

`breakfast` will only do it if you have "opened" the sea by talking to the man in the shack.

Issue #1687, opened 2023-10-27
[BUG] non-barf turn delay doesn't account for reserved adventures


Rinn:
**Describe the bug**
I also believe there's an off-by-one error with this count, so the last non-barf turn is not used. The issue is here:
```
export const NonBarfTurnQuest: Quest<GarboTask> = {
  name: "Non Barf Turn",
  tasks: NonBarfTurnTasks,
  completed: () =>
    !canContinue() ||
    clamp(myAdventures() - digitizedMonstersRemaining(), 1, myAdventures()) >=
      nonBarfTurns(),
};
```

We've discussed adding `ready` to the grimoire quest to make this logic easier to write, because having it in completed is pretty messy.

Adding a ticket so I remember to look into this and do the grimoire PR.

Rinn:
pstalcup already got to the grimoire change https://github.com/loathers/grimoire/pull/115

PR #1691, opened 2023-10-28
use quest ready for nonbarf turns

Rinn:
#1687 Fixes off by one and accounts for saveTurns when determining if non-barf actions are ready

testing now, will report back in a couple hours


Rinn:
testing worked

PR #1692, opened 2023-10-28
fix: don't assume we can get a quake of arrows

midgleyc:
Having an Obtuse Angel but not a quake of arrows would lead to an abort of "unable to satisfy famspec". Try to work around that.

It costs ~50k in the mall and can rise substantially, so it can breach your auto-buy price limit.

### DIFF
--- a/packages/garbo/src/fights.ts
+++ b/packages/garbo/src/fights.ts
@@ -63,7 +63,6 @@ import {
   $effect,
   $effects,
   $familiar,
-  $familiars,
   $item,
   $items,
   $location,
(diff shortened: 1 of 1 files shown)


Rinn:
maybe it should just prefer reanimator and then only set the quake of arrows as the famequip if you have it, because getting 2 copies is better than none.

packages/garbo/src/fights.ts
```diff
@@ -525,9 +524,17 @@ export function dailyFights(): void {
...
 
-        const bestCopier = $familiars`Obtuse Angel, Reanimated Reanimator`.find(
-          have,
-        );
+        // use obtuse angel if have + have quake of arrows, otherwise reanimator
+        // quake of arrows is PvP-stealable and costs ~50k, so don't assume we have it
+        let bestCopier: Familiar | undefined;
+        if (
+          have($familiar`Obtuse Angel`) &&
+          retrieveItem($item`quake of arrows`)
```
s-k-z, line 532: I don't think this is the best solution because `retrieveItem` may unequip the quake of arrows unnecessarily.

midgleyc: I think this should sort it?

s-k-z: Yeah!

s-k-z (approved):

PR #1697, opened 2023-10-29
rollover cutoff not using grimoirized_post

horrible-little-slime:


horrible-little-slime:
Just realized this works just fine even on grimoirized post (although I'll do some stuff to make it better)

packages/garbo/src/lib.ts
```diff
@@ -170,6 +171,12 @@ export function expectedEmbezzlerProfit(): number {
 }
 
 export function safeInterrupt(): void {
+  if (
+    globalOptions.prefs.rolloverBuffer * 60 * 1000 <
+    rollover() * 1000 - Date.now()
+  ) {
+    set("garbo_interrupt", true);
```
Rinn, line 178: this is real minor, but could you just throw an error saying the reason for the abort instead of setting garbo_interrupt because the garbo_interrupt message isn't technically correct

Rinn (changes requested):

horrible-little-slime:
Yeah I have to fiddle with it after the grimoirization of post anyways

Rinn (approved):

PR #1714, opened 2023-11-05
Automatically gather resin

pstalcup:


horrible-little-slime:
New libram is lit; update to support that or I will at EOD. Then we're in, baby!

PR #1728, opened 2023-11-11
Value leaf rakes

Rinn:


horrible-little-slime (approved):
I think technically we have no reason to disable this for embezzlers, because there's no decay or limit to it, but given that we think the price is inflated rn we should do it nevertheless

PR #1736, opened 2023-11-13
Burn leaves for 3x lit leaf lasso & day shortener

Rinn:
Do we want to reserve any leaves or just do this every day? Or maybe move to barfTurn so it will buy them when leaves are acquired?

### DIFF
--- a/packages/garbo/src/tasks/dailyItems.ts
+++ b/packages/garbo/src/tasks/dailyItems.ts
@@ -7,7 +7,6 @@ import {
   getClanLounge,
   getMonsters,
   Item,
-  itemAmount,
   itemDropsArray,
   itemPockets,
   mallPrice,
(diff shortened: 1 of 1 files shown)


packages/garbo/src/tasks/dailyItems.ts
```diff
@@ -408,6 +409,35 @@ const DailyItemTasks: GarboTask[] = [
...
+    name: "Burning Leaves day shortener",
+    ready: () =>
+      BurningLeaves.have() &&
+      itemAmount($item`inflammable leaf`) >=
+        BurningLeaves.burnFor.get($item`day shortener`)!,
+    completed: () => get("_leafDayShortenerCrafted"),
+    do: () => {
+      BurningLeaves.burnLeaves(
+        BurningLeaves.burnFor.get($item`day shortener`)!,
+      );
```
horrible-little-slime, line 437: I think `BurningLeaves` has a function where you can directly pass the item in, and avoid non-null assertions

Malibu-Stacey (approved):

Issue #1699, opened 2023-10-29
Not using Fire Extinguisher charges or replica batoomerang 


Fawoosh:
I am finding that these two resources are being unused in my second leg of garbo after CS.

I have also noticed that my extinguisher charges do not get used in leg 1 either.

Fawoosh:
[Butts_McGruff_20231029.txt](https://github.com/loathers/garbage-collector/files/13198997/Butts_McGruff_20231029.txt)

Tapionator:
These work for me but the only differing factor between myself and Fawoosh is that I have Stomping Boots and he does not. It may be an issue with the "// Fire Extinguisher on best available target" section of Garbo.

Fawoosh:
[Butts_McGruff_20231112.txt](https://github.com/loathers/garbage-collector/files/13328139/Butts_McGruff_20231112.txt)

Fawoosh:
Here is another log. Just ran my leg 1 and it did not use fire extinguisher whatsoever. Ended the day with 100% charge left.

PR #1738, opened 2023-11-15
move extinguisher above jellyfish & gingerbread

horrible-little-slime:
I've found my extinguisher isn't getting used, because I burn all my non-familiar free runs on the jellyfish, and because it requests that I use the xo skeleton.

Long-term, we should streamline this logic when we grimoirize, make it less terrible. Short-term, I feel like this is good. Gingerbread as a wanderer target means that it's less important to get in there with free runs, and imo this is more valuable than marginal jellyfish-turns (less valuable than your first jellyfish turn, but we do take care of that during barf at EOD)

### DIFF
--- a/packages/garbo/src/fights.ts
+++ b/packages/garbo/src/fights.ts
@@ -1383,6 +1383,76 @@ const freeRunFightSources = [
     },
     freeRunConstraints(true),
   ),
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
Resolves #1699

Rinn (approved):

PR #1745, opened 2023-11-20
Bofa wish fights

Shiverwarp:


packages/garbo/src/fights.ts
```diff
@@ -179,6 +182,11 @@ import { FreeFightQuest, runGarboQuests } from "./tasks";
 import { expectedFreeFights, possibleTentacleFights } from "./tasks/freeFight";
 import { bestMidnightAvailable } from "./resources";
 import { PostQuest } from "./tasks/post";
+import {
+  canAdventureOrUnlock,
+  canWander,
+  unlock,
+} from "garbo-lib/dist/wanderer/lib";
```
horrible-little-slime, line 189: This isn't how we should accomplish this; we should either use a new wanderer type or an existing wanderer type, or export those functions from garbo-lib through `./wanderer`

packages/garbo/src/fights.ts
```diff
@@ -2667,3 +2720,53 @@ function runShadowRiftTurn(): void {
...
+      !badAttributes.some((s) => m.attributes.includes(s)) && rates[m.name] > 0,
+  );
+
+  if (monsters.length === 0) {
+    return 0;
+  } else {
+    return (
+      sum(monsters, (m) => {
+        return itemFact(m) === $item`pocket wish` ? 1 : 0;
+      }) / monsters.length
```
horrible-little-slime, line 2738:
```suggestion
     monsters.filter((m) => itemFact(m) === $item`pocket wish`).length / monsters.length
```

packages/garbo/src/fights.ts
```diff
@@ -1383,6 +1391,51 @@ const freeRunFightSources = [
...
+      );
+    },
+    {
+      spec: () => {
+        const freeKill = findFreeKill();
+        if (!freeKill) {
+          throw `We should have a free kill for bofa wishes, but don't`;
+        }
+        const spec: OutfitSpec = freeKill.spec ?? {};
+        return spec;
```
horrible-little-slime, line 1435:
```suggestion
        return freeKill.spec ?? {};
```

packages/garbo/src/fights.ts
```diff
@@ -1383,6 +1391,51 @@ const freeRunFightSources = [
...
+      const freeKillMacro =
+        freeKill.macro instanceof Item
+          ? Macro.item(freeKill.macro)
+          : Macro.skill(freeKill.macro);
+      garboAdventure(
+        best,
+        Macro.if_(
+          wishMonsters.map((m) => `!monsterid ${m.id}`).join(" && "),
+          runSource.macro,
+        ).step(freeKillMacro),
```
horrible-little-slime, line 1425:
```suggestion
        Macro.if_(
          wishMonsters,
         freeKillMacro,
        ).step(runSource.macro),
```
just seems a little more straightforward

packages/garbo/src/fights.ts
```diff
@@ -1383,6 +1391,51 @@ const freeRunFightSources = [
     },
     freeRunConstraints(true),
   ),
+  // Use free runs and free kills to get our bofa wishes
+  new FreeRunFight(
+    () =>
+      have($skill`Just the Facts`) &&
+      get("_bookOfFactsWishes") < 3 &&
```
horrible-little-slime, line 1398: Is there a reason not to roll these checks into `shouldDoBofaFishing`?

packages/garbo/src/fights.ts
```diff
@@ -1383,6 +1391,51 @@ const freeRunFightSources = [
...
+      shouldDoBofaFishing() &&
+      findFreeKill() !== null &&
+      getBestBofaWishLocation() !== null &&
+      unlock(
+        getBestBofaWishLocation(),
+        (3 - get("_bookOfFactsWishes")) * garboValue($item`pocket wish`),
```
horrible-little-slime, line 1404: probably clamp this

packages/garbo/src/fights.ts
```diff
@@ -1383,6 +1391,51 @@ const freeRunFightSources = [
...
+      ),
+    (runSource: ActionSource) => {
+      const best = getBestBofaWishLocation();
+      if (!best) throw `Bofa wish location should exist, but doesn't`;
```
horrible-little-slime, line 1408:
```suggestion
      if (!best) throw new Error("Bofa wish location should exist, but doesn't");
```
We generally prefer Errors to strings when throwing, and we should only use backticks when we're doing template literal stuff

horrible-little-slime (changes requested):
A few quibbles, but mostly I'm not sure if this is actually better than just using our existing code for freekills like Free For All; feels like we should kinda just do that. Which I do assume for most players would get bofa stuff; I believe the wanderer type there is `"freefight"` but I'm not 100% sure offhand.

If you _do_ do that, be sure to pass the spec from the freekill into the wanderer function

Issue #1702, opened 2023-10-31
[BUG] looping after free fight when trainset cannot be reconfigured (too soon)


BadHorseMonkey:
**Describe the bug**
When running latest (with the grimorized post combat), a free fight causes an infinite loop, possibly failing to update my trainset

**To Reproduce**
In aftercore, I ran garbo ascend (and later garbo).  Following a free fight, the code got into an endless loop.

gCLI log:
Visit to The Red Zeppelin's Mooring: The Red Zeppelin in progress...
Preference lastAdventure changed from Barf Mountain to The Red Zeppelin

[718] The Red Zeppelin
Preference lastEncounter changed from garbage tourist to red butler
Encounter: red butler
Preference _lastCombatStarted changed from 20231031060500 to 20231031061107
Round 1: Your toy train moves ahead to the Meat Mine Sluice. The sluice opens and pours out some meat!
Preference trainsetPosition changed from 668 to 669
You gain 801 Meat.
Round 1: You lose 10 hit points
Round 2: Mad Carew wins the fight!
You acquire an item: lemon (4)
After Battle: Something falls out of your can of mixed everything.
You acquire an item: New Zealand iced tea
You gain 661 Meat
You gain 14 Meat.
After Battle: Nearly all credible zoologists agree that the red butler 's guts are so hot, even just imagining cutting one open and crawling inside will keep you warm for hours.
You acquire an effect: Imagining Guts (10)
You acquire an item: glark cable
After Battle: You gain 40 Fortitude
After Battle: You gain 21 Wizardliness
After Battle: You gain 33 Chutzpah
Preference _glarkCableUses changed from 4 to 5
Preference lastCopyableMonster changed from garbage tourist to red butler
This combat did not cost a turn

Preference _concoctionDatabaseRefreshes changed from 546 to 547
logPreferenceChangeFilter => _lastCombatStarted,auto_maximize_current,garboEmbezzlerCount,garboEmbezzlerDate,garboEmbezzlerSources,libram_savedMacro,maximizerMRUList,spadingData,testudinalTeachings
[...]

PR #1747, opened 2023-11-22
A partial fix for #1702

libraryaddict:
I don't believe this fixes the underlying issue, but it should prevent garbo encountering a non-stop loop.
What it does is if the trainset was not configured, but is still configurable, set the last trainset configured counter to 39 trainset positions ago.
This should then tell garbo that we can configure the trainset in 1 train station instead of now.


horrible-little-slime:
I don't love this solution but I accept it.

PR #1748, opened 2023-11-22
Burning Leaf Implementation

Shiverwarp:


packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -327,6 +329,46 @@ const NonBarfTurnTasks: AlternateTask[] = [
...
+        mallPrice($item`day shortener`) + 5 * get("valueOfAdventure"),
+    completed: () => get(`_garboDayShortenersUsed`, 0) >= 3, // Arbitrary cap at 3, since using 3 results in only 1 adventure
+    do: () => {
+      acquire(
+        1,
+        $item`day shortener`,
+        garboValue($item`extra time`) - 5 * get("valueOfAdventure"),
+      );
+      use($item`day shortener`);
+      set(`_garboDayShortenersUsed`, get(`_garboDayShortenersUsed`, 0) + 1);
```
horrible-little-slime, line 346: Does this not currently have mafia tracking? We should just add mafia tracking

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -327,6 +329,46 @@ const NonBarfTurnTasks: AlternateTask[] = [
...
+    ready: () =>
+      globalOptions.ascend &&
+      garboValue($item`extra time`) >
+        mallPrice($item`day shortener`) + 5 * get("valueOfAdventure"),
+    completed: () => get(`_garboDayShortenersUsed`, 0) >= 3, // Arbitrary cap at 3, since using 3 results in only 1 adventure
+    do: () => {
+      acquire(
+        1,
+        $item`day shortener`,
+        garboValue($item`extra time`) - 5 * get("valueOfAdventure"),
```
horrible-little-slime, line 343: Do we want to set `throwOnFail` to be false here?

packages/garbo/src/tasks/dailyItems.ts
```diff
@@ -80,6 +80,23 @@ function drawBestCards(): void {
...
+        entry[0] instanceof Item &&
+        entry[0].tradeable &&
+        !ignored.includes(entry[0]),
+    )
+    .map(([item]) => item);
+  const valuePerLeaf = (leafItem: Item): number => {
+    const leafCost = BurningLeaves.burnFor.get(leafItem)!;
+    return garboValue(leafItem) / leafCost;
+  };
+  return maxBy(leafItems, valuePerLeaf);
```
horrible-little-slime, line 97:
```suggestion
  const leafItems = [...BurningLeaves.burnFor]
    .filter(
      (entry): entry is [Item, number] =>
        entry[0] instanceof Item &&
        entry[0].tradeable &&
        !ignored.includes(entry[0]),
    );
  const valuePerLeaf = ([item, leaves]: [Item, number])=> garboValue(item) / leaves;
  return maxBy(leafItems, valuePerLeaf)[0];
```
We should enable the thing in the linter to yell about non-nullish assertions

packages/garbo/src/tasks/freeFight.ts
```diff
@@ -576,6 +577,18 @@ const FreeFightTasks: GarboFreeFightTask[] = [
       clamp(3 - CombatLoversLocket.reminiscesLeft() - locketsToSave(), 0, 3),
   },
   { ...doCandyTrick(), combatCount: () => 5, tentacle: true },
+  // leaf burning fights
+  {
+    name: "Burning Leaves Flaming Leaflet Fight",
+    ready: () =>
+      BurningLeaves.have() &&
+      BurningLeaves.numberOfLeaves() >=
+        BurningLeaves.burnFor.get($monster`flaming leaflet`)!,
```
horrible-little-slime, line 586:
```suggestion
        BurningLeaves.burnFor.get($monster`flaming leaflet`) ?? 0,
```

horrible-little-slime (changes requested):

packages/garbo/src/diet.ts
```diff
@@ -276,6 +276,16 @@ export function nonOrganAdventures(): void {
   if (globalOptions.ascend) {
     useIfUnused($item`borrowed time`, "_borrowedTimeUsed", 20 * MPA);
   }
+
+  while (get("_extraTimeUsed", 3) < 3) {
```
Rinn, line 280: tbh while loops like this always sketch me out, because if mafia somehow fails to update the property it's gonna infinite loop. can you instead calculate how many extra time should be used without a loop then just use that many?

Rinn (changes requested):

packages/garbo/src/tasks/dailyItems.ts
```diff
@@ -429,6 +442,15 @@ const DailyItemTasks: GarboTask[] = [
     do: () => BurningLeaves.burnSpecialLeaves($item`day shortener`),
     spendsTurn: false,
   },
+  {
+    name: "Burning Leaves Best Unlimited Tradeable Item",
```
Rinn, line 446: i'm not 100% we should be doing this yet

packages/garbo/src/tasks/dailyItems.ts
```diff
@@ -429,6 +442,15 @@ const DailyItemTasks: GarboTask[] = [
...
+    completed: () =>
+      !BurningLeaves.have() ||
+      BurningLeaves.numberOfLeaves() <
+        500 + (BurningLeaves.burnFor.get(bestLeafBurn()) ?? Infinity), // Keep a reserve of 500 leaves for limited items and fights
```
Rinn, line 450: If we do this I think I'd like to see a higher reserve, like enough to get items and fights for several days / a week at least

packages/garbo/src/tasks/dailyItems.ts
Rinn: we don't trade anything else unlimited right now from iotms and i just don't want to burn all of someone's leaves if they're saving up for something. it also wouldn't automatically equalize the mall right away unless everything generated makes it way into the mall and not everyone has philter set up

Rinn (approved):

Issue #1772, opened 2023-12-13
cannot call maxBy on an empty Array


danbeardan:
Ran Garbo after a CS loop and received this error:

JavaScript exception: Error: Cannot call maxBy on an empty array! (file:/E:/KOLMafia/scripts/garbage-collector/garbo.js#25167)

[slik_20231213.txt](https://github.com/loathers/garbage-collector/files/13657469/slik_20231213.txt)

giu-scopely:
I am receiving this as well constantly today, garbo runs 1 turn, then gives this error:
`JavaScript exception: Error: Cannot call maxBy on an empty array! (file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js#25167)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:25167 (f)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:25269 (freefightFactory)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:25536 (bestWander)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:25567 (wanderWhere)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:25573 (wanderWhere)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:25573 (wanderWhere)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:25573 (wanderWhere)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:25731 (getTarget)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:25737 (getChoices)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:41445
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:1107 (undelay)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:29080 (setChoices)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:28924 (execute)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:47531 (execute)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:28857 (run)
at file:/D:/Loathing/Kol/scripts/garbage-collector/garbo.js:47588 (runGarboTasks)
[...]

PR #1773, opened 2023-12-13
Handle empty lists of valid locations

pstalcup:


horrible-little-slime:
Going to merge this, but not close #1772 because it is mysterious as to why that array is empty

PR #1795, opened 2024-01-16
Cane, wardrobe, thesis mood skip

Shiverwarp:
A few small-ish things

- Just adding wardrobe usage to start no matter our level. I figure we're better off using it than never using it at all. We can change this to be more sophisticated later if someone wants to.
- Remove the freeFightMood execution from the thesis delivery function, since this can be called at the end of the day and will end up using things like blue taffy. This function is already called before thesis in our free fights section, so this should only effect the end-of-day thesis during barfturn
- Getting meat from the candy cane sword cane hidden city shrine


packages/garbo/src/fights.ts
```diff
@@ -1847,7 +1847,7 @@ const freeKillSources = [
   new FreeFight(
     () =>
       !get("_missileLauncherUsed") &&
-      getCampground()["Asdon Martin keyfob"] !== undefined,
+      getCampground()["Asdon Martin keyfob (on ring)"] !== undefined,
```
horrible-little-slime: switch this to `getWorkshedItem` or whatever the heck it's called

packages/garbo/src/tasks/dailyItems.ts
```diff
@@ -432,6 +434,31 @@ const DailyItemTasks: GarboTask[] = [
...
+    ready: () =>
+      have($item`candy cane sword cane`) &&
+      canAdventure($location`An Overgrown Shrine (Northeast)`),
+    completed: () => get("_candyCaneSwordOvergrownShrine", true),
+    do: () => {
+      visitUrl("adventure.php?snarfblat=348");
+      runChoice(4);
+      runChoice(6);
+    },
+    outfit: () => ({
```
horrible-little-slime, line 456: imo don't have this one be a function, it's a constant, that's fine

horrible-little-slime (commented):
LGTM

PR #1797, opened 2024-01-17
Candy cane sword cane failsafes

Shiverwarp:
Because this is an actual snarfblat it needs some extra edge case handling.

We don't want to go there overdrunk unless we have wineglass, and I added extra tries to the skip limit in case we hit things like June Cleaver there (might have it equipped in offhand when not drunk?)

### DIFF
--- a/packages/garbo/src/tasks/dailyItems.ts
+++ b/packages/garbo/src/tasks/dailyItems.ts
@@ -3,15 +3,18 @@ import {
   abort,
   buy,
   canAdventure,
+  canEquip,
   cliExecute,
   getCampground,
   getClanLounge,
   getMonsters,
+  inebrietyLimit,
   Item,
   itemDropsArray,
   itemPockets,
   mallPrice,
   meatPockets,
+  myInebriety,
   pickedPockets,
   pocketItems,
   pocketMeat,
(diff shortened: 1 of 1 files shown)


eegeeZA:
Can you also add the case for when there is a "dense liana"? I used an antique machete after completing the first fight to clear all four areas for future.

```
[486976] An Overgrown Shrine (Northeast)
Preference lastEncounter changed from The Mad Tea Party to dense liana
Encounter: dense liana
Round 0: eegee wins initiative!
Round 1: Wafferson takes a pull on the hookah, and immediately goes into a coughing fit, blowing a cloud of smoke in your face. Since this is magic smoke, instead of giving you cancer (or, perhaps in addition to giving you cancer), it gives you an effect:
You acquire an effect: Demonic Taint (6)
Round 1: Your toy train moves ahead to the Trackside Diner. Somebody dropped some food by the tracks, you pick it up!
Preference trainsetPosition changed from 162659 to 162660
You acquire an item: bowl of cottage cheese
Preference _concoctionDatabaseRefreshes changed from 180 to 181
Round 1: eegee executes a macro!
KoLmafia thinks it is round 2 but KoL thinks it is round 1
You're on your own, partner. (Macro Aborted: 37 instructions (In a row?!) executed without any actions being taken.)
```

eegeeZA:
I've never ascended 😇

horrible-little-slime:
I agree about not supporting it; having never ascended is both an unsupported character state and _extremely_ easy to rectify.

gausie (approved):

PR #1799, opened 2024-01-20
Don't wander at Pandamonium Slums

Shiverwarp:
Superlikely unskippable "Highway to Hey Deze" non combat


Rinn (approved):

gausie:
We should maybe comment the reason each zone is in this list.

PR #1800, opened 2024-01-21
use VIP clan preference vs BAFH

atbell1:
This uses the global VIP clan preference vs hard coding BAFH, which may not be desired

### DIFF
--- a/packages/garbo/src/tasks/daily.ts
+++ b/packages/garbo/src/tasks/daily.ts
@@ -573,11 +573,11 @@ const DailyTasks: GarboTask[] = [
       getClanLounge()["Clan Carnival Game"] !== undefined &&
       isOnline("OnlyFax") &&
       Clan.getWhitelisted().find(
-        (c) => c.name === "Bonus Adventures from Hell",
+        (c) => c.name === globalOptions.prefs.vipClan,
       ) !== undefined,
     completed: () => get("_clanFortuneConsultUses") >= 3,
     do: (): void => {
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
We already check for whitelist access to BAfH, which is free and automated.

Additionally, this won't actually work--we need OnlyFax to be in the required clan when we send the Zatara request, and OnlyFax remains stationed in BAfH (except when delivering faxes, which takes little enough time that the default behavior of repeating a task thrice should work almost always).

atbell1:
I see your point regarding the functionality itself given the context with OnlyFax, however the assumption that the player can swap to BAFH just because they are whitelisted to it is false at least in my edge case where I'm the clan leader of my VIP preferred clan. Given that this won't work, perfectly reasonable to reject it, but it's there a way to skip this daily task through some other pref? If I need to just fork something for my edge case forever, so be it but I'm happy to try to implement a solution for anyone else in the same boat if you can provide a thought of what would be reasonable for the script.

gausie:
I'll close this PR. I would welcome a PR that skips this step if you cannot leave your current clan due to being its leader. My advice to you is to create a multi and set it as the leader of your clan so you are free to move.

atbell1:
Good advice, didn't know that was kosher with MultiCzar, but will look into that for my personal setup. I will also see if I can't test some code that catches the can't move clans exception and skips instead of fails for this and try again with a PR just to see if this can't be made more robust. Really appreciate the responses!

Issue #1806, opened 2024-01-28
[BUG] No longer usinging a pocket wish to fight embezzlers when we can't clover


thingolober:
**Describe the bug**
In my run of garbo post CS, after the ween merge garbo will attempt to use a clover to fight an embezzler instead of a pocket wish. I don't have access to clover an embezzler, so I stay lucky the whole time and none of my copies are used. Before this merge garbo used pocket wishes correctly

**To Reproduce**
Don't have access to a lucky embezzler (or locket/any other preferred sources) and run garbo while the clover condition would return true otherwise. Relevant part of my log attached
[temp_embezzler_fail_log.txt](https://github.com/loathers/garbage-collector/files/14074758/temp_embezzler_fail_log.txt)

**Expected behavior**
Garbo uses a pocket wish instead of a clover for an embezzler fight if you can't access embezzlers

**Additional context**
Pretty sure this is the offending line, https://github.com/loathers/garbage-collector/blob/85c23872c0766e8ae75a7eb4b2be8ba7679aff15/packages/garbo/src/embezzler/fights.ts#L884 which should be `===`

VuvuzelaVirtuoso:
[Vuvuzela_Virtuoso_postCS_lucky.txt](https://github.com/loathers/garbage-collector/files/14088176/Vuvuzela_Virtuoso_postCS_lucky.txt)

Adding a snippet of this happening in my log as well--I experienced this bug on the 25th, during the only CS aftercore I've been in since January 4th. Issue has also been noted noted as happening in the same way by Borzon (#2348974) elsewhere.

PR #1807, opened 2024-01-29
Flip equality on embezzler clover check

thingolober:
Flip equality when we're targeting embezzlers specifically

### DIFF
--- a/packages/garbo/src/embezzler/fights.ts
+++ b/packages/garbo/src/embezzler/fights.ts
@@ -881,7 +881,7 @@ export const emergencyChainStarters = [
     () => {
       const potential = Math.floor(copyTargetCount());
       if (potential < 1) return false;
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
I think this is _more_ correct than existing behavior, but still not quite correct. Importantly, if `target` isn't embezzler (which, right now, it always will be. But as you've likely inferred, some day it may not be), we shouldn't be able to start a chain using a clover, but this incarnation of the code suggests we will.

I think we want the logic to look a little more like
```
if (globalOptions.target !== embezzler) return false;
if (!canAdventure($location`Knob Goblin Treasury`)) return false;
```
followed by the rest of the code

thingolober:
If we want to assume it'll only be path for embezzlers going forward I'll add those and move them up in the function

horrible-little-slime:
Resolves #1806

PR #1810, opened 2024-02-01
Blank out recovery script while running garbo

gausie:
Universal Recovery messes up June Cleaver stuff


horrible-little-slime:
If we're going to do this, which I'm not 100% confident we should, we should also update the readme

gausie:
Well, UR can break garbo now so we need to do something smart. Maybe blank out UR before we choose to be beaten up Beaten Up from a June Cleaver NC?

horrible-little-slime:
I think that makes sense. Should it be a pre and a post, or a `with` around
the do, using adv1 instead of returning a location?

PR #1816, opened 2024-03-01
use the wardrobe-o-matic at level 20

horrible-little-slime:
Upon seeing that it was in dailyItems, I'm not actually sure how desirable this is.

### DIFF
--- a/packages/garbo/src/tasks/post/index.ts
+++ b/packages/garbo/src/tasks/post/index.ts
@@ -294,6 +294,16 @@ function leafResin(): GarboPostTask {
   };
 }
 
+function wardrobeOMatic(): GarboPostTask {
+  return {
+    name: "Wardrobe-o-Matic",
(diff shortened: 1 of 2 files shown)


packages/garbo/src/tasks/post/index.ts
```diff
@@ -294,6 +294,16 @@ function leafResin(): GarboPostTask {
   };
 }
 
+function wardrobeOMatic(): GarboPostTask {
+  return {
+    name: "Wardrobe-o-Matic",
+    available: have($item`wardrobe-o-matic`),
+    ready: () => myLevel() >= 20,
```
Rinn, line 301: maybe lower the threshold if ascending? after i cs loop i usually enter day 2 at like lv17 and i don't think usually even hit 20 by eod.

Rinn (approved):

PR #1823, opened 2024-03-06
fix MPA session calculations

horrible-little-slime:


horrible-little-slime:
drafted until we merge this: https://github.com/pstalcup/garbage-collector/pull/4

PR #1833, opened 2024-03-27
pick up and exchange the fused fuse at EOD

### LINKED ISSUE #1713: [FEATURE] Get fused fuse at the end of the day \ overdrunk turns
Now that we have the ability to run end of day \ overdrunk turns we should just automatically use clara's bell to do the fused fuse volcano quest instead of printing a message about it.

horrible-little-slime:
Resolves #1713

### DIFF
--- a/packages/garbo/src/tasks/barfTurn.ts
+++ b/packages/garbo/src/tasks/barfTurn.ts
@@ -34,6 +34,7 @@ import {
   GingerBread,
   have,
   questStep,
+  realmAvailable,
   set,
   SourceTerminal,
   sum,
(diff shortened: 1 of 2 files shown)


packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -260,6 +261,17 @@ function willDrunkAdventure() {
   return have($item`Drunkula's wineglass`) && globalOptions.ascend;
 }
 
+function canGetFusedFuse() {
+  return (
+    realmAvailable("hot") &&
+    ([1, 2, 3] as const).some(
+      (it) => get(`_volcanoItem${it}`) === $item`fused fuse`.id,
+    ) &&
+    !globalOptions.clarasBellClaimed &&
```
Rinn, line 270: won't globalOptions.clarasBellClaimed be true here because it's set in checkVolcanoQuest()?

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -260,6 +261,18 @@ function willDrunkAdventure() {
   return have($item`Drunkula's wineglass`) && globalOptions.ascend;
 }
 
+function canGetFusedFuse() {
+  if (get("noncombatForcerActive")) return true;
```
Rinn, line 265: the return true here doesn't check for the realm being available or the fused fuse being available in _volcanoItem#

Rinn (approved):

PR #1837, opened 2024-04-01
Add April Fool's Handling for Rufus

pstalcup:


horrible-little-slime:
I'm not convinced this will work, because `chooseQuest` is actually secretly a function that takes args and the crash will happen there:
```typescript
export function chooseQuest(
  chooser: ({
    entity,
    artifact,
    items,
  }: {
    entity: Monster;
    artifact: Item;
    items: Item;
  }) => 1 | 2 | 3 | 4
): boolean {
  if (get("questRufus") !== "unstarted") return false;
  if (!have()) return false;
  withChoice(1497, "", () => {
    directlyUse(item);
    runChoice(
      chooser({
        artifact: toItem(get("rufusDesiredArtifact")),
        entity: get("rufusDesiredEntity") ?? $monster.none,
        items: toItem(get("rufusDesiredItems")),
      })
    );
  });
  ```
Although, wait, we use `toItem` for those so we should be fine?

horrible-little-slime (approved):

PR #1840, opened 2024-04-04
Support delayed support choices

gausie:
- **Delay unsupportedchoices in garbo-lib and add Ice Hotel choice**
- **Delay quartet choice too**
- **Bump for breaking changes to garbo-lib**

### DIFF
--- a/packages/garbo-lib/src/wanderer/index.ts
+++ b/packages/garbo-lib/src/wanderer/index.ts
@@ -9,7 +9,16 @@ import {
   print,
   totalTurnsPlayed,
 } from "kolmafia";
-import { $familiar, $items, $location, get, maxBy } from "libram";
+import {
+  $familiar,
+  $item,
+  $items,
+  $location,
+  Delayed,
+  get,
+  maxBy,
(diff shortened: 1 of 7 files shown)


pstalcup (approved):

packages/garbo-lib/src/wanderer/index.ts
```diff
@@ -247,8 +263,37 @@ export class WandererManager {
...
+        sum(getActiveEffects(), (e) => this.options.effectValue(e, 1));
+      if (location === $location`The Ice Hotel`) {
+        const valueOfCertificates = get("_iceHotelRoomsRaided")
+          ? 0
+          : this.options.itemValue($item`Wal-Mart gift certificate`) * 3;
+        if (valueOfCertificates > valueOfTurn) {
+          return { ...choices, 1116: 5 };
+        }
+      }
+    }
```
horrible-little-slime, line 294: I don't think I love the placement of this logic here. maybe delay `unsupportedChoices` with an `options` arg and put the logic inside there?

PR #1844, opened 2024-04-07
Fix logical error in calculating extra time value

pstalcup:


horrible-little-slime (commented):
This....should do nothing? It looks like you doubly inverted it

horrible-little-slime:
https://www.youtube.com/watch?v=jRg6Q-JFL9Y

horrible-little-slime:
from @Shiverwarp in discord:
Actually, I think it should be what I just posted instead. Isn't there a bug in current implementation where here:

```ts
    for (let i = extraTimeRemaining; i > 0; i--) {
      if (extraTimeValue(i) > mallPrice($item`extra time`)) {
        if (acquire(1, $item`extra time`, extraTimeValue(i), false)) {
          use($item`extra time`);
        }
      } else break;
    }```

It checks the *lowest* value extra time first, and then will break out of the for loop early?

PR #1858, opened 2024-04-23
filter experienceFamiliars by a lower experience threshold for piccolo

### LINKED ISSUE #1856: [BUG] Apriling piccolo bug
If the script picks apriling piccolo, the script was silently freezing when it tries to use it for Pocket Professor.  It appears the issue is that the piccolo becomes unusable by a familiar past a certain weight/experience threshold.  My Pocket Professor was 19 pounds, and when trying to use it manually I got a message that 'Your familiar doesn't feel like playing the piccolo right now'.

Using the piccolo up on my dramadery fixed it enough to allow the script to proceed.

Recommended fix: If the piccolo is picked, the script should watch for the fail-to-use message, and if it comes up, use the piccolo for another familiar.

horrible-little-slime:
Resolves #1856

### DIFF
--- a/packages/garbo/src/tasks/dailyItems.ts
+++ b/packages/garbo/src/tasks/dailyItems.ts
@@ -137,11 +137,13 @@ function pickCargoPocket(): void {
 }
 
 const chooseAprilFamiliar = () => {
-  const experienceFamiliars = getExperienceFamiliars();
+  const experienceFamiliars = getExperienceFamiliars().filter(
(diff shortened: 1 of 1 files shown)


lewismd13 (approved):
I had to log into GitHub to review this because I haven't looked at code in so long that the cookie had expired and logged me out. But to my extremely rusty eyes it looks fine. 😅

ajcoppa (approved):

PR #1859, opened 2024-04-23
add jill to bestFairy list

horrible-little-slime:


Rinn:
Is the mode being set anywhere? I only see references to the candle being set to ultraviolet to use it as a meat familiar.

gausie (approved):

PR #1861, opened 2024-04-28
Choose best lucky adventure for saxophone

Shiverwarp:
For now this just chooses between using the Lucky adventure on Inhalers or on Embezzlers. We can add more later if we want.

I'm unsure if I have to make two versions of the nonBarfTurnTask for both sober and overdrunk (like we do for Daily Dungeon)

I based it more on what we have for getting the fused fuse

### DIFF
--- a/packages/garbo/src/resources/aprilingband.ts
+++ b/packages/garbo/src/resources/aprilingband.ts
@@ -12,22 +12,28 @@ const instruments: {
   {
     instrument: "Apriling band quad tom",
     value: () =>
-      globalOptions.prefs.valueOfFreeFight +
-      0.02 * garboValue($item`spice melange`),
+      (globalOptions.prefs.valueOfFreeFight +
+        0.02 * garboValue($item`spice melange`)) *
+      3,
   },
   {
     instrument: "Apriling band saxophone",
-    value: () => getBestLuckyAdventure().value(),
+    value: () => getBestLuckyAdventure().value() * 3,
   },
   {
     instrument: "Apriling band piccolo",
     value: () =>
       Math.max(
         0,
-        ...getExperienceFamiliars().map(
-          ({ expectedValue }) => expectedValue / 12,
-        ),
-      ) * 40,
+        ...getExperienceFamiliars().map(({ familiar, expectedValue }) => {
+          const usesAllowed = clamp(
+            Math.floor((400 - familiar.experience) / 40),
+            0,
+            3,
+          );
+          return (expectedValue / 12) * 40 * usesAllowed;
+        }),
+      ),
   },
 ];
 


packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -340,6 +342,30 @@ const NonBarfTurnTasks: AlternateTask[] = [
     spendsTurn: true,
     choices: { 1091: 7 },
   },
+  {
+    name: "Saxophone Lucky",
+    completed: () => !AprilingBandHelmet.canPlay("Apriling band saxophone"),
+    ready: () =>
+      have($item`Apriling band saxophone`) &&
+      getBestLuckyAdventure().location ===
+        $location`The Castle in the Clouds in the Sky (Top Floor)` &&
```
Rinn, line 351: I don't see a reason to check .location against the top floor? this won't work as soon as another location is added

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -340,6 +342,30 @@ const NonBarfTurnTasks: AlternateTask[] = [
...
+      getBestLuckyAdventure().value() > get("valueOfAdventure"),
+    do: $location`The Castle in the Clouds in the Sky (Top Floor)`,
```
Rinn, line 353: same here, just use .location

packages/garbo/src/lib.ts
```diff
@@ -1054,3 +1054,37 @@ export function aprilFoolsRufus() {
     visitUrl("questlog.php?which=7");
   }
 }
+
+type LuckyAdventure = {
```
Rinn, line 1058: this should contain an optional outfit and macro to be able to handle lucky adventures that are combats (mirror), as well as optional choices because a few also have choice adventures

packages/garbo/src/lib.ts
```diff
@@ -1054,3 +1054,37 @@ export function aprilFoolsRufus() {
...
+  location: Location;
+  value: () => number;
+};
+
+const luckyAdventures = [
+  {
+    location: $location`Cobb's Knob Treasury`,
+    value: () =>
+      canAdventure($location`Cobb's Knob Treasury`)
```
Rinn, line 1067: i would remove canAdventure from the value func and just filter those locations out in determineBestLuckyAdventure()

Rinn: it does mean that this could possibly return nothing and you'd need to handle that unless you can guarantee some lucky location will always be available

Rinn:
imo I feel like available lucky adventures and their results should go into garbo-lib or libram so it's shareable.

packages/garbo/src/tasks/barfTurn.ts
Rinn: why would we bother having separate tasks when here we only would care about the single most profitable adventure that isn't embezzlers? can't this just be generic to be able to handle whichever location is returned

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -340,6 +342,58 @@ const NonBarfTurnTasks: AlternateTask[] = [
...
+    prepare: () => {
+      if (!have($effect`Lucky!`)) {
+        AprilingBandHelmet.play($item`Apriling band saxophone`);
+      }
+    },
+    outfit: () => (sober() ? {} : { offhand: $item`Drunkula's wineglass` }),
```
horrible-little-slime, line 386: I don't think you need this when we're splitting by sobriety

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -340,6 +342,58 @@ const NonBarfTurnTasks: AlternateTask[] = [
...
```
horrible-little-slime, line 360: I don't think you need this when we're splitting by sobriety

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -340,6 +342,58 @@ const NonBarfTurnTasks: AlternateTask[] = [
...
+    combat: new GarboStrategy(() =>
+      Macro.abortWithMsg("Hit unexpected combat!"),
```
horrible-little-slime, line 362: Give us just a lil more detail in this abort message

packages/garbo/src/tasks/barfTurn.ts
horrible-little-slime: What I'd do is add a parameter to `LuckyAdventure` that describes what phase of garbo we use it in--`yachtzee` for the eventual fishy, `embezzler` for embezzlers, and `barf` for this, and do `getBestLuckyAdventure()?.phase === "barf"` or whatever. I'd also make it explicitly return null when the list of viable options is empty, rather than not filtering the list and always using the maxBy. But I guess that can be problematic, and it gets annoying to tell grimoire to adventure in a place that might not exist.

horrible-little-slime: Definitely do the first part tho

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -374,10 +373,9 @@ const NonBarfTurnTasks: AlternateTask[] = [
     ready: () =>
       !globalOptions.ascend &&
       have($item`Apriling band saxophone`) &&
-      getBestLuckyAdventure().location ===
-        $location`The Castle in the Clouds in the Sky (Top Floor)` &&
+      getBestLuckyAdventure().phase === "barf" &&
       getBestLuckyAdventure().value() > get("valueOfAdventure"),
-    do: $location`The Castle in the Clouds in the Sky (Top Floor)`,
+    do: getBestLuckyAdventure().location,
```
horrible-little-slime, line 378:
```
    do: () => getBestLuckyAdventure().location,
```
otherwise you're locking in one value at runtime

packages/garbo/src/resources/aprilingband.ts
```diff
@@ -12,12 +12,13 @@ const instruments: {
...
-      globalOptions.prefs.valueOfFreeFight +
-      0.02 * garboValue($item`spice melange`),
+      (globalOptions.prefs.valueOfFreeFight +
+        0.02 * garboValue($item`spice melange`)) *
+      3,
   },
   {
     instrument: "Apriling band saxophone",
-    value: () => getBestLuckyAdventure().value(),
+    value: () => getBestLuckyAdventure().value() * 3,
```
horrible-little-slime, line 21: Strip the * 3s from this and from the piccolo, and correct the piccolo value to 40 (that one's on me)

horrible-little-slime: (and from the quad tom ofc)

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -340,6 +342,56 @@ const NonBarfTurnTasks: AlternateTask[] = [
...
+    combat: new GarboStrategy(() =>
+      Macro.abortWithMsg("Unexpected combat while attempting Lucky! adventure"),
+    ),
+    turns: () =>
+      have($item`Apriling band saxophone`)
+        ? $item`Apriling band saxophone`.dailyusesleft
+        : 0,
+    sobriety: "sober",
+    spendsTurn: true,
+  },
```
horrible-little-slime, line 394: instead of repeating all this code I'd either compress them into one task or make a function like we see for daily dungeon

packages/garbo/src/resources/aprilingband.ts
```diff
@@ -17,22 +17,17 @@ const instruments: {
...
   },
   {
     instrument: "Apriling band piccolo",
     value: () =>
       Math.max(
         0,
-        ...getExperienceFamiliars().map(
-          ({ expectedValue }) => expectedValue / 12,
+        ...getExperienceFamiliars().map(({ familiar, expectedValue }) =>
+          familiar.experience < 281 ? expectedValue / 400 : 0,
```
horrible-little-slime, line 28: where is 281 coming from?

horrible-little-slime: Ahh I think I see--I would bring this back up to 361. It'll give overestimates more often than underestimates, but I still think it's accurate.

It's also worth noting that `expectedValue` is the value _per turn_. The reason I was dividing by 12 is because right now, experience familiars are written with the expectation that they'll get 12 exp per turn.

horrible-little-slime (approved):

PR #1862, opened 2024-04-28
Spring shoes wandering

Shiverwarp:


packages/garbo-lib/src/wanderer/gingerbreadcity.ts
```diff
@@ -0,0 +1,31 @@
...
+import { $item, $location, GingerBread } from "libram";
+
+export function gingerbreadFactory(
+  type: DraggableFight,
+  locationSkiplist: Location[],
+  options: WandererFactoryOptions,
+): WandererTarget[] {
+  if (
+    ["freefight", "freerun"].includes(type) &&
+    !locationSkiplist.some((loc) => GingerBread.LOCATIONS.includes(loc)) &&
```
horrible-little-slime, line 12: Is there a reason why we're going with this general a check instead of checking for the specific zone we're looking for?

packages/garbo-lib/src/wanderer/gingerbreadcity.ts
```diff
@@ -0,0 +1,31 @@
...
+      (availableAmount($item`sprinkles`) > 5 ||
+        haveOutfit("gingerbread best"))) ||
+      GingerBread.minutesToNoon() > 0)
+  ) {
+    return [
+      new WandererTarget(
+        "Gingerbread Minutes",
+        $location`Gingerbread Civic Center`,
+        // Arbitrary cutoff, where it's unlikely we'll reach the next NC. 50 value is also arbitrary, just to put it above default
+        options.estimatedTurns() < 100 ? 0 : 50,
```
horrible-little-slime, line 26: I feel like we can hone these down a little

packages/garbo-lib/src/wanderer/guzzlr.ts
```diff
@@ -94,7 +94,7 @@ export function guzzlrFactory(
   locationSkiplist: Location[],
   options: WandererFactoryOptions,
 ): WandererTarget[] {
-  if (Guzzlr.have()) {
+  if (Guzzlr.have() && _type !== "freerun") {
```
horrible-little-slime, line 97: If we're actually using the variable it shouldn't be preceded by an underscore anymore

packages/garbo-lib/src/wanderer/lib.ts
```diff
@@ -307,8 +310,12 @@ export function wandererTurnsAvailableToday(
       ? clamp(get(source.property), 0, source.max)
       : 0,
   );
+  // This is the ELG cooldown for spring shoes
+  const freeRun = canWanderCache["freerun"]
+    ? Math.floor(options.estimatedTurns() / 30)
+    : 0;
 
-  return digitize + pigSkinnerRay + yellowRay + wanderers;
+  return digitize + pigSkinnerRay + yellowRay + wanderers + freeRun;
```
horrible-little-slime, line 318: I don't think we want this. This function is used to determine things like "can we complete this guzzlr quest", and freeRuns aren't used for that. I guess they are for gingerbread city? Ugh I hate that things are different

packages/garbo-lib/src/wanderer/lovebugs.ts
```diff
@@ -27,7 +27,7 @@ export function lovebugsFactory(
   _locationSkiplist: Location[],
   options: WandererFactoryOptions,
 ): WandererTarget[] {
-  if (get("lovebugsUnlocked")) {
+  if (get("lovebugsUnlocked") && _type !== "freerun") {
```
horrible-little-slime, line 30: If we're actually using the variable it shouldn't be preceded by an underscore anymore

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -642,6 +644,28 @@ const BarfTurnTasks: GarboTask[] = [
       sobriety: "sober",
     },
   ),
+  wanderTask(
+    "freerun",
```
horrible-little-slime, line 648:
```suggestion
    "Spring Away",
```

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -642,6 +644,28 @@ const BarfTurnTasks: GarboTask[] = [
...
+      equip: $items`spring shoes`,
+      bonuses: new Map([[$item`carnivorous potted plant`, 500]]),
+    },
+    {
+      name: "Spring Shoes Freerun",
+      ready: () =>
+        have($item`spring shoes`) &&
+        romanticMonsterImpossible() &&
+        (getWorkshed() !== $item`model train set` ||
+          TrainSet.next() !== TrainSet.Station.GAIN_MEAT),
```
horrible-little-slime, line 659: Meat is special cased because it disappears when we run, not because all train stuff disappears when we run and meat is the best, right?

packages/garbo-lib/src/wanderer/gingerbreadcity.ts
horrible-little-slime: ahh that's fair: we'd be using a made up number for trains which means that everything is the fruit of the poisoned tree.

packages/garbo/src/tasks/barfTurn.ts
horrible-little-slime: oh god I thought it was the name, ignore me

horrible-little-slime: The name's right there below, I am The Fool

PR #1878, opened 2024-05-07
Add Postfreefights task to tend/pick mushroom crop

Tokoeka:


horrible-little-slime:
It's not clear to me when a user would encounter this situation. I thought we handled doing mushroom garden fights, why isn't this already being handled there?

PR #1883, opened 2024-05-11
Value Familiar Experience in outfits when using Chest Mimic

Shiverwarp:


horrible-little-slime (approved):
LGTM but I'd love one extra review on it just because this is delicate. Great work though, in the immortal words of Margot and the Nuclear So-and-Sos, I've got a bad case of the shivers!

PR #1884, opened 2024-05-11
Buff and extend before freefights

Shiverwarp:


packages/garbo/src/index.ts
```diff
@@ -533,6 +533,8 @@ export function main(argString = ""): void {
         });
 
         // 2. do some embezzler stuff
+        potionSetup(true);
+        useBuffExtenders(); // Buff and use extenders pre free fights primarily to extend buffs like Shadow Affinity
```
Tokoeka, line 537: Should the call of UseBuffExtenders() further down (new line 548) be removed as redundant?

Tokoeka (approved):

PR #1886, opened 2024-05-12
Powerful glove and macrometeorite are also switchmonster

Shiverwarp:


pstalcup (approved):

horrible-little-slime (approved):
It might be worth adding a `switchmonster` property to embezzler fights and passing that in, rather than using the string names. But that's for the future.

PR #1887, opened 2024-05-12
constantValueFamiliars mode

Shiverwarp:
Currently we're valuing constant value familiars during free fights that only get value during barf


pstalcup (dismissed):

horrible-little-slime (changes requested):
Instead of making the entire array a function, make the `value` function accept that argument

PR #1900, opened 2024-05-19
check your terrarium for familiar equipment when returning to the stash

### LINKED ISSUE #1899: Mayflower Bouquet Not Pulled or Returned to Clan Stash [BUG]
Mayflower Bouquet in clan stash is not pulled or returned to stash. Thought maybe this was due to it being not optimal but, when I have one in my inventory it does equip it during the run (during free fights). Thanks!

horrible-little-slime:
resolves #1899

### DIFF
--- a/packages/garbo/src/clan.ts
+++ b/packages/garbo/src/clan.ts
@@ -6,6 +6,7 @@ import {
   equip,
   equippedAmount,
   equippedItem,
+  familiarEquippedEquipment,
   getClanId,
   getClanName,
   handlingChoice,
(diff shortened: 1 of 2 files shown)


packages/garbo/src/clan.ts
```diff
@@ -201,6 +205,14 @@ export class StashManager {
             slots.forEach((s) => equip(s, $item.none));
           }
 
+          if (toSlot(item) === $slot`familiar` && !have(item, count)) {
+            for (const familiar of $familiars.all().filter(have)) {
+              if (familiarEquippedEquipment(familiar) === item) {
+                equip(familiar, $item.none);
+              }
+            }
```
gausie, line 213: Now I may be a simple country lawyer but

```ts
$familiars.all().filter(have).filter(f => familiarEquippedEquipment(f) === item).forEach(f => equip(f, $item.none))
```

horrible-little-slime:
Might as well eliminate the first filter, then

packages/garbo/src/clan.ts
horrible-little-slime: not a big `forEach` guy here

PR #1901, opened 2024-05-19
BALLS improvements

Rinn:
* use Macro.makeBALLSPredicate() rather than hasskill (or pass the skill directly to if_
* use updated trySkillRepeat instead of while_ as it now checks repeat hasskill

Requires a pending libram PR https://github.com/loathers/libram/pull/616


horrible-little-slime:
improvema balls

PR #1904, opened 2024-05-21
Add Mayam Support

Ignose:


packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,249 @@
...
+        0,
+        ...getExperienceFamiliars("free").map(
+          ({ expectedValue }) => expectedValue / 12,
+        ),
+      ) * 100,
+    ring: 1,
+  },
+  {
+    choice: "vessel",
+    value: () => 0, // How do we value MP regen?
```
Rinn, line 96: probably not worth it

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,249 @@
...
+    ring: 1,
+  },
+  {
+    choice: "eye",
+    value: () => 5, // +30% item drop, I don't know how to value that?
+    ring: 1,
+  },
+  {
+    choice: "chair",
+    value: () => 5, // +5 free rests, I don't know how to value that?
```
Rinn, line 80: I expect the only thing interesting is using free rests to generate cinch for more possible [Fiesta Exit](https://kol.coldfront.net/thekolwiki/index.php/Cincho:_Fiesta_Exit) casts

Also you'll want to double check if mafia is tracking these free tests, unsure if that was added yet.

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,249 @@
...
+    ring: 4,
+  },
+  {
+    choice: "sword",
+    value: () => 0,
+    ring: 1,
+  },
+  {
+    choice: "eye",
+    value: () => 5, // +30% item drop, I don't know how to value that?
```
Rinn, line 75: probably not worth it

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,249 @@
...
+    ring: 1,
+  },
+  {
+    choice: "lightning",
+    value: () => 0,
+    ring: 2,
+  },
+  {
+    choice: "bottle",
+    value: () => 0, // How do we value HP regen?
```
Rinn, line 106: def not worth it

gausie:
You could potentially do `new Potion($item.none, { effect, duration }).gross(copyTargetCount())` for the effect values, and get a future-proof zero. But probably not worth it.

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,249 @@
...
+    ring: 3,
+  },
+  {
+    choice: "eyepatch",
+    value: () => 0,
+    ring: 3,
+  },
+  {
+    choice: "explosion",
+    value: () => 0,
```
gausie, line 136: I would love a `valueOfPvpFight`. Even if just as a garbo pref

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,249 @@
...
+    ring: 4,
+  },
+  {
+    choice: "clock",
+    value: () => {
+      if (globalOptions.prefs.valueOfAdventure === undefined) {
+        return 4000 * 5;
+      } else return globalOptions.prefs.valueOfAdventure * 5;
```
gausie, line 144: Why have this fallback? Seems problematic to embed a 4000 default here.

In any case, you would write this as

```suggestion
      return (globalOptions.prefs.valueOfAdventure ?? 4000) * 5;
```

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  for (const { combination } of combinationValues) {
+    const symbols = combination.split(" ");
+    if (symbols.every((symbol) => !usedSymbols.has(symbol))) {
+      selectedCombinations.push(combination);
+      symbols.forEach((symbol) => usedSymbols.add(symbol));
+      const yam4 = () => (yamUses().includes("yam4") ? 1 : 0);
+      const explosion = () => (yamUses().includes("explosion") ? 1 : 0);
+      const clock = () => (yamUses().includes("clock") ? 1 : 0);
+      const remainingUses = 3 - (yam4() + explosion() + clock());
```
gausie, line 210: Why are these functions if you're going to immediately execute them?

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,249 @@
...
+  const combinations = generateCombinations();
+  const combinationValues = combinations.map((combination) => ({
+    combination: combination.join(" "),
+    value: calculateCombinationValue(combination),
+  }));
+
+  combinationValues.sort((a, b) => b.value - a.value);
+
+  const selectedCombinations: string[] = [];
+  const usedSymbols: Set<string> = new Set();
```
gausie, line 200: Doesn't this get filled by the first loop? Is it cleared anywhere?

gausie (commented):
I want to see some comments when you're writing functions that are not immediately obvious. Just add a jsdoc block explaining what you're trying to achieve! The trick to making that easy is to do it before you write the function not after

packages/garbo/src/resources/mayam.ts
Ignose: 4K is the assumed MPA for Garbo, or was? at one time?

packages/garbo/src/resources/mayam.ts
Ignose: Changed all this logic, therefore closing

packages/garbo/src/resources/mayam.ts
Ignose: Beyond me. : )

packages/garbo/src/resources/mayam.ts
Rinn: the valueOfAdventure preference will always be set to something, it's a requirement to run garbo

horrible-little-slime: The real answer here is we should never use `globalOptions.prefs.valueOfAdventure`--we don't actually have any reason to access `valueOfAdventure` through globalOptions, and should instead access it through `get` whenever humanly possible. It's a native mafia preference! I don't even know why it's in `globalOptions` to begin with!

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+    value: () => 0, // +5 free rests, I don't know how to value that?
+    ring: 1,
+  },
+  {
+    choice: "fur",
+    value: () =>
+      Math.max(
+        0,
+        ...getExperienceFamiliars("free").map(
+          ({ expectedValue }) => expectedValue / 12,
```
horrible-little-slime, line 89: We no longer use 12 specifically, I should update the piccolo code to match the new reality

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+};
+
+const RESONANCE: Resonance[] = [
+  { rings: "eye yam2 eyepatch yam4", item: $item`Mayam spinach` },
+  { rings: "vessel yam2 cheese explosion", item: $item`stuffed yam stinkbomb` },
+  { rings: "yam1 meat cheese yam4", item: $item`yam and swiss` },
+  { rings: "yam1 lightning yam3 clock", item: $item`yam battery` },
+  { rings: "yam1 yam2 yam3 explosion", item: $item`thanksgiving bomb` },
+  // {rings: "yam1 meat eyepatch yam4", item: $item`yamtility belt`}
+];
```
horrible-little-slime, line 27: Several of these functions are made redundant by libram support. I would nix the redundant functions and update the type of `MayamRingSummon` to use a more precise typing rather than just "string"

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+          findBestRingValue() as MayamCalendar.CombinationString;
+        const includesFur = ringOptions
+          .split(" ")
+          .some((option) => option.includes("fur"));
+
+        // If we're going to use Fur, determine the best experience familiar and use it before using any rings
+        if (includesFur) {
+          const bestFamiliar = maxBy(
+            getExperienceFamiliars("free"),
+            (f) => f.expectedValue,
```
horrible-little-slime, line 242:
```suggestion
           "expectedvalue",
```
`maxBy` has an overload for this very use case!

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+          useFamiliar(bestFamiliar.familiar);
+        }
+
+        MayamCalendar.submit(ringOptions);
+
+        ringOptions.split(" ").forEach((choice, index) => {
+          const command = `mayam ${choice}`;
+          // Adjust ring index (1-based) and add delay between summons
+          setTimeout(() => cliExecute(command), index * 1000);
+        });
```
horrible-little-slime, line 253: Why do we have a timeout on this?

packages/garbo/src/resources/mayam.ts
horrible-little-slime: Out of scope for now for sure

packages/garbo/src/resources/mayam.ts
horrible-little-slime: rather than doing `as [MayamSymbol, MayamSymbol, MayamSymbol, MayamSymbol]`, we need `as MayamCalendar.Combination`. I'll add some type-safe helper functions to `MayamCalendar` at some point, to make it so you don't have to cast the types

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+export function findResonanceValue(input: string): number {
+  const result = findResonance(input);
+
+  // Check if the result is found
+  if (result) {
+    return garboValue(result.item);
+  }
+
+  return 0;
+}
```
horrible-little-slime, line 45:
```suggestion
export function getResonanceValue(combination: MayamCalendar.CombinationString): number {
  const result = MayamCalendar.getResonanceResult(combination);

 if (!result) return 0;
 if (result instanceof Item) {
   if (result === $item`yamtility belt`) return yamtilityValue();
   return garboValue(result);
 }
 return effectValue(result);
}
```
Are all effects of the same duration? or do we need to access that information somewhere?

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+  },
+  {
+    choice: "bottle",
+    value: () => 0, // How do we value HP regen?
+    ring: 2,
+  },
+  {
+    choice: "meat",
+    value: () => clamp(myLevel() * 100, 100, 1500),
+    ring: 2,
```
horrible-little-slime, line 112: We shouldn't need to store that, we can access it by looking at the `choice` parameter. Alternately, I'd break this up into either an object containing four arrays, or just four separate arrays, one for each ring.

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+);
+
+// Check whether any component of a resonance is used before adding it to the list for MaxBy
+function isResonanceAvailable(resonance: Resonance): boolean {
+  const usedChoices = yamUses()
+    .split(",")
+    .map((choice) => choice.trim());
+  const resonanceChoices = resonance.rings.split(" ");
+  return resonanceChoices.every((choice) => !usedChoices.includes(choice));
+}
```
horrible-little-slime, line 162: just use `MayamCalendar.available`

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+  );
+
+  // Check for the highest value resonance
+  const maxResonance = maxBy(AVAILABLE_RESONANCES(), (resonance) =>
+    findResonanceValue(resonance.rings),
+  );
+
+  // Sum the total ring value and determine the resonance value
+  const totalRingValue =
+    (ring1?.value() || 0) +
```
horrible-little-slime, line 197:
```suggestion
    (ring1?.value() ?? 0) +
```

On one hand, you should use nullish coalescence. On the other hand, what you really should do is use `sumNumbers` and define all four of your rings at once by saying `const rings = new Array(4).fill().map((_, i) => maxBy(filteredOptions.filter((option) => option.ring === i + 1), (option) => option.value());`. On the third hand, it shouldn't even look exactly like that, because I think having `filteredOptions` be one giant hodgepodge array feels wrong.

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+    (ring2?.value() || 0) +
+    (ring3?.value() || 0) +
+    (ring4?.value() || 0);
+  const maxResonanceValue = maxResonance
+    ? findResonanceValue(maxResonance.rings)
```
horrible-little-slime, line 202: imo just make the function handle an undefined input by returning `0` instead of using this ternary here

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+    return combination;
+  }
+}
+
+function summonTask(): GarboTask {
+  return {
+    name: "Mayam Summons",
+    completed: () =>
+      ["yam4", "explosion", "clock"].every((sym) =>
+        get("_mayamSymbolsUsed").includes(sym),
```
horrible-little-slime, line 224:
```suggestion
      MayamCalendar.remainingUses() === 0
```

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+      ),
+    do: () => {
+      while (
+        !["yam4", "explosion", "clock"].every((sym) =>
+          get("_mayamSymbolsUsed").includes(sym),
```
horrible-little-slime, line 229:
```suggestion
        MayamCalendar.remainingUses() > 0
```

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+        )
+      ) {
+        const ringOptions =
+          findBestRingValue() as MayamCalendar.CombinationString;
```
horrible-little-slime, line 233: Give the function this return type instead of casting it here. That being said, I think you can alltogether avoid using the `string` type in this resource file

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+          useFamiliar(bestFamiliar.familiar);
+        }
+
+        MayamCalendar.submit(ringOptions);
+
+        ringOptions.split(" ").forEach((choice) => {
+          const command = `mayam ${choice}`;
+          // Adjust ring index (1-based) and add delay between summons
+          cliExecute(command);
+        });
```
horrible-little-slime, line 253: Isn't this redundant? What's the purpose of this second `for` loop? I assume it was copied from fax code, which has a lot of specialized fallthrough code because it involves interacting with bots. This doesn't, it's more similar to apriling band or clip art.

horrible-little-slime (commented):
Strong start so far

packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,262 @@
...
+    ring: 1,
+  },
+  {
+    choice: "eye",
+    value: () => 0, // +30% item drop, I don't know how to value that?
+    ring: 1,
+  },
+  {
+    choice: "chair",
+    value: () => 0, // +5 free rests, I don't know how to value that?
```
horrible-little-slime, line 80: Imo this is going to pretty consistently be one of the 3 best outer ring glyphs, so it's worth bare minimum putting a dummyvalue in.

I would value it as a `CinchoDeMayo.have() ? 3 * 5 * felizValue() : 0`? maybe?

garbage-collector
horrible-little-slime, line 1: Anyone have any idea what's going on here?

packages/garbo/src/resources/mayam.ts
Ignose: Tis. Removed.

packages/garbo/src/resources/mayam.ts
```diff
@@ -188,8 +171,9 @@ function findBestRingValue():
   );
 
   // Check for the highest value resonance
-  const maxResonance = maxBy(AVAILABLE_RESONANCES(), (resonance) =>
-    findResonanceValue(resonance.rings),
+  const maxResonance = maxBy(
+    availableResonances(),
+    (resonance) => resonance.value,
```
horrible-little-slime, line 176: you can just pass `"value"` as a string in for this arg

packages/garbo/src/resources/mayam.ts
horrible-little-slime: This comment is the key to your most recent issue: you either want to structure it as like ```
const MAYAM_RING_OPTIONS = {
  1: [
      { choice: " vessel", value: () => 0 },
      etc
    ],
    2: [ etc ]
    etc
  } as const;
  ```
  or just as four separate arrays. But I'd do the former.

horrible-little-slime: the `as const` will help typescript handle the types, at least a little

Ignose: Four separate arrays was easier for me to understand and manage.

horrible-little-slime (commented):
It's actually easier for me to just do the things I've been typing up

PR #1905, opened 2024-05-25
attempt to summon pants with a taco shell if porquoise is too expensive

Rinn:


horrible-little-slime (commented):
A few things:
1. This codeblock has gotten pretty big and unwieldy, let's move it to `/resources` and chunk it out into a couple of functions, then import it
2. When I first wrote this code, we weren't fighting 200+ embezzlers per day, and we could safely assume that your porquoise-clover pants were being used for almost every barf turn. Repaid Diaper + Robortender often beats porquoise-clover for embezzlers, and taco pants aren't guaranteed to consistently win for barf. I think we might need to re-evaluate some of the assumptions that go into this. Then again, if we can guarantee that 10-leaf clover and taco shell are always cheap, this probably isn't the end of the world.
3. Do we have any reason to believe ten-leaf clovers will ever rise in price? Taco shells are nicely pegged by the fact that they're in an npc shop, but I don't have the same guarantee for clovers. I guess a lot are being made by newark-chugging robortenders? That's probably fine then, ignore this last bullet.
4. Should we make an excavator project for pantogram pants droprates?

Rinn:
yeah that's fair, i don't have time to go more into this rn so gonna close this one and we can reevaluate

PR #1913, opened 2024-06-03
feat: minimally support the Mini Kiwi as a drop familiar

midgleyc:
Requires a new libram for the familiar and drop item.

This should be enough to crash the item prices without aviator goggles support.


horrible-little-slime:
That being said I'll cut a new `eslint-plugin-libram` momentarily

midgleyc:
I thought constant value familiars were those where every turn increments towards something, e.g. the Rockin' Robin's "drop every 30 turns", which makes every turn spent with them valuable. Meanwhile drop familiars were those where drops are random, as the Mini Kiwi's is.

horrible-little-slime:
I'd be happy to document this if we can think of the best place to do it.

midgleyc:
Okay, I just threw it above the `const` declaration in both files, giving something to search for to find the other file.

horrible-little-slime (approved):
yell heah

PR #1916, opened 2024-06-05
Abort on breaking diet price cap

Shiverwarp:


horrible-little-slime (approved):
I would describe myself as "pretty sure" this is correct

PR #1918, opened 2024-06-06
Add Kiwi Aioli support

Ignose:
Use Kiwi Aioli if price < MPA * fullness of the food about to be consumed. Create Kiwi Aioli if mini kiwis are cheaper in the mall.


packages/garbo/src/diet.ts
```diff
@@ -109,6 +109,13 @@ function eatSafe(qty: number, item: Item) {
   if (mallPrice($item`fudge spork`) < 3 * MPA && !get("_fudgeSporkUsed")) {
     eat($item`fudge spork`);
   }
+  if (
+    mallPrice($item`mini kiwi aioli`) < MPA * item.fullness ||
+    mallPrice($item`mini kiwi`) * 5 < MPA * item.fullness
+  ) {
+    acquire(1, $item`mini kiwi aioli`, MPA);
+    use($item`mini kiwi aioli`);
+  }
```
horrible-little-slime, line 118: Nix this, add it to `itemActions`

packages/garbo/src/diet.ts
```diff
@@ -1027,6 +1036,13 @@ export function consumeDiet(diet: Diet<Note>, name: DietName): void {
   );
   acquire(seasoningCount, $item`Special Seasoning`, MPA);
 
+  const aioliCount = sum(diet.entries, ({ menuItems, quantity }) =>
+    menuItems.some((menuItem) => menuItem.item === $item`mini kiwi aioli`)
+      ? quantity
+      : 0,
+  );
+  acquire(aioliCount, $item`mini kiwi aioli`, MPA * 1.5); // We reallly should set the price limit based on the adventure gain of the actual food
```
horrible-little-slime, line 1044: We similarly shouldn't need this because it's not similar to seasoning, it's similar to forks

packages/garbo/src/diet.ts
Ignose: No idea how to make `itemActions` work with this

horrible-little-slime:
That is indeed what I was looking for.

PR #1931, opened 2024-06-10
Remove symlink

gausie:
I can't even work out where this came from


horrible-little-slime:
I remember catching this and I still accidentally merged it. I think it was part of #1918, although it's in #1904 as well

PR #1941, opened 2024-06-19
Support bindlestocking crit drops

Rinn:
Unsure if the math will ever work out for this especially since it takes two slots but worth calculating. Pretty much limited to only seal clubbers as otherwise it is difficult to crit when the weapon and off-hand slots are used.

The wiki drop rates I sourced from are probably slightly off but all that really matters here is what appears to be a 1% combined drop rate for fancy chocolate and jawbruisers


horrible-little-slime (approved):

eegeeZA:
Happy to report that `bindlestocking` was equipped during my Garbo run today, and it even managed to get a `fancy chocolate` drop. Thanks :)

PR #1948, opened 2024-06-21
Account for weightAdjustment in maxFamiliarDamage

Shiverwarp:


packages/garbo/src/lib.ts
```diff
@@ -949,22 +950,27 @@ function maxCarriedFamiliarDamage(familiar: Familiar): number {
 function maxFamiliarDamage(familiar: Familiar): number {
   switch (familiar) {
     case $familiar`Cocoabo`:
-      return familiarWeight(familiar) + 3;
+      return familiarWeight(familiar) + weightAdjustment() + 3;
```
Rinn, line 953: can you pull the total weight into a variable so we're not writing the same addition for every familiar

Rinn (changes requested):

Rinn (approved):

PR #1976, opened 2024-07-18
rebased version of the bullseye branch

horrible-little-slime:
I can't force-push to ignose


horrible-little-slime:
@gausie we need to hammer out a time tomorrow to fix the mafia bug behind this

PR #1986, opened 2024-07-28
use Shadow Bricks for sandworms if profitable to do so

Ignose:
The logic for when it's profitable may need work, as we're checking current item% modifier rather than first building a sandworm outfit. If there's a better way to do this, I'm open to it.

### DIFF
--- a/packages/garbo/src/fights.ts
+++ b/packages/garbo/src/fights.ts
@@ -22,6 +22,7 @@ import {
   isBanished,
   Item,
   itemAmount,
+  itemDropModifier,
   itemDropsArray,
   Location,
   mallPrice,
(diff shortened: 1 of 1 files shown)


packages/garbo/src/fights.ts
```diff
@@ -1928,6 +1929,36 @@ const freeKillSources = [
     },
   ),
 
+  new FreeFight(
+    () =>
+      globalOptions.ascend &&
+      0.1 * garboValue($item`spice melange`) * itemDropModifier() >
```
horrible-little-slime, line 1935: use either libram's `getModifier` or mafia's `numericModifier` rather than `itemDropModifier`

Also this will always be an extremely huge number

packages/garbo/src/fights.ts
```diff
@@ -1928,6 +1929,36 @@ const freeKillSources = [
...
+        garboValue($item`shadow brick`)
```
horrible-little-slime, line 1936: We use garboValue for items we're selling, and mallPrice for items we're buying

garbage-collector
horrible-little-slime, line 1: Stop that!

Ignose: Stop what?

packages/garbo/src/fights.ts
```diff
@@ -1932,8 +1931,8 @@ const freeKillSources = [
   new FreeFight(
     () =>
       globalOptions.ascend &&
-      0.1 * garboValue($item`spice melange`) * itemDropModifier() >
-        garboValue($item`shadow brick`)
+      0.1 * garboValue($item`spice melange`) * numericModifier("Item Drop %") >
```
gausie, line 1934: Is `Item Drop %` valid? I've never seen it before. It can't be

PR #1987, opened 2024-07-29
Add support for fighting heavy rains wanderers

pstalcup:


gausie (approved):

packages/garbo/src/combat.ts
```diff
@@ -108,6 +110,25 @@ export class Macro extends StrictMacro {
...
+      myPath() === HeavyRains.path,
+      Macro.if_([...HeavyRains.wanderers], macro),
+    );
+  }
+
+  static ifHeavyRainsWanderer(macro: Macro): Macro {
+    return new Macro().ifHeavyRainsWanderer(macro);
+  }
+
+  ifWanderer(macro: Macro): Macro {
```
horrible-little-slime, line 124: I think this name is too general given that we use `wanderer` to refer to things like sausage goblins and deliberately encountered embezzlers in garbo

packages/garbo/src/combat.ts
```diff
@@ -108,6 +110,25 @@ export class Macro extends StrictMacro {
...
+    return this.ifHolidayWanderer(macro).ifHeavyRainsWanderer(macro);
```
horrible-little-slime, line 125:
```suggestion
    return this.if_([...getTodaysHolidayWanderers(), ...HeavyRains.wanderers], macro);
```
imo

packages/garbo/src/combat.ts
gausie: Hmm. ifPassiveWanderer?

horrible-little-slime (approved):

PR #1996, opened 2024-08-09
mayam micro-optimizations

horrible-little-slime:
This is a little bit of a git disaster thanks to the base change, but the total diff is correct and just and morally true.


packages/garbo/src/resources/mayam.ts
```diff
@@ -0,0 +1,182 @@
...
+        const result = [] as MayamCalendar.CombinationString[][];
+        for (const combinationGroup of acc) {
+          result.push(...expandCombinationGroup(combinationGroup));
+        }
+        return result;
+      },
+      [[]] as MayamCalendar.CombinationString[][],
+    );
+  return maxBy(combinationGroups, (group) => sum(group, valueCombination));
+}
```
Pantocyclus, line 161:
```ts
function getValidCombinations(
  forbiddenSymbols: MayamCalendar.MayamSymbol[]
): MayamCalendar.Combination[] {
  const validCombinations: MayamCalendar.Combination[] = [];
  const validRing0Symbols = MayamCalendar.RINGS[0].filter(
    (sym) => !forbiddenSymbols.includes(sym)
  );
  const validRing1Symbols = MayamCalendar.RINGS[1].filter(
    (sym) => !forbiddenSymbols.includes(sym)
  );
  const validRing2Symbols = MayamCalendar.RINGS[2].filter(
    (sym) => !forbiddenSymbols.includes(sym)
  );
  const validRing3Symbols = MayamCalendar.RINGS[3].filter(
    (sym) => !forbiddenSymbols.includes(sym)
  );

  validRing0Symbols.forEach((sym0) =>
    validRing1Symbols.forEach((sym1) =>
      validRing2Symbols.forEach((sym2) =>
        validRing3Symbols.forEach((sym3) => {
          const configuration: MayamCalendar.Combination = [
            sym0,
            sym1,
            sym2,
            sym3,
          ];
          validCombinations.push(configuration);
        })
      )
    )
  );

  return validCombinations;
}

// Given an array of existing combinations, and a new combination,
// Check that the new combination does not use any symbol that was
// already used in the existing combinations
function validNewCombination(
  existingCombinationIndices: number[],
  newCombinationIndex: number,
  validCombinations: MayamCalendar.Combination[]
): boolean {
  const combination: MayamCalendar.Combination[] =
    existingCombinationIndices.map((idx) => validCombinations[idx]);
  const configuration: MayamCalendar.Combination =
    validCombinations[newCombinationIndex];
  return combination.every((conf) =>
    conf.every((sym) => !configuration.includes(sym))
  );
}

export function getBestMayamCombinations(): MayamCalendar.Combination[] {
  const forbiddenSymbols = MayamCalendar.symbolsUsed();
  const n = MayamCalendar.remainingUses();
  if (n <= 0) {
    return [];
  }

  // Instead of splitting the combinations by the symbol of the innermost ring,
  // we lump them all in a single array - this helps with the optimization further down
  const validCombinations = getValidCombinations(forbiddenSymbols);

  // Compute the value of each combination
  const combinationValues = validCombinations.map((combination) =>
    valueCombination(combination)
  );
  // Then do an argsort() on the values from large to small
  const sortedIndices = combinationValues
    .map((val, idx) => [val, idx])
    .sort(([val1], [val2]) => val2 - val1)
    .map(([, idx]) => idx);

  const groups: MayamCalendar.Combination[][] = [];
  const groupValues: number[] = [];

  // This is the crux of the optimization
  // Let the combinations be ordered from the largest to the smallest such that, e.g.
  // Combination 1: [yam1, yam2, yam3, clock], Combination 2: ...
  // Let us pick groups of combinations that do not have overlapping symbols, e.g.
  // Group 1: [Combination 1, Combination 5, Combination 15]
  // Group 2: [Combination 2, Combination 3, Combination 9]
  // ...
  // For any given group = [a, b, c], we have value(a) >= value(b) >= value(c)
  // Furthermore, we pick the groups such for
  // Group i = [ai, bi, ci], Group j = [aj, bj, cj], we have
  // value(ai) >= value(aj) for all i < j
  // Let us collate the last index of each group, i.e. [c1, c2, ..., cj] = [15, 9, ..., cj]
  // Suppose we pick a new Group k = [ak, bk, ck] = [ci, bk, ck] where ci exists in [c1, c2, ..., cj]
  // Then we know that Group i = [ai, bi, ci] is always more optimal than (or equal to) Group k = [ci, bk, ck], since
  // ai + bi + ci >= ci + ci + ci >= ci + bk + ck, since ai >= bi >= ci >= bk >= ck
  // Furthermore, we also know that Group i must be more optimal than (or equal to) and group after k,
  // since ai >= bi >= ci = ak >= al >= bl >= cl for all l > k
  // We thus only need to evaluate Groups 1 to k - 1 to determine the optimal group
  const lastCombinationIndices: number[] = [];

  for (let i = 0; i < combinationValues.length; i++) {
    const idx1 = sortedIndices[i]; // For the i-th group, choose the i-th largest combination
    if (idx1 in lastCombinationIndices) {
      break; // if ai already exists in [c1, ..., ci-1], we do not have to search any further
    }

    const combinationIndices: number[] = [idx1];
    let totalValue = combinationValues[idx1];
    for (let j = 1; j < n; j++) {
      // Populate the group with the requisite number of combinations (can be < 3)
      for (let k = idx1 + 1; k < combinationValues.length; k++) {
        const idx2 = sortedIndices[k];
        // Greedily find the next largest bi, ci such that ai >= bi >= ci
        if (validNewCombination(combinationIndices, idx2, validCombinations)) {
          combinationIndices.push(idx2);
          totalValue += combinationValues[idx2];
          break;
        }
      }
    }
    groups.push(combinationIndices.map((idx) => validCombinations[idx]));
    groupValues.push(totalValue);
    lastCombinationIndices.push(
      combinationIndices[combinationIndices.length - 1] // Track the last index of this group
    );
  }

  // Of all the groups we have found, determine the group that has the maximal value
  const sortedGroupIndices = groupValues
    .map((val, idx) => [val, idx])
    .sort(([val1], [val2]) => val2 - val1)
    .map(([, idx]) => idx);

  print("Top 5 combinations");
  for (let i = 0; i < 5; i++) {
    const idx = sortedGroupIndices[i];
    print(
      `${i + 1}: ${groups[idx]
        .map(
          (combination) =>
            `${combination.join(", ")} (${valueCombination(combination)})`
        )
        .join(" | ")} [total: ${groupValues[idx]}]`
    );
  }

  return groups[sortedGroupIndices[0]];
}
```

I'm not sure the greedy approach above is optimal, so I tested out an optimal implementation myself by replacing the functions here with my own. Tested with 0, 1, 2 and 3 mayam uses left and seems to work fine (and runs fast too). It's not pretty but I think it could be worth considering/referencing if we want to optimize this part.

Pantocyclus: image
Here's an example at EOD (no embezzlers left) where the optimal group has a lower valued first combination (48697.3) than the first combination of the 3rd- (51400.1) and 5th-most (56798.2) optimal groups

Pantocyclus: image
Here is a nice example at EOD (no embezzlers left) where the optimal group is better than a potential greedy solution (the 1st and 2nd combinations are strictly worse than those in the second group).

We can verify its correctness by removing the `if (lastCombinationIndices.includes(idx1))` (early exit) optimization to see that it gives the same result for the first `n` groups. (In this case, the optimization exits early at group ~120, but tbh running the computations for the remaining 240 groups isn't expensive at all)

horrible-little-slime: I'm not sure if this achieves what we want. Our current implementation _should_ be producing optimal results, but is looking at too many combinations and thus takes an entire fifth of a second to calculate. I'd like to reduce the number of duplicate combinations we look at.

This looks like it's more computationally laborious, but I'm not confident it produces better results than the current implementation (it may well produce better results than this PR, which might be wrong).

Can you identify anything wrong or suboptimal about the current implementation? It uses the relatively simple algorithm of starting with `[[]]` and iteratively "expanding" each array into a collection of arrays by trying to add each possible resonance and trying to add a "greedy" option as well. (We discussed this briefly in discord but I wanted to post it here as well)

Pantocyclus: Yeah after digging into it a bit deeper it seems this is better - this prunes a tree rather than pruning a list, which chops off undesireable entries faster

(The underlying data structure being different also reconciles why I had the notion that both pruning techniques were incompatible with each other)

PR #2008, opened 2024-08-14
Alternative Option worth considering.

Ignose:
Folks don't want to item farm. That's fine. These changes create the option.

Draft PR because I can't be assed to do more than offer it as an option.

### DIFF
--- a/packages/garbo/src/combat.ts
+++ b/packages/garbo/src/combat.ts
@@ -260,6 +260,14 @@ export class Macro extends StrictMacro {
       .familiarActions()
       .tryEgg()
       .externalIf(
+        myFamiliar() === $familiar`Grey Goose` &&
+          $familiar`Grey Goose`.experience >= 36,
+        Macro.if_(
+          globalOptions.target,
+          Macro.trySkill($skill`Emit Matter Duplicating Drones`),
+        ),
+      )
+      .externalIf(
         have($skill`Extract Oil`) && get("_oilExtracted") < 15,
         Macro.if_(
           $monster`garbage tourist`,
--- a/packages/garbo/src/config.ts
+++ b/packages/garbo/src/config.ts
@@ -1,6 +1,7 @@
 import { Args } from "grimoire-kolmafia";
 import { Item, print } from "kolmafia";
 import {
+  $familiar,
   $item,
   $items,
   $monster,
@@ -8,6 +9,7 @@ import {
   ChestMimic,
   CombatLoversLocket,
   get,
+  have,
 } from "libram";
 
 const workshedAliases = [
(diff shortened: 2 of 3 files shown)


packages/garbo/src/combat.ts
```diff
@@ -259,6 +259,14 @@ export class Macro extends StrictMacro {
       .trySingAlong()
       .familiarActions()
       .tryEgg()
+      .externalIf(
+        myFamiliar() === $familiar`Grey Goose` &&
+          $familiar`Grey Goose`.experience >= 36,
```
docrostov, line 264: it might be good to make that xp reference an easily edited variable just in case we eventually swap out knights for something else that requires more XP to dupe. but as of now this looks like a good start! thanks seraph 💜

docrostov (commented):
not approving because it is a draft and we probably need to modmax famxp on those free wanderers to make this really sing, but i like the idea a lot. this + ensuring diner is always on should be a pretty good mix for massive horseradish generation

packages/garbo/src/config.ts
```diff
@@ -83,6 +85,10 @@ const defaultTarget =
 export const targettingMeat = () =>
   $monsters`cheerless mime executive, cockroach`.includes(globalOptions.target);
 
+export const targettingItems = () =>
+  have($familiar`Grey Goose`) &&
+  $monster`Witchess Knight` === globalOptions.target;
```
Rinn, line 90: it'd be more robust if we checked the monsters drops for items that can be duped

Ignose: Is there an easy way to do this? It doesn't seem to be any of the proxy records assigned to items.

packages/garbo/src/combat.ts
Ignose: We'd need to value the full item array and make some decisions, which should be feasible. We'd also need to do as Rinn suggested and determine whether the full array, or part, or even just the expensive target was dupable which may not be trivial.

packages/garbo/src/config.ts
Rinn: [itemDropsArray()](https://wiki.kolmafia.us/index.php/Item_drops_array) something similar to how we find [good free fights from the locket](https://github.com/loathers/garbage-collector/blob/815600b94a32e71c34b944c9597dee31e888586c/packages/garbo/src/tasks/freeFight.ts#L98) i'd wager

Ignose: You'd ideally want a function that maps the itemDropsArray() for each garboValue item and determine the profitability of emitting drones - but then (as Scotch noted) you'd also need to plan for more than one drone AND ensure that a constantValueFamiliar wasn't just better.

All in all, it would take a lot to be done well.

PR #2009, opened 2024-08-14
Restore HP if it's low during copychain

Ignose:
People are DYING out there, DYING!!! - so let's make sure they heal.

### DIFF
--- a/packages/garbo/src/embezzler/fights.ts
+++ b/packages/garbo/src/embezzler/fights.ts
@@ -10,8 +10,11 @@ import {
   mallPrice,
   myAdventures,
   myHash,
+  myHp,
+  myMaxhp,
   myRain,
   print,
+  restoreHp,
   retrieveItem,
   runChoice,
   runCombat,
@@ -152,6 +155,9 @@ export class CopyTargetFight implements CopyTargetFightConfigOptions {
     print(
       `Now running ${globalOptions.target} fight: ${this.name}. Stay tuned for details.`,
     );
+    if (myHp() < myMaxhp() * 0.5 || myHp() > 3000) {
+      restoreHp(clamp(myHp() / 2, 300, 3000));
+    }
     this.execute(options);
   }
 }


horrible-little-slime:
Not convinced on this for two reasons:
1. The condition seems wrong--shouldn't it be `&&`, not `||`?
2. We do restore hp between embezzler fights, when we run `postCombatActions` in fights.ts. If we're going to augment our healing, we should do so there.

garbage-collector
horrible-little-slime, line 1: In addition to my previous review, please stop committing whatever this is

Rinn: postCombatActions() is run after free fights in src\fights.ts and some other places but not during src\embezzlers\fights.ts and i don't think it even does hp recovery. maybe should be just calling safeRestore()?

Rinn:
> 
>     2. We do restore hp between embezzler fights, when we run `postCombatActions` in fights.ts. If we're going to augment our healing, we should do so there.

postCombatActions() is run after free fights in src\fights.ts and some other places but not during src\embezzlers\fights.ts and i don't think it even does hp recovery. maybe should be just calling safeRestore()?

horrible-little-slime:
> postCombatActions() is run after free fights in src\fights.ts and some other places but not during src\embezzlers\fights.ts and i don't think it even does hp recovery. maybe should be just calling safeRestore()?

```typescript
      // REMAINING EMBEZZLER FIGHTS
      let nextFight = getNextCopyTargetFight();
      while (nextFight !== null && myAdventures()) {
        print(`Running fight ${nextFight.name}`);
        const startTurns = totalTurnsPlayed();

        if (
          nextFight.draggable === "backup" &&
          have($skill`Musk of the Moose`) &&
          !have($effect`Musk of the Moose`)
        ) {
          useSkill($skill`Musk of the Moose`);
        }

        if (nextFight.gregariousReplace) {
          const crateIsSabered = get("_saberForceMonster") === $monster`crate`;
          const notEnoughCratesSabered = get("_saberForceMonsterCount") < 2;
          const weWantToSaberCrates = !crateIsSabered || notEnoughCratesSabered;
          if (weWantToSaberCrates) saberCrateIfSafe();
        }

        const location = new EmbezzlerFightRunOptions(nextFight).location;
        const underwater = location.environment === "underwater";

        const famSpec = familiarSpec(underwater, nextFight.name);

        setLocation(location);
        embezzlerOutfit({ ...nextFight.spec, ...famSpec }, location).dress();

        runEmbezzlerFight(nextFight, { action: nextFight.name });
        postCombatActions();

        print(`Finished ${nextFight.name}`);
        if (
          totalTurnsPlayed() - startTurns === 1 &&
          get("lastCopyableMonster") === globalOptions.target &&
          (nextFight.wrongEncounterName ||
            get("lastEncounter") === globalOptions.target.name)
        ) {
          eventLog.initialCopyTargetsFought++;
          eventLog.copyTargetSources.push(nextFight.name);
        }

        nextFight = getNextCopyTargetFight();

        if (
          romanticMonsterImpossible() &&
          (!nextFight || !nextFight.draggable)
        ) {
          doSausage();
          yachtzee();
        }
        doGhost();
        startWandererCounter();
      }
    });
    ```
it does get run in the embezzler parts of fights.ts!

Re-using `safeRestore` is a good idea, but again, I'd like to identify where the actual problem is, because we are running our post combat actions after these fights.

Rinn:
bleah i wish i had time to convert this to grimoire tasks it's too complicated right now

PR #2015, opened 2024-08-15
Activate babymode

pstalcup:


Malibu-Stacey (approved):

packages/garbo/src/config.ts
```diff
@@ -82,18 +83,28 @@ function stringToWorkshedItem(s: string): Item | null {
...
+          valueDrops,
+        ) || $monster`Witchess Knight`
+      );
+    }
+  }
+  return $monster`Knob Goblin Elite Guard Captain`;
+}
 
 export const targettingMeat = () =>
   $monsters`cheerless mime executive`.includes(globalOptions.target);
```
Rinn, line 110: this needs to consider guard capt. maybe just have it check monster base meat is above some threshhold and not free

PR #2022, opened 2024-08-19
Finally, a reason to ascend Turtle Tamer

Ignose:


horrible-little-slime:
We might want to add this just for mimes, and not for barf itself; out mp restoration can be pretty inefficient and maintaining these buffs can be a drain.

Tokoeka (approved):

horrible-little-slime:
Likely rendered redundant by #2028

PR #2034, opened 2024-08-24
feat: use rhino 1.7.15, remove corejs

midgleyc:
In https://github.com/kolmafia/kolmafia/pull/2345 (r27967) we updated rhino to 1.7.15. Update the build system to use it.

It includes both `flat` and `Object.fromEntries` (the latter from 1.7.14, apparently) so you shouldn't need core-js any more.


packages/garbo/src/resources/mayam.ts
```diff
@@ -104,7 +103,7 @@ function getBestGreedyCombination(
 function expandCombinationGroup<N extends number>(
   group: Tuple<MayamCalendar.Combination, N>,
 ): [...Tuple<MayamCalendar.Combination, N>, MayamCalendar.Combination][] {
-  const forbiddenSymbols = [...flat(group), ...MayamCalendar.symbolsUsed()];
+  const forbiddenSymbols = [...group.flat(), ...MayamCalendar.symbolsUsed()];
```
horrible-little-slime, line 106: I think we need to define a depth here, right? should this be `.flat(2)`?

(libram's `flat` behavior differs from js's, I think)

midgleyc: Interesting -- libram defaults to a depth of Infinity, while js defaults to a depth of 1.

PR #2035, opened 2024-08-24
use conspiracy keycards automatically

Rinn:


horrible-little-slime (commented):
Are these regular useable items?

Rinn:
> Are these regular useable items?

yes, usable quest items

PR #2041, opened 2024-08-26
move `safeInterrupt` to `available`

horrible-little-slime:
we rely on `postCombatActions` to `safeInterrupt` during many of our non-grimoirized sections of code; this will ensure that we can `safeInterrupt` even if we never select a task to execute.


docrostov (approved):
so this should effectively make it so you don't need to execute a specific task to hit the interrupt? i'm in, that's a good-ass change. i always facepalm when i start garbo without remembering to do something, cast safe interrupt, and have it go through all the diet procedures up to the first fight buffs before it stops. good change imo, assuming you have tested it this feels totally fine to merge to my tired eyes

PR #2042, opened 2024-08-26
Improve defaultTarget logic

Ignose:


packages/garbo/src/config.ts
```diff
@@ -83,12 +83,16 @@ function stringToWorkshedItem(s: string): Item | null {
   return validWorksheds[0].item;
 }
 
+// We want this to be a const because we don't want to determine we can no longer target Cheerless Mime Executive simply because we used all reminisces to start a chain
+const locketAvailable = CombatLoversLocket.reminiscesLeft() > 0;
```
docrostov, line 87: so this check ensures that you have a reminisce left that can be used before declaring it to be your way to start the chain. makes sense, reasonable addition.

packages/garbo/src/config.ts
```diff
@@ -83,12 +83,16 @@ function stringToWorkshedItem(s: string): Item | null {
...
+
 function defaultTarget() {
   if ($skills`Curse of Weaksauce, Saucegeyser`.every((s) => have(s))) {
     if (
       ChestMimic.have() ||
-      have($item`Clan VIP Lounge key`) ||
+      (have($item`Clan VIP Lounge key`) && !get("_photocopyUsed")) ||
       (CombatLoversLocket.have() &&
+        locketAvailable &&
```
docrostov, line 95: my only small caveat is that this doesn't quite handle edge cases -- i believe the locket monster pref doesn't remove monsters you've fought today, so this could be improved by making the locketAvailable logic also check that you have not reminisced the desired target today. but this + the check on photocopy usage seem straightforward and helpful.

docrostov (approved):
@horrible-little-slime you probably should be the one to merge but this seems like a perfectly fine start that might solve the issue loopers are having where they don't start their target on the 2nd try -- since you can't have a guaranteed start with a pocket wish, like in our old embezzie world, you have to ensure these sources are usable before declaring it your starting source.

packages/garbo/src/config.ts
horrible-little-slime: dont' define this as a const outside of the thing, just toss it into the big pile of ternary unless you're going to do that for everything

Ignose: We want it to be a const because the value of it could change over the course of the garbo run, and we don't want to accidentally change the target during the course of the garbo run. Otherwise, every time we call the function (which should be, but may not be, exactly once) we run the risk of changing targets.

packages/garbo/src/config.ts
Ignose: Locket does track the monsters reminisced, so I've included a check for this. Again, as part of the const so that reminiscing doesn't somehow change our target retroactively.

packages/garbo/src/config.ts
```diff
@@ -83,12 +83,20 @@ function stringToWorkshedItem(s: string): Item | null {
   return validWorksheds[0].item;
 }
 
+// We want this to be a const because we don't want to determine we can no longer target Cheerless Mime Executive simply because we used all reminisces to start a chain
+const locketAvailable =
+  CombatLoversLocket.reminiscesLeft() > 0 &&
+  !CombatLoversLocket.monstersReminisced().includes(
+    $monster`cheerless mime executive`,
+  );
+
```
horrible-little-slime, line 92:
```suggestion
```

packages/garbo/src/config.ts
```diff
@@ -83,12 +83,20 @@ function stringToWorkshedItem(s: string): Item | null {
...
   if ($skills`Curse of Weaksauce, Saucegeyser`.every((s) => have(s))) {
     if (
       ChestMimic.have() ||
-      have($item`Clan VIP Lounge key`) ||
+      (have($item`Clan VIP Lounge key`) && !get("_photocopyUsed")) ||
       (CombatLoversLocket.have() &&
+        locketAvailable &&
         CombatLoversLocket.unlockedLocketMonsters().includes(
           $monster`cheerless mime executive`,
         ))
```
horrible-little-slime, line 102:
```suggestion
      CombatLoversLocket.availableLocketMonsters().includes($monster`cheerless mime executive`) 
```

horrible-little-slime (changes requested):

PR #2043, opened 2024-08-26
Gooso dupe witchess knights

Ignose:
Also value trainset food drops more better

### DIFF
--- a/packages/garbo/src/fights.ts
+++ b/packages/garbo/src/fights.ts
@@ -438,6 +438,20 @@ function familiarSpec(underwater: boolean, fight: string): OutfitSpec {
       if (have($familiar`Reanimated Reanimator`)) {
         return { familiar: $familiar`Reanimated Reanimator` };
       }
+      if (
+        $monsters`Witchess Knight, Witchess Bishop`.includes(
+          globalOptions.target,
+        )
+      ) {
+        if (
+          get("gooseDronesRemaining") < copyTargetCount() &&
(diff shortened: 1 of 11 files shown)


packages/garbo/src/fights.ts
```diff
@@ -438,6 +438,20 @@ function familiarSpec(underwater: boolean, fight: string): OutfitSpec {
       if (have($familiar`Reanimated Reanimator`)) {
         return { familiar: $familiar`Reanimated Reanimator` };
       }
+      if (
+        $monsters`Witchess Knight, Witchess Bishop`.includes(
```
Rinn, line 442: I'd really like to see something where we check the target's item drops to detect if goose is appropriate instead of hard coding the monster list in several places. Could you at the moment move this to a function so it's easier to update?

Rinn (changes requested):

packages/garbo/src/fights.ts
Ignose: Updated to a function; we'd need some large mafia changes to make that work. Specifically, we'd need drops to be presented in drop order in addition to whether they're duplicatable to make this truly generic.

Ignose:
@Rinn I think I covered the request to make this generic as best I can; as discussed, we can do a much more robust Gooso Drone check if the itemDropsArray were ordered by drop position, but barring that, the best we can do is value it only if the array value is 1.

PR #2061, opened 2024-09-04
Heal during the PProf Chain using New Age Healing Crystals; acquire e…

### LINKED ISSUE #2051: [BUG] dyin' to mimes 2: the re-death of ivan garbovich
**Describe the bug**
i've waited a few days to report this because i wanted to make sure it wasn't a one-off thing. but my garbo has started dying to mimes again after being perfectly fine for a bit after #2019 was resolved. it may be due to some restoration settings, but i believe garbo does actually set up mafia to restore, so i am kind of at a loss as to what's going on. log for today attached:

[Captain_Scotch_20240829.txt](https://github.com/user-attachments/files/16803832/Captain_Scotch_20240829.txt)

**To Reproduce**
not sure, really! it sounds like this is mostly resolved for most others. it seems to happen daily at a few fights through the lecture chain.

**Expected behavior**
do not die to the mimes yolo

**Session Log**
[Captain_Scotch_20240829.txt](https://github.com/user-attachments/files/16803832/Captain_Scotch_20240829.txt)

at the end, haven't run any turns since it died

**Additional context**
here is my account state:
- level 32
- 540 mus, 596 myst, 1033 mox
- AT in normal aftercore, with most things unlocked
- pretty high shiny but not using any stash items

Ignose:
…nough crystals to cover the chain

Resolves #2051 (hopefully)

### DIFF
--- a/packages/garbo/src/fights.ts
+++ b/packages/garbo/src/fights.ts
@@ -32,10 +32,8 @@ import {
   myBuffedstat,
   myClass,
   myFamiliar,
-  myHp,
   myInebriety,
   myLevel,
-  myMaxhp,
   myPath,
   myThrall,
   myTurncount,
@@ -195,7 +193,10 @@ import { fakeSources } from "./target/fights";
 const firstChainMacro = () =>
   Macro.if_(
     globalOptions.target,
-    Macro.externalIf(isStrongScaler(globalOptions.target), Macro.delevel())
+    Macro.externalIf(
+      isStrongScaler(globalOptions.target),
+      Macro.delevel().tryNewAgeHeal(),
+    )
       .if_(
         `!${Macro.makeBALLSPredicate($skill`lecture on relativity`)}`,
         Macro.externalIf(
@@ -203,11 +204,6 @@ const firstChainMacro = () =>
           Macro.tryCopier($skill`Digitize`),
         )
           .tryCopier($item`Spooky Putty sheet`)
-          .externalIf(
-            myHp() <
-              Math.min(myMaxhp() * 0.3, get("garbo_restoreHpTarget", 2000)),
-            Macro.tryHaveItem($item`New Age healing crystal`),
-          )
           .tryCopier($item`Rain-Doh black box`)
           .tryCopier($item`4-d camera`)
           .tryCopier($item`unfinished ice sculpture`)
@@ -223,7 +219,10 @@ const firstChainMacro = () =>
 const secondChainMacro = () =>
   Macro.if_(
     globalOptions.target,
-    Macro.externalIf(isStrongScaler(globalOptions.target), Macro.delevel())
+    Macro.externalIf(
+      isStrongScaler(globalOptions.target),
+      Macro.delevel().tryNewAgeHeal(),
+    )
       .if_(
         `!${Macro.makeBALLSPredicate($skill`lecture on relativity`)}`,
         Macro.trySkill($skill`Meteor Shower`),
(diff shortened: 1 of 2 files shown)


packages/garbo/src/fights.ts
```diff
@@ -226,6 +235,11 @@ const secondChainMacro = () =>
           Macro.tryCopier($skill`Digitize`),
         )
           .tryCopier($item`Spooky Putty sheet`)
+          .externalIf(
+            myHp() <
+              Math.min(myMaxhp() * 0.3, get("garbo_restoreHpTarget", 2000)),
+            Macro.tryHaveItem($item`New Age healing crystal`),
+          )
```
horrible-little-slime, line 242: These can't be externalIfs, unfortunately; we're only evaluating the macro once, before we run the fight, when you're at full HP. Use the existing BALLS predicate!

packages/garbo/src/fights.ts
```diff
@@ -199,6 +203,11 @@ const firstChainMacro = () =>
...
```
horrible-little-slime, line 210: same as below

packages/garbo/src/fights.ts
horrible-little-slime: No, because it's still an externalIf

Ignose: Okay; I don't understand BALLS. We're want a regular ass if_ then?

packages/garbo/src/fights.ts
```diff
@@ -215,6 +218,7 @@ const secondChainMacro = () =>
   Macro.if_(
     globalOptions.target,
     Macro.externalIf(isStrongScaler(globalOptions.target), Macro.delevel())
+      .if_("hppercentbelow 30", Macro.tryItem($item`New Age healing crystal`))
```
horrible-little-slime, line 221: Is there a world where the mime is outhitting healing crystals? should we consider funkslinging?

packages/garbo/src/fights.ts
```diff
@@ -7,6 +7,7 @@ import {
   canEquip,
   cliExecute,
   closetAmount,
+  count,
```
horrible-little-slime, line 10: I have no idea what the kolmafia `count` function does but if I had to guess it's not what you want it to do here specifically

packages/garbo/src/fights.ts
```diff
@@ -509,6 +513,13 @@ export function dailyFights(): void {
             }
           }
 
+          if (
+            mallPrice($item`New Age healing crystal`) < expectedTargetProfit()
+          ) {
+            // Better to live and continue the chain than lose and not pprof fight at all
+            retrieveItem($item`New Age healing crystal`, count(fakeSources));
+          }
```
horrible-little-slime, line 521:
```suggestion
          if (
            mallPrice($item`New Age healing crystal`) < expectedTargetProfit() &&
            have($familiar`pocket professor`) && !get("_garbo_weightChain", false)
          ) {
            retrieveItem($item`New Age healing crystal`, clamp(15 - get("_pocketProfessorLectures"), 0, 15);
          }
```

We don't have complex estimation for the number of profcopies we fight

Ignose: Then we'll buy 20 and hope for the best. The logs I've seen, there are 3 - 5 remaining in the chain, we start with 20 and if issues persist we raise it.

packages/garbo/src/fights.ts
Ignose: The logs I've seen (including my own!) have universally been sub-500 damage in a fight near the end of the chain, with most being in the 100-150 damage range. We could funksling, but I'm worried it will be overkill. These are cheap though. Your call.

packages/garbo/src/fights.ts
Malibu-Stacey: > We don't have complex estimation for the number of profcopies we fight

It can be calculated once suitably attired for the chain
`ceil(square_root(familiar_weight(my_familiar()) + weight_adjustment() + my_familiar().soup_weight))`
is how I do it in ASH. Don't forget to add 2 for the memory chip.

Rinn:
probably unnecessary now, let's hold off unless we find there's some other target we need to deal with

horrible-little-slime:
A worse version of this somehow got merged earlier

PR #2063, opened 2024-09-04
Better item familiar valuations

Ignose:


packages/garbo/src/familiar/itemFamiliar.ts
```diff
@@ -61,7 +61,7 @@ export function bestFairy(): Familiar {
     }
 
     const bonuses = [
-      ...menu({ includeExperienceFamiliars: false }),
+      ...menu({ includeExperienceFamiliars: true }),
```
horrible-little-slime, line 64:
```suggestion
      ...menu(),
```
including experience familiars is the default, so we can safely do this.

PR #2064, opened 2024-09-04
execute meatMood during barf turns

horrible-little-slime:


packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -922,6 +923,7 @@ export const BarfTurnQuest: Quest<GarboTask> = {
             Macro.meatKill(),
           ).abort(),
       ),
+      prepare: () => meatMood().execute(estimatedGarboTurns()),
```
Rinn, line 926: minor: if dinseyRollercoasterNext is true there's no need to buff this turn

Rinn (approved):

PR #2071, opened 2024-09-06
Ensure we stop droning, in the event our free fight familiar is gooso…

Ignose:
… to charge the meat weight

### DIFF
--- a/packages/garbo/src/lib.ts
+++ b/packages/garbo/src/lib.ts
@@ -28,6 +28,7 @@ import {
   Monster,
   mpCost,
   myBjornedFamiliar,
+  myBuffedstat,
   myEnthronedFamiliar,
   myFamiliar,
   myFullness,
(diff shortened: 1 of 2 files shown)


packages/garbo/src/lib.ts
```diff
@@ -1146,7 +1148,8 @@ export function scalerCap(monster: Monster): number {
   return cap;
 }
 
-const STRONG_SCALER_THRESHOLD = 1_000;
+// What constitutes a strong scaler depends on our ability to take hits. Let's compare mainstat with 50 removed for safety.
+const STRONG_SCALER_THRESHOLD = myBuffedstat(myPrimestat()) - 50;
```
horrible-little-slime, line 1152: This is unrelated to this PR; remove it, and we can discuss it at another time

Rinn:
```suggestion
const STRONG_SCALER_THRESHOLD = 1_000;
```

Rinn (approved):

PR #2097, opened 2024-09-13
Automatically acquire Sept-Ember Censer items

Ignose:
Draft until:

1. Pref for having checked Sept-Ember exists
2. Lint updated
3. Code tested with both


packages/garbo/src/tasks/dailyItems.ts
```diff
@@ -199,6 +200,41 @@ const DailyItemTasks: GarboTask[] = [
     },
     spendsTurn: false,
   },
+  {
+    name: "Sept-Ember",
```
Rinn, line 204: Make two different tasks, one for checking the shop and one for purchasing

Rinn (changes requested):

packages/garbo/src/tasks/dailyItems.ts
Ignose: Okay

Rinn (approved):

Rinn (approved):

PR #2106, opened 2024-09-15
fix: use sellPrice over coinmasterPrice

midgleyc:
`coinmasterPrice` was a hardcoded list, which returned 0 for the censer items and lead to an infinite loop.

Remove the trap and use the native Mafia function.


horrible-little-slime:
Well, i just ran into this and arrived at the same fix as you. Whoops!

PR #2116, opened 2024-09-16
Better Gooso 2 electric boogaloo

Ignose:
Value drone emission as a free fight familiar; defer to Garbo math to determine when to emit drones vs use another free fight familiar.

Given the value of free copytargets is near mallmin (lol, lmao) we should probably only emit drones when the opportunity cost makes it make sense to do so.

### DIFF
--- a/packages/garbo/src/familiar/freeFightFamiliar.ts
+++ b/packages/garbo/src/familiar/freeFightFamiliar.ts
@@ -32,7 +32,7 @@ type MenuOptions = Partial<{
   extraFamiliars: GeneralFamiliar[];
   includeExperienceFamiliars: boolean;
   allowAttackFamiliars: boolean;
-  mode: "barf" | "free";
+  mode: "barf" | "free" | "target";
 }>;
 const DEFAULT_MENU_OPTIONS = {
   canChooseMacro: true,
@@ -73,20 +73,20 @@ export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
     }
 
     if (
-      mode === "free" &&
+      mode === "target" &&
       gooseDroneEligible() &&
       get("gooseDronesRemaining") < copyTargetCount()
     ) {
       familiarMenu.push({
         familiar: $familiar`Grey Goose`,
         expectedValue:
-            // It takes 9 experience to go from level 5 to 6 and emit a drone
-            clamp(
-              getModifier("Familiar Experience") / 9,
-              0,
-              // The limit to how valuable any emission can be is how many drones are actually gonna hit the copyTarget
-              copyTargetCount() - get("gooseDronesRemaining"),
-            ) * valueDrops(globalOptions.target),
+          // It takes 9 experience to go from level 5 to 6 and emit a drone
+          clamp(
+            getModifier("Familiar Experience") / 9,
+            0,
+            // The limit to how valuable any emission can be is how many drones are actually gonna hit the copyTarget
+            copyTargetCount() - get("gooseDronesRemaining"),
+          ) * valueDrops(globalOptions.target),
         leprechaunMultiplier: 0,
         limit: "experience",
       });
--- a/packages/garbo/src/familiar/freeFightFamiliar.ts
+++ b/packages/garbo/src/familiar/freeFightFamiliar.ts
@@ -73,26 +73,24 @@ export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
     }
 
     if (
+      mode === "free" &&
       gooseDroneEligible() &&
       get("gooseDronesRemaining") < copyTargetCount()
     ) {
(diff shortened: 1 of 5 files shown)


packages/garbo/src/familiar/freeFightFamiliar.ts
```diff
@@ -68,6 +72,25 @@ export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
...
+          clamp(
+            getModifier("Familiar Experience") / 9,
+            0,
+            // The limit to how valuable any emission can be is how many drones are actually gonna hit the copyTarget
+            copyTargetCount() - get("gooseDronesRemaining"),
+          ) * valueDrops(globalOptions.target),
+        leprechaunMultiplier: 0,
+        limit: "experience",
+      });
+    }
```
horrible-little-slime, line 92: Put this behind a check on `mode` or else it'll unnecessarily run on barf constantly

horrible-little-slime:
No one should merge this until mafia support for `differentiableQuantity` is live

packages/garbo/src/familiar/freeFightFamiliar.ts
Ignose: Do we call freeFightFamiliar for barf? I was under the impression the whole codeblock was build to mode: "free":

`const DEFAULT_MENU_OPTIONS = {
  canChooseMacro: true,
  location: $location`none`,
  extraFamiliars: [],
  includeExperienceFamiliars: true,
  allowAttackFamiliars: true,
  mode: "free",`

horrible-little-slime:
Mode check is relevant because the `menu` function does get passed into
barfFamiliar (even though free fight familiar does not)

packages/garbo/src/familiar/freeFightFamiliar.ts
```diff
@@ -68,6 +72,27 @@ export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
...
+                0,
+                // The limit to how valuable any emission can be is how many drones are actually gonna hit the copyTarget
+                copyTargetCount() - get("gooseDronesRemaining"),
+              ) * valueDrops(globalOptions.target)
+            : 0,
+        leprechaunMultiplier: 0,
+        limit: "experience",
+      });
+    }
+
```
horrible-little-slime, line 95:
```suggestion
    if (
      mode === "free" &&
      gooseDroneEligible() &&
      get("gooseDronesRemaining") < copyTargetCount()
    ) {
      familiarMenu.push({
        familiar: $familiar`Grey Goose`,
        expectedValue:
            // It takes 9 experience to go from level 5 to 6 and emit a drone
            clamp(
              getModifier("Familiar Experience") / 9,
              0,
              // The limit to how valuable any emission can be is how many drones are actually gonna hit the copyTarget
              copyTargetCount() - get("gooseDronesRemaining"),
            ) * valueDrops(globalOptions.target),
        leprechaunMultiplier: 0,
        limit: "experience",
      });
    }
```

horrible-little-slime: The intention of the `mode` check is that `copyTargetCount` can involve pagehits--I just merged a libram change that means it takes a lot fewer, but we still don't want to be checking it willy nilly

packages/garbo/src/lib.ts
```diff
@@ -198,10 +198,8 @@ export const gooseDroneEligible = () =>
   have($familiar`Grey Goose`);
 
 export function averageTargetNet(): number {
-  const goose = gooseDroneEligible() ? 2 : 1;
-
   return targettingItems()
-    ? valueDrops(globalOptions.target) * goose
+    ? valueDrops(globalOptions.target)
```
horrible-little-slime, line 202: How is this related to the current changes?

spaghetti-squash (changes requested):
One concern I have here is that right now, you'll be running goose happily against things like sausage goblins and proton ghosts, which doesn't seem like it's desirable. I think we need to add a third `mode` and start passing that `mode` in during our copy target fights. I can take a look at that later

packages/garbo/src/lib.ts
Ignose: We can no longer assume we'll use gooso, so we cannot add that multiplier to determine the benefit of fighting target. The modifier existed in tandem with the cost in Fight.ts that guaranteed gooso usage.

Ignose:
> One concern I have here is that right now, you'll be running goose happily against things like sausage goblins and proton ghosts, which doesn't seem like it's desirable. I think we need to add a third `mode` and start passing that `mode` in during our copy target fights. I can take a look at that later

I tried to address this; let me know if I did it correctly.

spaghetti-squash (approved):
Test this for a couple of days but LGTM

PR #2123, opened 2024-09-19
make tearaway pants a viable wanderer target

horrible-little-slime:


Ignose:
Would a simpler way to do this be to pass the wanderer in question to `freeFightOutfit()` and then value the pants as part of the outfight generated there when the wanderer in question happens to be a plant?

horrible-little-slime:
> Would a simpler way to do this be to pass the wanderer in question to `freeFightOutfit()` and then value the pants as part of the outfight generated there when the wanderer in question happens to be a plant?

Well, we're going to do that, but that's not a simpler way of making it a wanderer target.

PR #2132, opened 2024-09-26
Add Snapper as a valuable familiar for copyTarget freeFightFamiliars

Ignose:


packages/garbo/src/fights.ts
```diff
@@ -528,6 +529,13 @@ export function dailyFights(): void {
 
         const famSpec = familiarSpec(underwater, nextFight.name);
 
+        if (
+          famSpec === $familiar`Red-Nosed Snapper` &&
```
horrible-little-slime, line 533: Pretty confident this line will never evaluate to true

packages/garbo/src/familiar/freeFightFamiliar.ts
```diff
@@ -42,7 +42,10 @@ const DEFAULT_MENU_OPTIONS = {
   allowAttackFamiliars: true,
   mode: "free",
 } as const;
-export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
+export function menu(
+  options: MenuOptions = {},
+  fight?: string,
```
horrible-little-slime, line 47: Instead of doing this, add an `excludeFamiliar: Familiar[]` parameter to `MenuOptions`

packages/garbo/src/familiar/freeFightFamiliar.ts
```diff
@@ -73,7 +76,7 @@ export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
     }
 
     if (
-      mode === "target" &&
+      mode.includes("target") &&
```
horrible-little-slime, line 79: undo this as well

Rinn, line 79:
```suggestion
      mode === "target" &&
```

packages/garbo/src/familiar/lib.ts
```diff
@@ -131,3 +134,22 @@ export function estimatedBarfExperience(): number {
...
+export function snapperValue(): number {
+  const item = Snapper.phylumItem.get(globalOptions.target.phylum);
+  const denominator =
+    11 -
+    (Snapper.getTrackedPhylum() === globalOptions.target.phylum
+      ? Snapper.getProgress()
+      : 0);
+  if (denominator > copyTargetCount()) return 0;
+
+  if (!item) return 0;
```
Rinn, line 149:
```suggestion
  if (!item) return 0;
  
  const denominator =
    11 -
    (Snapper.getTrackedPhylum() === globalOptions.target.phylum
      ? Snapper.getProgress()
      : 0);
  if (denominator > copyTargetCount()) return 0;
```

packages/garbo/src/familiar/freeFightFamiliar.ts
```diff
@@ -91,6 +94,16 @@ export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
         limit: "experience",
       });
     }
+
+    if (mode.includes("target")) {
```
Rinn, line 98:
```suggestion
    if (mode === "target") {
```

packages/garbo/src/familiar/freeFightFamiliar.ts
```diff
@@ -91,6 +94,16 @@ export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
         limit: "experience",
       });
     }
+
+    if (mode === "target") {
```
Rinn, line 98: needs to check if you have snapper right?

Rinn: sorry also i made a few small tweaks from the browser and a file need to be formatted which is why the linter is complaining that's my bad

packages/garbo/src/familiar/lib.ts
```diff
@@ -131,3 +134,22 @@ export function estimatedBarfExperience(): number {
...
+  
+  const denominator =
+    11 -
+    (Snapper.getTrackedPhylum() === globalOptions.target.phylum
+      ? Snapper.getProgress()
+      : 0);
+  if (denominator > copyTargetCount()) return 0;
+
+  if (item === $item`envelope full of Meat`) {
```
Rinn, line 151: you can handle this in [value.ts](https://github.com/loathers/garbage-collector/blob/f22fa29b29e2f2897906e2e2392cb9a84811c14d/packages/garbo-lib/src/value.ts#L32)

packages/garbo/src/familiar/freeFightFamiliar.ts
Ignose: I don't think so? If we need a filter, I'd rather add a universal filter.

horrible-little-slime: This does need to check if you have snapper

horrible-little-slime: And the check _should_ take place here

packages/garbo/src/familiar/freeFightFamiliar.ts
```diff
@@ -128,13 +141,15 @@ export function menu(options: MenuOptions = {}): GeneralFamiliar[] {
...
-
-  return familiarMenu;
+  return familiarMenu.filter(
+    ({ familiar }) =>
+      (allowAttackFamiliars ||
+        !(familiar.physicalDamage || familiar.elementalDamage)) &&
+      !excludeFamiliar.some(
+        (excludedFamiliar) => excludedFamiliar === familiar,
+      ) &&
+      have(familiar),
```
horrible-little-slime, line 151: Nope, move it on up

packages/garbo/src/fights.ts
```diff
@@ -398,6 +399,14 @@ function familiarSpec(underwater: boolean, fight: string): OutfitSpec {
   }
 
   if (isFreeAndCopyable(globalOptions.target)) {
+    if (["Macrometeorite", "Powerful Glove"].includes(fight)) {
```
horrible-little-slime, line 402:
```suggestion
    if (fight.gregariousReplace) {
```

Ignose: This doesn't actually work and requires passing more than we are today to define the fight.

horrible-little-slime (approved):
Approved pending testing

Ignose:
> Approved pending testing

I've continued running this fork; when targeting Dudes it uses Snapper (expected) when targeting Goblins it does not (also expected). When targeting Hobos it does (lastly expected). Existing Gooso logic still works as well. Other free fight fams continue to be used.

Issue #2138, opened 2024-09-28
[BUG] sandworms skipped, freekill sources ignored ?


capncastom:
Hi!

Latest mafia / latest garbo

Ive noticed the past couple days, when running `garbo target="cockroach"` specifically, that garbo has been skipping sandworm kills and freekill uses.

not doing steely-eyed squint, 
jokester gun
shattering punches
batoomerang
shadow bricks, etc...

These resources also appear in tourguide, so mafia knows about them.

Fired up a 5 day-old version of garbo (thanks Tapion) and it immediately started fighting sandworms and using freekills.

a session log:

[capn_castom_20240928.txt](https://github.com/user-attachments/files/17175591/capn_castom_20240928.txt)

Tapionator:
Newest version skips sandworm fights as well.

capncastom:
r28073 is free-killing sandworms.
Thank you for your service!

PR #2139, opened 2024-09-28
Sandworm-valuation-changes

horrible-little-slime:
This has some weird order-of-operations issues with champagne bottle, and fails to solve the issue of the quad-tom

### DIFF
--- a/packages/garbo/src/tasks/freeGiantSandworm.ts
+++ b/packages/garbo/src/tasks/freeGiantSandworm.ts
@@ -24,6 +24,7 @@ import {
   AsdonMartin,
   clamp,
   get,
+  getModifier,
   have,
   set,
   sum,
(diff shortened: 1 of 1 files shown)


Rinn:
We can just put the price check on all of the tasks except for tom, I only put it on the quest so i wouldn't duplicate it but w/e.

packages/garbo/src/tasks/freeGiantSandworm.ts
```diff
@@ -293,12 +294,28 @@ function hasWorms(): boolean {
...
+const BASE_RATE = 1 / 100;
+let _sandwormRate: number;
+function sandwormRate(): number {
+  if (!_sandwormRate) {
+    Outfit.from(
+      sandwormSpec(),
+      new Error("Failed to generate Sandworm outfit"),
+    ).dress();
+    _sandwormRate =
+      REJECTION * BASE_RATE * (1 + getModifier("Item Drop") / 100);
```
Rinn, line 307: the issue here is the Item Drop modifier is inaccurate until we maximize and buff. actually are we even buffing with the common item drop effects?

horrible-little-slime: I don't think so but great question

horrible-little-slime: I think this isn't crazy accurate but the question is will this be _more_ accurate than what we're currently getting? it looks like we used to use `0.02 * price of melange` and now use `0.01`, which is causing #2138

Rinn: yeah it's fine we should just go with this for now

packages/garbo/src/tasks/freeGiantSandworm.ts
```diff
@@ -228,7 +238,9 @@ const SandwormTasks: GarboFreeFightTask[] = [
     },
     {
       name: "Yellow Ray",
-      ready: () => have($skill`Fondeluge`) || have($item`Jurassic Parka`),
+      ready: () =>
+        (drumMachineWorthIt() && have($skill`Fondeluge`)) ||
+        have($item`Jurassic Parka`),
```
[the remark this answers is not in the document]
Rinn: minor logic error here right?

Rinn (approved):

Issue #2121, opened 2024-09-19
[BUG] Innate wanderers are killed during free run fights instead of ran from


Shiverwarp:
**Describe the bug**
`customizeMacro ` will kill innate wanderers unless we're using backup camera. We don't want this during free run fights because it eats up our buff turns.

```ts
function customizeMacro<M extends StrictMacro>(macro: M) {
  return Macro.if_(
    $monsters`giant rubber spider, time-spinner prank`,
    Macro.kill(),
  )
    .externalIf(
      have($effect`Eldritch Attunement`),
      Macro.if_($monster`Eldritch Tentacle`, Macro.basicCombat()),
    )
    .ifInnateWanderer(
      Macro.externalIf(
        haveEquipped($item`backup camera`) &&
          get("_backUpUses") < 11 &&
          get("lastCopyableMonster") === globalOptions.target &&
          myFamiliar() === meatFamiliar(),
        Macro.skill($skill`Back-Up to your Last Enemy`).step(macro),
        Macro.basicCombat(),
      ),
    )
    .step(macro);
}
```

**To Reproduce**
Running free run fights during days with holiday wanderers (I ran into it on pirate day during my mayfly runs)

**Expected behavior**
We use our planned free run on the holiday wanderer during freeRunFights

PR #2140, opened 2024-09-28
slightly restructure GarboStrategy to allow customizing behavior with unexpected guests

horrible-little-slime:
This should allow us to handle #2121 more easily

### DIFF
--- a/packages/garbo/src/combat.ts
+++ b/packages/garbo/src/combat.ts
@@ -863,25 +863,42 @@ export class Macro extends StrictMacro {
   }
 }
 
-function customizeMacro<M extends StrictMacro>(macro: M) {
+type CustomizeMacroOptions = {
+  freeWanderer: (macro: StrictMacro) => Macro;
+  tentacle: (macro: StrictMacro) => Macro;
+  innateWanderer: (macro: StrictMacro) => Macro;
+};
+const DEFAULT_MACRO_OPTIONS = {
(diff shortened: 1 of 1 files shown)


packages/garbo/src/combat.ts
```diff
@@ -863,25 +863,42 @@ export class Macro extends StrictMacro {
...
 
-function customizeMacro<M extends StrictMacro>(macro: M) {
+type CustomizeMacroOptions = {
+  freeWanderer: (macro: StrictMacro) => Macro;
+  tentacle: (macro: StrictMacro) => Macro;
+  innateWanderer: (macro: StrictMacro) => Macro;
+};
+const DEFAULT_MACRO_OPTIONS = {
+  freeWanderer: () => Macro.kill(),
+  tentacle: () => Macro.basicCombat(),
```
Rinn, line 873: I don't remember why we use kill() vs basicCombat() for free wanderers, maybe they should both use basicCombat()?

Rinn (dismissed):

packages/garbo/src/combat.ts
horrible-little-slime: basicCombat includes stasising for hobo monkeys and involves deleveling; kill just kills. so actually it feels like if anything, basiccombat should be on the time prank or rubber spider. so eyah I agree, basic combat for both

PR #2148, opened 2024-10-03
Framework for Bat Wings

Ignose:
Requires:
1. Better understanding of free fight rate
2. Mafia tracking of free fights (limit 5)
3. Mafia tracking of bat skills
4. lint update


packages/garbo/src/fights.ts
```diff
@@ -1297,6 +1299,8 @@ const freeRunFightSources = [
           .trySkillRepeat(
             $skill`Fire Extinguisher: Polar Vortex`,
             $skill`Perpetrate Mild Evil`,
+            // eslint-disable-next-line libram/verify-constants
+            $skill`Swoop like a Bat`,
```
Rinn, line 1303: are we nearing the round limit with this?

Ignose: I don't know. We could be. The plan is to test it tomorrow and find out.

Ignose: To answer your question, it didn't swoop! So I still don't know. I think Libram/eslint need to be updated first to test that.

horrible-little-slime: No, you just never wrote code to equip the wings.

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -95,6 +97,9 @@ import { lavaDogsAccessible, lavaDogsComplete } from "../resources/doghouse";
 import { hotTubAvailable } from "../resources/clanVIP";
 import { meatMood } from "../mood";
 
+let batWingsTurns = 0;
+let batWingTriggers = 0;
```
horrible-little-slime, line 101: This doesn't belong here, it belongs in excavator

PR #2149, opened 2024-10-04
Outline of pizza cube maximizer

Ignose:


packages/garbo/src/resources/worksheds.ts
```diff
@@ -157,3 +163,62 @@ export function grabMedicine(): void {
...
+function calculateItemValue(item: Item, effect: Effect): number {
+  return (
+    new Potion($item`diabolic pizza`, {
+      effect: effect,
+      duration: Math.sqrt(autosellPrice(item)),
+    }).gross(copyTargetCount()) -
+    mallPrice(item) +
+    (toSlot(item) === $slot`familiar`
+      ? mallPrice($item`box of Familiar Jacks`)
+      : 0) +
```
horrible-little-slime, line 177: You don't get multiple jacks from having multiple familiar items unfortunately

packages/garbo/src/resources/worksheds.ts
```diff
@@ -157,3 +163,62 @@ export function grabMedicine(): void {
...
+
+// Simulation to calculate pizza cost and benefit
+export function simCreatePizza(effect: Effect): [number, number] {
+  const pizzaItems = designPizza(effect);
+
+  const benefit = pizzaItems.reduce(
+    (acc, it) => acc + calculateItemValue(it, effect),
+    0,
+  );
+  const cost = pizzaItems.reduce((acc, it) => acc + mallPrice(it), 0);
```
horrible-little-slime, line 210:
```suggestion
  const cost = sum(pizzaItems, mallPrice);
```

packages/garbo/src/resources/worksheds.ts
```diff
@@ -157,3 +163,62 @@ export function grabMedicine(): void {
...
+}
+
+// Simulation to calculate pizza cost and benefit
+export function simCreatePizza(effect: Effect): [number, number] {
+  const pizzaItems = designPizza(effect);
+
+  const benefit = pizzaItems.reduce(
+    (acc, it) => acc + calculateItemValue(it, effect),
+    0,
+  );
```
horrible-little-slime, line 209:
```suggestion
  const benefit = sum(pizzaItems, (it) => calculateItemValue(it, effect));
```

packages/garbo/src/resources/worksheds.ts
```diff
@@ -157,3 +163,62 @@ export function grabMedicine(): void {
...
+// Consolidated function to get the best item for a given letter
+function bestPizzaItemForLetter(letter: string, effect: Effect): Item {
+  const items = $items.all().filter((i) => i.name.startsWith(letter));
+
+  const itemsWithValues = items.map((item) => ({
+    item,
+    value: calculateItemValue(item, effect),
+  }));
+
+  return maxBy(itemsWithValues, (entry) => entry.value).item;
```
horrible-little-slime, line 191:
```suggestion
  return maxBy(itemsWithValues, "value").item;
```

horrible-little-slime: Also fwiw if you aren't using the values outside of the maxby it makes more sense to not do the map at all and to instead just ```typescript
return maxBy(items, (it) => calculateItemValue(it, effect));
```

PR #2163, opened 2024-10-28
Add Automatic Pirate Realm to Trash Island to support Cockroaches

Ignose:
Still in draft stage, tasks are broken up into chunks in case we want to do them at weird times given rollover buffs.

### DIFF
--- a/packages/garbo/src/resources/piraterealm.ts
+++ b/packages/garbo/src/resources/piraterealm.ts
@@ -1,6 +1,14 @@
 import { Outfit, step } from "grimoire-kolmafia";
-import { $item, $location, $monster, get, have } from "libram";
-import { runChoice, visitUrl } from "kolmafia";
+import { $item, $location, $monster, $stat, get, have, uneffect } from "libram";
+import {
+  Effect,
+  myBuffedstat,
+  myEffects,
+  numericModifier,
+  runChoice,
+  toEffect,
+  visitUrl,
+} from "kolmafia";
 import { GarboTask } from "../tasks/engine";
 
 import { GarboStrategy, Macro } from "../combat";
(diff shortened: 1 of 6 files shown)


packages/garbo/src/resources/piraterealm.ts
```diff
@@ -0,0 +1,174 @@
...
+        // Check if the effect modifier includes the stat and not "meat"
+        if (
+          numericModifier(ef, "muscle") &&
+          !(numericModifier(ef, "meat drop") > 0)
+        ) {
+          uneffect(ef); // Remove the effect
+        }
+      });
+    }
+  });
```
horrible-little-slime, line 39: use for loops, not forEach

packages/garbo/src/resources/piraterealm.ts
```diff
@@ -0,0 +1,174 @@
...
+  );
+
+  stats.forEach((stat) => {
+    if (myBuffedstat(stat) > 100) {
+      // Get effect names from myEffects and convert them to Effect instances
+      effects.forEach((ef) => {
+        // Check if the effect modifier includes the stat and not "meat"
+        if (
+          numericModifier(ef, "muscle") &&
+          !(numericModifier(ef, "meat drop") > 0)
```
horrible-little-slime, line 33: use libram's `getModifier`
and probably `stat.name` rather than just "muscle"

packages/garbo/src/resources/piraterealm.ts
```diff
@@ -0,0 +1,174 @@
...
+        ) {
+          uneffect(ef); // Remove the effect
```
horrible-little-slime, line 35: prioritize shruggable effects imo

packages/garbo/src/resources/piraterealm.ts
horrible-little-slime: check for familiar weight and probably item drop as well

packages/garbo/src/resources/piraterealm.ts
horrible-little-slime: that way you can `break` the loop (or `return` out of the function) when you get below the stat threshold

packages/garbo/src/resources/piraterealm.ts
```diff
@@ -0,0 +1,174 @@
...
+  return [runGiantGiantCrab(), selectTrashIsland()];
+}
+
+export function getEyepatch(): GarboTask {
+  return {
+    name: "Start Pirate Realm",
+    completed: () => have(eyepatch),
+    ready: () =>
+      get("prAlways") ||
+      (get("_prToday") &&
```
horrible-little-slime, line 61: use `realmAvailable`

packages/garbo/src/resources/piraterealm.ts
horrible-little-slime, line 1: Given that this is just a file full of tasks, it should go in `/tasks/` rather than `/resources/`; you can slice out some of the logic used here into functions and keep those functions in resources, though. Like getting your PR crew, and stuff

Ignose:
Closing as I PR the good stuff to Shiver's PR and let the bad stuff burn

PR #2165, opened 2024-10-31
Add Quantum Familiar as a new constant value fam

Ignose:
Needs lint/libram updates first


horrible-little-slime:
You did not, in fact, add this as a constant value familiar. but that's where it should be

horrible-little-slime:
eslint-plugin-libram has been bumped

PR #2166, opened 2024-11-02
Push some improvements and fixes to make this more accessible.

Ignose:


horrible-little-slime:
Everything I said in #2163 still stands tbh

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +15,91 @@ import {
...
-    if (have($effect`Gummiskin`)) uneffect($effect`Gummiskin`);
-  }
-  if (
-    myBuffedstat($stat`Moxie`) >= 100 ||
-    myBuffedstat($stat`Mysticality`) >= 100 ||
-    myBuffedstat($stat`Muscle`) >= 100
-  ) {
-    uneffect($effect`Having a Ball!`);
-  }
+  const stats = [$stat`Muscle`, $stat`Moxie`, $stat`Mysticality`];
```
horrible-little-slime, line 37: `Stat.all()` exists

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +15,91 @@ import {
...
+  const effects: Effect[] = Object.keys(myEffects()).map((effectName) =>
+    toEffect(effectName),
+  );
+
+  stats.forEach((stat) => {
```
horrible-little-slime, line 42: switch from a `forEach` to a `for`

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +15,91 @@ import {
...
+    while (myBuffedstat(stat) > 100) {
+      if (
+        !have($effect`Mush-Mouth`) &&
+        mallPrice($item`Fun-Guy spore`) < 5_000
```
horrible-little-slime, line 46: don't use an arbitrary price cap, use a non-arbitrary price cap

packages/garbo/src/tasks/cockroachPrep.ts
horrible-little-slime: that way you can `break` the loop

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +15,91 @@ import {
...
+      }
+
+      if (mallPrice($item`Mr. Mediocrebar`) < 2_000 && !have($effect`Apathy`)) {
+        retrieveItem($item`Mr. Mediocrebar`);
+        use($item`Mr. Mediocrebar`);
+      }
+
+      if (have($effect`Feeling Excited`)) uneffect($effect`Feeling Excited`);
+      // Get effect names from myEffects and convert them to Effect instances
+      effects.forEach((ef) => {
```
horrible-little-slime, line 86: and this especially should be a `for` rather than a `forEach`

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +15,91 @@ import {
...
+        // Check if the effect modifier includes the stat and not "meat"
+        if (
+          numericModifier(ef, `${stat.toString}`) &&
```
horrible-little-slime, line 89: [garf]

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +15,91 @@ import {
...
+          !(
+            numericModifier(ef, "meat drop") > 0 ||
```
horrible-little-slime, line 91: use `getModifier` rather than `numericModifier` here; compare them to 0 rather than checking if they're falsy, because negative famweight effects exist.. And make them an array that you test with `some` or `every`

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -99,7 +154,9 @@ function chooseCrew(): void {
 
 export const CockroachSetup: Quest<GarboTask> = {
   name: "Setup Cockroach Target",
-  completed: () => get("_lastPirateRealmIsland") === $location`Trash Island`,
+  completed: () =>
+    get("_lastPirateRealmIsland") === $location`Trash Island` ||
+    !get("pirateRealmUnlockedAnemometer"),
```
horrible-little-slime, line 159: I would probably make the anemometer a `ready` rather than a `completed`

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -135,7 +196,9 @@ export const CockroachSetup: Quest<GarboTask> = {
       completed: () => questStep("_questPirateRealm") > 1,
       prepare: () => checkAndFixOvercapStats(),
       do: () => adv1($location`Sailing the PirateRealm Seas`),
-      outfit: { equip: $items`PirateRealm eyepatch` },
+      outfit: (): Outfit => {
+        return freeFightOutfit({ acc3: $items`PirateRealm eyepatch` });
+      },
```
horrible-little-slime, line 201:
```suggestion
      outfit: () => freeFightOutfit({ acc3: $items`PirateRealm eyepatch` }),
```

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -209,7 +272,7 @@ export const CockroachSetup: Quest<GarboTask> = {
       do: () => adv1(get("_lastPirateRealmIsland", $location`none`)),
       outfit: () =>
         freeFightOutfit({
-          equip: $items`PirateRealm eyepatch, PirateRealm party hat, carnivorous potted plant, Red Roger's red left foot, Space Trip safety headphones`,
+          equip: $items`PirateRealm eyepatch`,
```
horrible-little-slime, line 275: assign bonuses to the FUN equipments probably? and the carnplant for sure.

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -253,12 +316,10 @@ export const CockroachSetup: Quest<GarboTask> = {
       },
       outfit: () => {
         if (get("_lastPirateRealmIsland") === $location`Crab Island`) {
-          const spec: OutfitSpec = {
+          const spec = meatTargetOutfit({
             modifier: ["meat"],
             equip: $items`PirateRealm eyepatch`,
-            avoid: $items`cursed pirate cutlass`, // Gives +25 muscle which often overcaps
```
horrible-little-slime, line 259: seems like you're cutting out helpful and load-bearing lines without replacing them with anything

Ignose: Helpful maybe, but hard coding them is inappropriate in this case. While we should value the red roger equips, avoiding the cursed pirate cutlass is strictly unnecessary when we're managing buffs more effectively.

packages/garbo/src/tasks/cockroachPrep.ts
Ignose: I'm not sure the relative value of each of these or even how to determine that, given all they do for us is... reduce stats.

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +15,84 @@ import {
...
+      return effect && !(effect in myEffects());
+    })
+    .filter((it) => getModifier(statName, it) < 0)
+    .map(
+      (it) => [it, mallPrice(it) / getModifier(statName, it)] as [Item, number],
+    ); // getModifier should return a negative value, flipping these negative
+
+  // ...so that when we maxBy we pick the smallest negative value, spending the least meat
+  return maxBy(debuffMenu, ([, value]) => value)[0];
+}
```
horrible-little-slime, line 53: function getBestDebuffItem(stat: Stat): Item {
  const debuffs = Item.all()
    .map((item) => ({ item, effect: effectModifier(it, "Effect") })
    .filter(({ effect }) => effect !== $effect.none && !have(effect) && getModifier(stat.toString(), effect) < 0);
    
  return maxBy(debuffs, ({ item, effect }) => mallPrice(item) / getModifier(stat.toString(), effect)).item;
}

This still has the issue of ignoring the various `Percent` buffs, but I suppose those are unlikely to matter much when wearing the eyepatch

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +15,84 @@ import {
...
-    uneffect($effect`Having a Ball!`);
+  const stats = Stat.all();
+  const effects: Effect[] = Object.keys(myEffects()).map((effectName) =>
+    toEffect(effectName),
+  );
+
+  // Use a traditional for loop for stats
+  for (let i = 0; i < stats.length; i++) {
+    const stat = stats[i];
+    const statName = stat.toString();
```
horrible-little-slime, line 65:
```suggestion
  for (const stat of Stat.all()) {
    const stat = stats[i];
    const statName = stat.toString();
```

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +15,84 @@ import {
...
-  if (myBuffedstat($stat`Moxie`) >= 100) {
-    if (have($effect`Gummiskin`)) uneffect($effect`Gummiskin`);
-  }
-  if (
-    myBuffedstat($stat`Moxie`) >= 100 ||
-    myBuffedstat($stat`Mysticality`) >= 100 ||
-    myBuffedstat($stat`Muscle`) >= 100
-  ) {
-    uneffect($effect`Having a Ball!`);
+  const stats = Stat.all();
```
horrible-little-slime, line 57:
```suggestion
```

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +15,84 @@ import {
...
+  );
+
+  // Use a traditional for loop for stats
+  for (let i = 0; i < stats.length; i++) {
+    const stat = stats[i];
+    const statName = stat.toString();
+
+    while (myBuffedstat(stat) > 100) {
+      for (const isShruggablePass of [true, false]) {
+        for (let j = 0; j < effects.length; j++) {
```
horrible-little-slime, line 69: We absolutely don't need to use a nested loop here, we can achieve this much more efficiently by just handling things consecutively

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -97,8 +145,18 @@ function chooseCrew(): void {
...
 }
 
+const funPointValue = garboValue($item`PirateRealm guest pass`) / 600;
+const carnPlantValue =
+  get("valueOfAdventure") / (20 + get("_carnivorousPottedPlantWins"));
+const funPointBonuses = new Map<Item, number>([
+  [$item`Red Roger's red left foot`, funPointValue],
+  [$item`PirateRealm party hat`, funPointValue],
+  [$item`carnivorous potted plant`, carnPlantValue],
+]);
```
horrible-little-slime, line 155:
```suggestion
function outfitBonuses() {
  const funPointValue = garboValue($item`PirateRealm guest pass`) / 600;
  return new Map([
    [$item`carnivorous potted plant`,   get("valueOfAdventure") / (20 + get("_carnivorousPottedPlantWins"))],
    [$item`Red Roger's red left foot`, funPointValue],
    [$item`PirateRealm party hat`, funPointValue],
  ]);
}
```

we don't want mallsearches in top level scope (because they'll lead to mallsearches before we do `mallPrices("allitems")` and we don't want carn plant in top level scope (because we don't want it to stay fixed at the initial value

packages/garbo/src/tasks/cockroachPrep.ts
Ignose: Marking as resolved because we just choose the cheapest thing, now.

packages/garbo/src/tasks/cockroachPrep.ts
Ignose: I think we test it and if perfect buffs matter, reassess.

packages/garbo/src/index.ts
```diff
@@ -569,6 +569,7 @@ export function main(argString = ""): void {
 
         // 2. do some target copy stuff
         freeFights();
+        runGarboQuests([CockroachFinish]);
         runGarboQuests([SetupTargetCopyQuest]);
```
horrible-little-slime, line 573:
```suggestion
        runGarboQuests([CockroachFinish, SetupTargetCopyQuest]);
```

packages/garbo/src/tasks/cockroachPrep.ts
```diff
@@ -16,35 +16,83 @@ import {
...
+      ({ effect }) =>
+        effect !== $effect.none &&
+        !have(effect) &&
+        getModifier(stat.toString(), effect) < 0,
+    );
+
+  return maxBy(
+    debuffs,
+    ({ item, effect }) =>
+      mallPrice(item) / getModifier(stat.toString(), effect),
```
horrible-little-slime, line 53:
```
> js Item.all().filter((i) => effectModifier(i, "Effect") !== Effect.none && mallPrice(i) < 0)

Returned: aggregate item [4]
0 => Bee's Knees
1 => Sockdollager
2 => Ish Kabibble
3 => Hot Socks
```
[garf]

Is the easiest fix to this to check for item type potion?

Ignose: We don't want to diet here at all, so a filter for item.potion is a good idea. Updated.

PR #2172, opened 2024-11-22
fix ready/completed check for portscan macro mushroom

### LINKED ISSUE #2171: [BUG] PORTSCAN causes Government agent to be present in Shadow Rift when trying to acquire Shadow Waters
**Describe the bug**
During execution of my 2nd leg of a Community Service loop, Garbo fails to acquire Shadow Waters from the Shadow Rift because a Government agent appears before the non-combat. `PORTSCAN` seems to be used during `Free Fight/Mushroom garden`.

**To Reproduce**
Unsure, it's repeatable when using my CS loop wrapped `bLooper`, and only occurs after running the Community Service path in my loop.

**Expected behavior**
Acquire Shadow Waters

**Session Log**
[The_Powers_That_Be_20241120.txt](https://github.com/user-attachments/files/17837601/The_Powers_That_Be_20241120.txt)

Task `Free Fight/Mushroom garden` starts on line 92858.
You can see `PORTSCAN` being used on line 92885.
Then you can see the Government agent fight that should be the Like a Loded Stone NC on line 93581.

**Additional context**
This always happens post-loop, but never pre-loop. Has been occurring for several weeks, I usually just manually adventure in the rift to get Shadow Waters then re-start my wrapper.

horrible-little-slime:
resolves #2171

### DIFF
--- a/packages/garbo/src/tasks/freeFight.ts
+++ b/packages/garbo/src/tasks/freeFight.ts
@@ -558,10 +558,9 @@ const FreeFightTasks: GarboFreeFightTask[] = [
       (have($item`packet of mushroom spores`) ||
         getCampground()["packet of mushroom spores"] !== undefined) &&
       !doingGregFight() &&
-      Counter.get("portscan.edu") === 0 &&
       have($skill`Macrometeorite`) &&
       get("_macrometeoriteUses") < 10,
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
fuck it we ball

PR #2177, opened 2024-12-10
autosell autosellable melting junk before RO

horrible-little-slime:


gausie (commented):
Just to stop this annoying someone in future, I would add a filter for things that provide a rollover effect too

gausie (approved):

PR #2187, opened 2024-12-23
Handle Autumnaton through garbo-lib

horrible-little-slime:


pstalcup (dismissed):

packages/garbo/src/resources/autumnaton.ts
gausie, line 1: I would potentially keep this file and export the AutumnatonManager instance from here

PR #2192, opened 2025-01-04
feat: stop hiding target flag

midgleyc:


docrostov (approved):
conversation has dried up in #garbo-dev and i would like to have this in the readme ASAP. now we can move on to greener pastures, the reward for those who update the readme on public volunteer projects. the windows 95 screensaver awaits us, ryo, out into the wide green expanse of a better world. i will shed a tear for those of us who never get to see the wonder and glory of --[someone shoots me with a gun]

PR #2201, opened 2025-01-08
moving help above mallprices & main.php check

docrostov:
as discussed in #garbo-dev

### DIFF
--- a/packages/garbo/src/index.ts
+++ b/packages/garbo/src/index.ts
@@ -120,6 +120,13 @@ export function main(argString = ""): void {
   sinceKolmafiaRevision(28151); // detect TakerSpace + basic related functionality
   checkGithubVersion();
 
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
Reading through this I can see why we haven't done it before:
- it's nice, but not strictly required, to set our `globalOptions.target`
immediately after the `Args.fill`
- it _is_ basically required to do our mallsearches before deciding our
`globalOptions.target`
- we need to do Args.fill before we call help iff we're checking the
globalOptions object for help instead of doing a simple atring match

So there's a couple of all equally fine options to choose from for how to
reconcile this:
- move up the Args.fill, make sure that doesn't break anything
- don't ask the args object whether to display help, just check if the
argstring matches

Either of those should work

horrible-little-slime (changes requested):
Changes detailed in previous comment

PR #2207, opened 2025-01-09
Parachute

Ignose:


horrible-little-slime:
I think this is basically ready but I would like infinite loop protection for the unlikely-but-plausible situation that `myLocation` is set to the wrong value (I think we even `setLocation` to barf occasionally so it'll actually matter

horrible-little-slime:
probably the easiest way is to like
```typescript
let lastParachuteFailure = 0;
// snip
  ready: () => CrepeParachute.have() && myLocation() === $location`barf mountain` && lastParachuteFailure !== totalTurnsPlayed(),
  do: () => {
    const result = CrepeParachute.fight($monster`garbage tourist`);
    if (!result) lastParachuteFailure = totalTurnsPlayed();
  }
```
or something along those lines

pstalcup (approved):

PR #2215, opened 2025-01-14
First pass CyberRealm

Shiverwarp:
Simplistic version just to use the free fights for now

### DIFF
--- a/packages/garbo/src/tasks/freeFight.ts
+++ b/packages/garbo/src/tasks/freeFight.ts
@@ -805,14 +805,14 @@ const FreeFightTasks: GarboFreeFightTask[] = [
   {
     name: "CyberRealm Overclock Fights",
     ready: () =>
-      have($item`server room key`) && // TODO support day pass? Not require tanktop
+      canAdventure($location`Cyberzone 1`) &&
       have($item`zero-trust tanktop`) &&
       have($skill`Torso Awareness`) &&
       have($skill`OVERCLOCK(10)`),
     completed: () => get("_cyberFreeFights") >= 10,
     do: $location`Cyberzone 1`, // TODO Support other zones with better equipment and valuing hacker drops
     tentacle: false,
-    choices: () => ({ 1545: 1 }), // Take damage, get 0's
+    choices: { 1545: 1 }, // Take damage, get 0's
     outfit: () =>
       freeFightOutfit({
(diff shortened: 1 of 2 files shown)


packages/garbo/src/tasks/freeFight.ts
```diff
@@ -801,6 +802,34 @@ const FreeFightTasks: GarboFreeFightTask[] = [
   // tied-up leaviathan (scaling, has 100 damage source cap and 2500 hp)
   // li'l ninja costume
   // closed-circuit pay phone (make into it's own Quest)
+  {
+    name: "CyberRealm Overclock Fights",
+    ready: () =>
+      have($item`server room key`) && // TODO support day pass? Not require tanktop
```
horrible-little-slime, line 808: To support the day pass, easiest way would be to use `canAdventure`, right?

packages/garbo/src/tasks/freeFight.ts
```diff
@@ -801,6 +802,34 @@ const FreeFightTasks: GarboFreeFightTask[] = [
...
+      have($item`zero-trust tanktop`) &&
+      have($skill`Torso Awareness`) &&
```
horrible-little-slime, line 810: I know some (unsupported) avatar paths have shirt access through another skill. Would it make more sense to use `canEquip`? ("No" is a perfectly fine answer here, I'm just spintballing)

packages/garbo/src/tasks/freeFight.ts
```diff
@@ -801,6 +802,34 @@ const FreeFightTasks: GarboFreeFightTask[] = [
...
+      have($skill`OVERCLOCK(10)`),
+    completed: () => get("_cyberFreeFights") >= 10,
+    do: $location`Cyberzone 1`, // TODO Support other zones with better equipment and valuing hacker drops
+    tentacle: false,
+    choices: () => ({ 1545: 1 }), // Take damage, get 0's
```
horrible-little-slime, line 815:
```suggestion
    choices: { 1545: 1 }, // Take damage, get 0's
```
doesn't matter too much, but if nothing ever changes it's nicer to not make it a function imo

horrible-little-slime (changes requested):
A few quick comments, some of which need action and most of which don't

PR #2223, opened 2025-01-29
Deviled candy

Shiverwarp:
I have a cached value here so that we aren't running the calculation every time to check the completion, but we still need to be able to get different candies each re-run, so we also use the non-cached version

### DIFF
--- a/packages/garbo/src/tasks/dailyItems.ts
+++ b/packages/garbo/src/tasks/dailyItems.ts
@@ -165,13 +165,18 @@ function bestDevilerCandy(): Item {
   const sortedInventoryCandies = inventoryCandies
     .sort((a, b) => garboValue(a) - garboValue(b))
     .slice(0, 50);
+  // remove any candies that cost more than the lowest priced one in the list
   const prunedInventoryCandies = sortedInventoryCandies.filter(
     (i) => garboValue(i) === garboValue(sortedInventoryCandies[0]),
   );
   // Lowest garbovalue candy that we have the largest amount of
   const bestInventoryCandy = maxBy(prunedInventoryCandies, itemAmount);
 
-  return maxBy([bestInventoryCandy, bestCandyFromMall], garboValue, true);
+  return maxBy(
+    [bestInventoryCandy, bestCandyFromMall],
+    (i) => (i === bestInventoryCandy ? garboValue(i) : mallPrice(i)),
+    true,
+  );
 }
 
 let cachedbestDevilerCandy: Item | null = null;
--- a/packages/garbo/src/tasks/dailyItems.ts
+++ b/packages/garbo/src/tasks/dailyItems.ts
@@ -732,12 +732,16 @@ const DailyItemTasks: GarboTask[] = [
         garboValue($item`deviled candy egg`),
         true,
       );
+      print(
+        `${getBestDevilerCandy()} will be deviled for expected profit of ${garboValue($item`deviled candy egg`) - garboValue(getBestDevilerCandy())}`,
+      );
       visitUrl(`inventory.php?action=eggdevil&pwd`);
       visitUrl(
         `choice.php?a=${toInt(getBestDevilerCandy())}&whichchoice=1544&option=1&pwd`,
       );
       cachedbestDevilerCandy = null;
     },
+    limit: { skip: 3 },
     spendsTurn: false,
   },
 ];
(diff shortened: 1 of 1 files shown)


packages/garbo/src/tasks/dailyItems.ts
```diff
@@ -138,6 +140,62 @@ function pickCargoPocket(): void {
   }
 }
 
+function bestDevilerCandy(): Item {
+  // These are replenishable notrade nodiscard items that we would prefer to use
+  const priorityUntradeableNoDiscardList =
+    $items`sugar shotgun, sugar shillelagh, sugar shank, sugar chapeau, sugar shorts, sugar shield, sugar shirt`.filter(
+      (i) => itemAmount(i) > 1,
```
horrible-little-slime, line 147: is there a reason we aren't just using `have` here?

packages/garbo/src/tasks/dailyItems.ts
```diff
@@ -138,6 +140,62 @@ function pickCargoPocket(): void {
...
+    );
+  const bestPriorityCandy =
+    priorityUntradeableNoDiscardList.length > 0
+      ? maxBy(priorityUntradeableNoDiscardList, (i) => itemAmount(i))
+      : null;
+
+  if (bestPriorityCandy) return bestPriorityCandy;
```
horrible-little-slime, line 154:
```suggestion
  if (priorityUntradeableNoDiscardList.length > 0) {
    return maxBy(priorityUntradeableNoDiscardList, itemAmount);
  }
```

packages/garbo/src/tasks/dailyItems.ts
```diff
@@ -138,6 +140,62 @@ function pickCargoPocket(): void {
...
+      ? maxBy(prunedInventoryCandies, itemAmount)
+      : null;
+
+  return bestInventoryCandy
+    ? maxBy(
+        [bestInventoryCandy, bestCandyFromMall],
+        (i) => (i === bestInventoryCandy ? garboValue(i) : mallPrice(i)),
+        true,
+      )
+    : bestCandyFromMall;
```
horrible-little-slime, line 188:
```suggestion
  const bestInventoryCandy = () => {
    if (inventoryCandies.length === 0) return null;
    const naiveBest = maxBy(inventoryCandies, garboValue, true);
    return maxBy(inventoryCandies.filter((c) => garboValue(c) === naiveBest(c)), itemAmount);
  }();
  
  return maxBy([bestInventoryCandy, bestCandyFromMall].filter(notNull), mallPrice, true);
```
I think this is smoother?

packages/garbo/src/tasks/dailyItems.ts
```diff
@@ -661,6 +719,31 @@ const DailyItemTasks: GarboTask[] = [
...
+        garboValue($item`deviled candy egg`),
+        true,
+      );
+      print(
+        `${getBestDevilerCandy()} will be deviled for expected profit of ${garboValue($item`deviled candy egg`) - garboValue(getBestDevilerCandy())}`,
+      );
+      visitUrl(`inventory.php?action=eggdevil&pwd`);
+      visitUrl(
+        `choice.php?a=${toInt(getBestDevilerCandy())}&whichchoice=1544&option=1&pwd`,
+      );
```
horrible-little-slime, line 741:
```suggestion
      runChoicel(1, `a=${toInt(getBestDevilerCandy())}`);
```

horrible-little-slime (changes requested):

packages/garbo/src/tasks/dailyItems.ts
horrible-little-slime: I think this is a nicer pattern given we're already doing an early return, you know?

packages/garbo/src/tasks/dailyItems.ts
horrible-little-slime: I'm not confident there's value in slicing the array if we aren't caching the array itself
Also, the pattern I use to define `bestInventoryCandy` here is what's called a [closure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Closures). The only utility we gain from using a closure here is being able to use the early return instead of the ternary, so I'd be willing to not use it. But I feel like it reads nicer this way?

packages/garbo/src/tasks/dailyItems.ts
horrible-little-slime: we can use `have(i, 2)` but I cop to not having noticed that it's > 1 instead of > 0, this is good

horrible-little-slime (approved):

PR #2224, opened 2025-01-30
Liana Parachute

Shiverwarp:


packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -955,6 +957,38 @@ const BarfTurnTasks: GarboTask[] = [
...
   },
+  {
+    name: "Liana Parachute",
+    ready: () =>
+      (sober() ||
+        (have($item`Drunkula's wineglass`) &&
+          canEquip($item`Drunkula's wineglass`))) &&
+      CrepeParachute.have() &&
+      shouldCheckParachute() &&
+      questStep("questL11Worship") > 3 &&
```
horrible-little-slime, line 968: I can see some flaws with this check but honestly I think the guy who encounters that problem deserves it

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -955,6 +957,38 @@ const BarfTurnTasks: GarboTask[] = [
...
+      Macro.abortWithMsg(
+        "Did not instantly kill the Liana, check what went wrong",
+      ),
+    ),
+    prepare: () => {
+      if (!sober()) {
+        freeFightOutfit({ offhand: $item`Drunkula's wineglass` }).dress();
+      }
+      withChoice(785, 6, () =>
+        adv1($location`An Overgrown Shrine (Northeast)`, -1, ""),
```
horrible-little-slime, line 983: long-term, unfortunately, I think we'll have to split this prepare into a separate task to handle holiday wanderers properly. But again, short-term, I'm just going to merge this I think

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -955,6 +957,38 @@ const BarfTurnTasks: GarboTask[] = [
...
+      );
+      if (!sober()) freeFightOutfit({ weapon: $item`antique machete` }).dress();
```
horrible-little-slime, line 985: I don't _love_ having to do this, but here we are.

PR #2238, opened 2025-02-07
add tracking for which familiars work on free run

horrible-little-slime:
We should test this for a couple of days before merging but also I think this is phase 0 of TCB support

### DIFF
--- a/packages/garbo/src/familiar/freeFightFamiliar.ts
+++ b/packages/garbo/src/familiar/freeFightFamiliar.ts
@@ -21,7 +21,12 @@ import { garboValue } from "../garboValue";
 import getConstantValueFamiliars from "./constantValueFamiliars";
 import getDropFamiliars from "./dropFamiliars";
 import getExperienceFamiliars from "./experienceFamiliars";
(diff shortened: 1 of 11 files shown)


packages/garbo/src/familiar/freeFightFamiliar.ts
```diff
@@ -63,6 +68,7 @@ export function menu(
           (Math.max(familiarWeight($familiar`Grey Goose`) - 5), 0) ** 4,
         leprechaunMultiplier: 0,
         limit: "experience",
+        worksOnFreeRun: true,
```
Ignose, line 71: Should this be true? Don't we lose meatify matter meat on free run?
Edit: This is droning. Droning on free run is lit AF. nevermind.

Ignose (dismissed):

packages/garbo/src/familiar/freeFightFamiliar.ts
horrible-little-slime: No this is pretty clearly for meatify and is indeed a mistake. Look at the ol' `timeToMeatify()` on top

PR #2245, opened 2025-02-10
Acquire Best Free Run in Case of Abort

Shiverwarp:


packages/garbo/src/tasks/post/index.ts
```diff
@@ -339,11 +342,27 @@ function handleDrenchedInLava(): GarboPostTask {
...
 }
 
+function acquireAbortFreeRun(): GarboPostTask {
+  const possibleFreeRuns = unlimitedFreeRunList.filter(
+    (i) => getAcquirePrice(i) < get("valueOfAdventure"),
+  );
+  const bestFreeRun =
+    possibleFreeRuns.length > 0
+      ? maxBy(possibleFreeRuns, getAcquirePrice, true)
+      : $item`none`;
```
horrible-little-slime, line 352: Imo rather than filtering and then doing maxby, do the maxby and then do the comparison, like this:
```suggestion
  const bestFreeRun = maxBy(unlimitedFreeRunList, getAcquirePrice, true);
  return (getAcquirePrice(bestFreeRun) < get("valueOfAdventure") ? bestFreeRun : $item.none;
```

packages/garbo/src/tasks/post/index.ts
```diff
@@ -339,11 +342,27 @@ function handleDrenchedInLava(): GarboPostTask {
   };
 }
 
+function acquireAbortFreeRun(): GarboPostTask {
```
horrible-little-slime, line 345: I don't love the way this function is designed. I think it would make more sense to have the `bestFreeRun` be like a cached value returned by a function. I guess it overall makes sense to have it in post, but something about the calculation logic here being done in the task-making function feels wrong to me for reasons I can't articulate.

packages/garbo/src/tasks/post/index.ts
```diff
@@ -339,11 +343,34 @@ function handleDrenchedInLava(): GarboPostTask {
...
+  return bestAbortFreeRun;
+}
+
+function acquireAbortFreeRun(): GarboPostTask {
+  return {
+    name: "Acquire Best Free Run in Case of Abort",
+    completed: () =>
+      getBestAbortFreeRun() === $item`none` || have(getBestAbortFreeRun()),
+    do: () =>
+      acquire(1, getBestAbortFreeRun(), getAcquirePrice(getBestAbortFreeRun())),
```
spaghetti-squash, line 364: pass in value of adventure as the third arg and make sure you pass the "pwease don't abort" arg too

after that we're good

spaghetti-squash (approved):

PR #2259, opened 2025-02-20
Value Toy Cupid Bow for sandworms

Ignose:
The values need work, but the layout should be correct.

As long as you're at <5000% item find, TCB is effectively an item% doubler.


packages/garbo/src/familiar/itemFamiliar.ts
```diff
@@ -37,7 +37,11 @@ export function bestFairy(): Familiar {
     );
 
     const highestFairyMult = findFairyMultiplier(
-      maxBy(viableFairies, findFairyMultiplier),
+      maxBy(viableFairies, (f) =>
+        f === $familiar`Jill-of-All-Trades` && have($item`toy Cupid bow`)
```
spaghetti-squash, line 41: is TCB really more valuable than the boost from using jill over a hounddog?

PR #2264, opened 2025-02-22
bump packages

spaghetti-squash:


spaghetti-squash:
Actually we need one more grimoire version imo, because of https://github.com/loathers/grimoire/pull/128

PR #2284, opened 2025-03-13
call `effectValue` less while condomaxxing

spaghetti-squash:


midgleyc (approved):
much faster!

so much faster I failed to accurately time the improvement

went from > 15 minutes to less than a minute.

PR #2288, opened 2025-03-15
Do mayam during barf if temple reset

Shiverwarp:


spaghetti-squash (changes requested):
Can we make `mayamCalendarSummons` a task-producing function instead of a task object, and call it in the two places we put it? Reusing task objects gives me the heeby jeebies for reasons that are probably no longer relevant but still feels Right imo

spaghetti-squash (approved):

spaghetti-squash:
When we more fully grimoirize we can make the nose a free run fight and put this in that section but for now this is good

PR #2290, opened 2025-03-15
Swap to using oreo

Shiverwarp:
Welcoming input on where we actually place this - it feels slightly strange to include it within this function, but it will always be the best place to run it as far as I know.

### DIFF
--- a/packages/garbo/src/familiar/constantValueFamiliars.ts
+++ b/packages/garbo/src/familiar/constantValueFamiliars.ts
@@ -1,4 +1,4 @@
-import { Familiar, familiarWeight, holiday, squareRoot } from "kolmafia";
+import { Familiar, holiday, squareRoot } from "kolmafia";
 import {
   $effect,
   $familiar,
@@ -125,7 +125,7 @@ const standardFamiliars: ConstantValueFamiliar[] = [
 ];
 
 function peaceTurkeyDropChance(): number {
-  return 0.24 + squareRoot(familiarWeight($familiar`Peace Turkey`)) / 100;
+  return 0.24 + squareRoot(totalFamiliarWeight($familiar`Peace Turkey`)) / 100;
 }
 
 export default function getConstantValueFamiliars(
--- a/packages/garbo/src/target/fights.ts
+++ b/packages/garbo/src/target/fights.ts
@@ -15,7 +15,6 @@ import {
   retrieveItem,
   runChoice,
   runCombat,
-  toInt,
   use,
   userConfirm,
   useSkill,
@@ -39,6 +38,7 @@ import {
   CombatLoversLocket,
   Counter,
   CrystalBall,
+  directlyUse,
   get,
   getBanishedMonsters,
   have,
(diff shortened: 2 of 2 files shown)


spaghetti-squash:
Is there a reason this isn't just a grimoire task?

spaghetti-squash:
Right, but, like, why is it in mood?

gausie:
It seems like you want a grimoire task that basically runs it whenever
there are any free mines available. That way it is resilient to a future
item that gives free mines during the day or something.

spaghetti-squash:
You can put it in `post`, but that's sort of goofy probably

spaghetti-squash:
I'm very happy to switch to oreo; I'm really not interested in having our volcano mining code occur in `mood`. We can further discuss where this can go to behave optimally, but we absolutely shouldn't have something that changes your outfit and runs turns* in mood

*I feel like this counts as running turns even though it's oreo 0

spaghetti-squash (approved):
Approving but not personally merging. Shiver you can if you've tested it.

spaghetti-squash (approved):

packages/garbo/src/potions.ts
```diff
@@ -999,9 +999,15 @@ export function effectValue(
...
 ): number {
-  return new Potion($item.none, { duration, effect }).gross(
-    copyTargetCount(),
-    maxTurns,
-  );
+  if (effect === $effect`Shadow Affinity`) {
+    return globalOptions.prefs.valueOfFreeFight * duration; // Each turn of Shadow Affinity gives us one free fight
+  }
+
+  if (effect === $effect`Loded`) {
```
spaghetti-squash, line 1008: Only if volcano is available

packages/garbo/src/resources/scepter.ts
```diff
@@ -61,10 +61,12 @@ const SKILL_OPTIONS: ScepterSkill[] = [
...
     value: () =>
-      new Potion($item`august scepter`, {
-        effect: $effect`Incredibly Well Lit`,
-        duration: 30,
-      }).gross(copyTargetCount()), // TODO: Yachtzee
+      effectValue(
+        $effect`Incredibly Well Lit`,
+        30,
+        undefined,
+        copyTargetCount(),
```
spaghetti-squash, line 68: This is an optional parameter sk you can omit these last two

spaghetti-squash (commented):
Lets get this milk

spaghetti-squash (dismissed):

spaghetti-squash (approved):

packages/garbo/src/potions.ts
```diff
@@ -998,10 +1001,43 @@ export function variableMeatPotionsSetup(
 export function effectValue(
   effect: Effect,
   duration: number,
-  maxTurns?: number,
+  maxTurnsWanted?: number,
```
spaghetti-squash, line 1004:
```suggestion
```
I don't think we should ever pass this arg

PR #2296, opened 2025-03-19
Don't use lastchanceburn

Ignose:


docrostov:
approved, but i still want a feature that just sends 4m to shiverwarp, i will not be denied

PR #2314, opened 2025-03-29
Comma Deep Machine Tunnels Fix

Ignose:
For some reason, the Libram function isn't actually setting the familiar but directly calling the URL is.

### DIFF
--- a/packages/garbo/src/tasks/freeFight.ts
+++ b/packages/garbo/src/tasks/freeFight.ts
@@ -26,11 +26,13 @@ import {
   myMaxhp,
   mySoulsauce,
   numericModifier,
+  refreshStatus,
   restoreHp,
   retrieveItem,
   runChoice,
   runCombat,
   Skill,
+  toInt,
   use,
   useSkill,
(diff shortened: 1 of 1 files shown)


spaghetti-squash:
Instead of doing this, let's figure out what's wrong with the libram function and fix that. The libram function looks like this
```typescript
/**
 * Transforms Comma Chameleon into the familiar of choice if you already have the appropriate familiar equipment
 * Will not transform if you do not
 * @param fam determines what to transform into
 * @returns Whether Comma has been successfully transformed
 */
export function transform(fam: Familiar): boolean {
  if (currentFamiliar() === fam) {
    return true;
  }
  const equipment = familiarEquipment(fam);
  if (equipment !== $item.none) return false;
  if (!have_(equipment)) return false;
  visitUrl(
    `inv_equip.php?which=2&action=equip&whichitem=${toInt(equipment)}&pwd`,
  );
  return true;
}
```
which of these early returns is getting triggered?

spaghetti-squash:
Found it.

PR #2324, opened 2025-04-03
Use daily affirmation for pvp fights if appropriate

gausie:


packages/garbo/src/index.ts
```diff
@@ -530,6 +531,11 @@ export function main(argString = ""): void {
       stashItems.push($item`origami pasties`);
     }
 
+    // Prepare Daily Affirmation for PvP fights if desired
+    if (shouldAffirmationHate()) {
+      retrieveItem($item`Daily Affirmation: Keep Free Hate in your Heart`);
```
spaghetti-squash, line 536: Can you add a price check here?

gausie: `shouldAffirmationHate` should already cover that

spaghetti-squash (dismissed):

PR #2333, opened 2025-04-07
Do Cyber Realm 3 if the rewards are good

Ignose:


packages/garbo/src/lib.ts
```diff
@@ -1150,3 +1151,8 @@ export function improvedStats(thing: Item | Effect): Stat[] {
 export function improvesAStat(thing: Item | Effect): boolean {
   return improvedStats(thing).length > 0;
 }
+
+export const doCyberRealmZone3 =
+  get("_cyberFreeFights") === 0 &&
+  have($item`zero-trust tanktop`) &&
+  garboValue(CyberRealm.zone3Rewards()) > get("valueOfAdventure") * 9;
```
spaghetti-squash, line 1158: we don't want ground-level variables that call garboValue

spaghetti-squash: (in other words, find a way to make this whole thing a function. I would combine _cyberFreeFights with the zone 3 fights pref)

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -500,6 +535,11 @@ const NonBarfTurnTasks: AlternateTask[] = [
       }),
     sobriety: "sober",
   },
+  {
+    name: "CyberRealm 3",
+    ...cyberRealm3(() => !willDrunkAdventure()),
```
spaghetti-squash, line 540: you shouldn't follow this pattern for a sober-only task, and indeed you shouldn't pass `() => !willDrunkAdventure()` if you want to run this while sober even in situations where we'll drunk adventure that day

packages/garbo/src/tasks/freeFight.ts
```diff
@@ -874,7 +880,7 @@ const FreeFightTasks: GarboFreeFightTask[] = [
       have($skill`Torso Awareness`) &&
       have($skill`OVERCLOCK(10)`),
     completed: () => get("_cyberFreeFights") >= 10,
-    do: $location`Cyberzone 1`, // TODO Support other zones with better equipment and valuing hacker drops
+    do: doCyberRealmZone3 ? $location`Cyberzone 3` : $location`Cyberzone 1`, // TODO Support other zones with better equipment and valuing hacker drops
```
spaghetti-squash, line 883: this will similarly have to become a `() => Location`

spaghetti-squash (changes requested):

spaghetti-squash:
Can we consistently win zone 3 combats with our rocks-and-tanktop strategy?

Ignose:
> Can we consistently win zone 3 combats with our rocks-and-tanktop strategy?

Should be. We can also require digibritches, which is definitely enough.

spaghetti-squash:
Doesn't this need a meaningful combat overhaul after your experiences with it?

Ignose:
No, it needs stat nerfing though

PR #2338, opened 2025-04-09
Fixes for upcoming GSH

Shiverwarp:
Tentacles are now capped at 11 per day

### DIFF
--- a/packages/garbo/src/tasks/post/index.ts
+++ b/packages/garbo/src/tasks/post/index.ts
@@ -3,6 +3,7 @@ import {
   availableChoiceOptions,
   canAdventure,
   cliExecute,
+  haveEffect,
   inebrietyLimit,
   Item,
   itemAmount,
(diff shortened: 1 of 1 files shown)


packages/garbo/src/tasks/post/index.ts
```diff
@@ -373,11 +373,21 @@ function acquireAbortFreeRun(): GarboPostTask {
   };
 }
 
+function uneffectAttunement(): GarboPostTask {
+  return {
+    name: "Uneffect Eldritch Attunement After 11 Tentacles",
+    ready: () => get("_eldritchTentaclesFoughtToday") >= 11,
+    completed: () => !have($effect`Eldritch Attunement`),
+    do: () => uneffect($effect`Eldritch Attunement`),
```
spaghetti-squash, line 381: Is it worth adding an acquire step for sgeea?

spaghetti-squash: I think right now we'd either abort or hit an infinite loop, right?

I know the price cap I outlined in discord:
 ```typescript
haveEffect($effect`Eldritch Attunement`) * get("valueOfAdventure")
```
is probably economically not right, but I think having an abort is strictly better than the silent failure.

spaghetti-squash: maybe we could clamp that with some sane bounds?

PR #2340, opened 2025-04-10
Let them eat snootees

Shiverwarp:


packages/garbo/src/diet.ts
```diff
@@ -112,12 +129,22 @@ function eatSafe(qty: number, item: Item) {
     eat($item`fudge spork`);
   }
   useIfUnused($item`milk of magnesium`, "_milkOfMagnesiumUsed", 5 * MPA);
-
-  if (!eat(qty, item)) throw "Failed to eat safely";
+  withProperties(
+    {
+      autoSatisfyWithCloset: usingMoonZoneRestaurant
+        ? false
```
spaghetti-squash, line 135: Instead of using a ternary here, just do `!usingMoonZoneRestaurant && get("autoSatisfyWithCloset")`

packages/garbo-lib/src/wanderer/index.ts
```diff
@@ -227,7 +227,7 @@ export class WandererManager {
     [$location`The Haiku Dungeon`, { 297: 3 }], // skip
     [$location`Frat House`, { 1425: 4 }], // fight eXtreme Sports Orcs
     [$location`Madness Bakery`, { 1061: 6 }],
-    [$location`The Skeleton Store`, { 1060: 6 }],
+    [$location`The Skeleton Store`, { 1060: 5 }],
```
spaghetti-squash, line 230: Sure

packages/garbo/src/diet.ts
```diff
@@ -112,12 +129,22 @@ function eatSafe(qty: number, item: Item) {
...
+        : get("autoSatisfyWithCloset"),
+      autoSatisfyWithMall: !usingMoonZoneRestaurant,
```
spaghetti-squash, line 137: Does our `valueOfInventory` not protect us from this?

packages/garbo/src/diet.ts
```diff
@@ -128,7 +155,17 @@ function drinkSafe(qty: number, item: Item) {
       );
     }
   }
-  if (!drink(qty, item)) throw "Failed to drink safely";
+  withProperties(
```
spaghetti-squash, line 158: If this is repeated, slice it out into its own function or w/e

packages/garbo/src/diet.ts
```diff
@@ -150,17 +187,36 @@ function consumeSafe(
...
+  const usingMoonZoneRestaurant = availableFromMoonZoneRestaurant(item);
+  if (
+    !skipAcquire &&
+    !usingMoonZoneRestaurant &&
+    (averageAdventures > 0 || additionalValue)
+  ) {
     const cap = Math.max(0, averageAdventures * MPA) + (additionalValue ?? 0);
     acquire(qty, item, cap, true);
-  } else if (!skipAcquire) {
+  } else if (!skipAcquire && !usingMoonZoneRestaurant) {
```
spaghetti-squash, line 198: If this condition is true for both consecutive ifs, you should either:
- resequence the ifs to move your else up top, if that makes sense
- make a big if with the shared statement and a smaller if...else within

spaghetti-squash (changes requested):

PR #2342, opened 2025-04-12
Take items back out of closet before returning

Shiverwarp:


packages/garbo/src/fights.ts
```diff
@@ -2190,16 +2190,20 @@ export function estimatedFreeFights(): number {
...
       const avail = source.tentacle ? source.available() : 0;
       return typeof avail === "number" ? avail : toInt(avail);
     }) +
     possibleFreeFightQuestTentacleFights() +
     possibleFreeGiantSandwormQuestTentacleFights();
-  return clamp(totalFreeFights, 0, 11); // Capped at 11 per day
+  return clamp(
+    totalFreeFights,
+    0,
+    11 - (get("questL02Larva") !== "unstarted" ? 1 : 0), // Capped at 11, minus one if we have access to Science Tent
```
spaghetti-squash, line 2205: we should account for the skill as well. but also account for whether these have already been used

packages/garbo/src/tasks/freeFight.ts
```diff
@@ -280,7 +280,9 @@ const FreeFightTasks: GarboFreeFightTask[] = [
   {
     name: "Eldritch Tentacle",
     ready: () => get("questL02Larva") !== "unstarted",
-    completed: () => get("_eldritchTentacleFought"),
+    completed: () =>
```
spaghetti-squash, line 283: can you update this for the skill as well?

spaghetti-squash (approved):

PR #2351, opened 2025-04-16
Support using walkie talkies for ghosts

Shiverwarp:
I put acquiring and using a walkie talkie for our `GarboFreeFightTasks` ghosts in our dailies rather than within that task to not explode the ready and completed conditions for the task, and to avoid aborting if the acquire fails during the prepare/acquire step.


spaghetti-squash:
Live fast die young bad girls do it well

PR #2367, opened 2025-05-02
Embezzler lucky

Shiverwarp:
Problem I'm unsure how to fix: We need to count how many embezzlers we're fighting here to account for them when buffing with potions (I don't think we should actually account for them cost wise, but we _are_ spending turns fighting them, so that might cause us to run out of turns of buffs before we finish all our copiers, so the copytargetcount might need updating to account for that?


packages/garbo/src/outfit/target.ts
```diff
@@ -44,6 +44,9 @@ export function meatTargetOutfit(
   if (target === $location`Crab Island`) {
     const meat = meatDrop($monster`giant giant crab`) + songboomMeat();
     outfit.modifier.push(`${meat / 100} Meat Drop`, "-tie");
+  } else if (target === $location`Cobb's Knob Treasury`) {
```
spaghetti-squash, line 47: unfortunately I think we need a more precise check for this given that wanderers can be placed in the treasury (they can't be placed in crab island)

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -340,6 +343,61 @@ function aprilingSaxophoneLucky(additionalReady: () => boolean) {
...
+}
+
+function pillKeeperLucky(additionalReady: () => boolean) {
+  return {
+    completed: () => get("_freePillKeeperUsed"),
+    ready: () =>
+      additionalReady() &&
+      getBestLuckyAdventure().phase === "barf" &&
+      getBestLuckyAdventure().value() > get("valueOfAdventure") &&
+      have($item`Eight Days a Week Pill Keeper`),
```
spaghetti-squash, line 381: move this check above the `getBestLuckyAdventure`s

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -340,6 +343,61 @@ function aprilingSaxophoneLucky(additionalReady: () => boolean) {
...
+    prepare: () => {
+      if (!have($effect`Lucky!`)) {
+        retrieveItem($item`Eight Days a Week Pill Keeper`);
+        cliExecute("pillkeeper semirare");
+        if (!have($effect`Lucky!`)) {
+          set("_freePillKeeperUsed", true);
+          return;
+        }
+      }
+    },
```
spaghetti-squash, line 392: I think we should use a system similar to the one you've set up in embezzler.ts and for similar reasons (that is, separate lucky source from lucky execution).

BUT, the `turns` field should remain on the sources rather than on the lucky adventure task.

packages/garbo/src/tasks/embezzler.ts
```diff
@@ -0,0 +1,76 @@
...
+          set("_freePillKeeperUsed", true);
+          return;
+        }
+      },
+      spendsTurn: false,
+    },
+    {
+      name: "Fight Lucky Embezzler",
+      outfit: () =>
+        meatTargetOutfit(undefined, $location`Cobb's Knob Treasury`),
```
spaghetti-squash, line 68:
```suggestion
        meatTargetOutfit({}, $location`Cobb's Knob Treasury`),
```
this is just a matter of preference but I do think it scans nicer

spaghetti-squash (commented):
Great first step. I think this will be a good value-add

packages/garbo/src/target/fights.ts
```diff
@@ -1098,8 +1015,9 @@ export const copyTargetSources = [
 ];
 
 export function copyTargetCount(): number {
```
spaghetti-squash, line 1017: Is this worth having two separate functions (or a boolean arg) for? is everywhere we call `copyTargetCount` a place where we want to include embezzlers? even if you're fighting, say, roaches?

packages/garbo/src/tasks/dailyFamiliars.ts
```diff
@@ -60,18 +59,15 @@ function drivebyValue(): number {
 }
 
 function entendreValue(): number {
-  const targets = copyTargetCount();
+  const targets = highMeatMonsterCount("Scepter"); // Scepter can cause circular logic
```
spaghetti-squash, line 62: could you do a quick refactor to both have `highmeatMonsterCount` passed in rather than calculated both times, and to deduplicate some of the logic here?

spaghetti-squash: `highMeatMonsterCount` is an expensive calculation to make; this is a small micro-optimization, but instead of calculating it in both `entendreValue` and `drivebyValue`, we can calculate it once (during the robort it task) and pass that value into both functions, saving us some computational power. It's not really saving us much, but it's one of those things where making a habit of calling that function sparingly will add up if we're consistent.

Also, I know why scepter causes circular issues with like diet, but I have no recollection why it gets funky with robort code--might be worth updating the comment to mention why, if you remember

PR #2368, opened 2025-05-02
End of day Yachtzees

Shiverwarp:
This is a lot, and still lots of work to be done I think.

Primary problem right now I'm wrestling with: Someone with all the free rests and all the other NC forcers is going to exceed the 10 turns of fishy from fishy pipe (8 from cincho, 3 from tuba, 1 from clara)

This is accounted for in the expected turns taken, but we need a way to allocate usage of resources so that we aren't wasting them (I think best use of excess will be on Cincho, and I think this is possible to just hardcode in based on excess of the other two types)


packages/garbo/src/tasks/post/index.ts
```diff
@@ -238,7 +238,7 @@ function fallbot(): GarboPostTask {
   };
 }
 
-function refillCinch(): GarboPostTask {
+export function refillCinch(): GarboPostTask {
```
spaghetti-squash, line 241: If we're doing this as part of post, as long as we're running post with our adventures, we don't need a separate task to do this.

packages/garbo/src/yachtzee/familiar.ts
spaghetti-squash, line 1: I think overall yachtzee should become a subfolder of tasks, but this code should maybe be pulled over to other familiar stuff. it doesn't have to if imports make that weird, though.

packages/garbo/src/resources/aprilingband.ts
```diff
@@ -10,6 +17,11 @@ const instruments: {
   instrument: AprilingBandHelmet.Instrument;
   value: () => number;
 }[] = [
+  {
+    instrument: "Apriling band tuba",
+    value: () =>
+      realmAvailable("sleaze") ? (20000 - get("valueOfAdventure")) * 3 : 0,
```
spaghetti-squash, line 23: we're so back

packages/garbo/src/resources/mayam.ts
```diff
@@ -31,7 +31,7 @@ const MAYAM_RING_VALUES = {
         ({ expectedValue }) => expectedValue / 12,
       ),
     ) * 100,
-  chair: () => (CinchoDeMayo.have() ? 3 * 5 * felizValue() : 0),
+  chair: () => (CinchoDeMayo.have() ? 3 * 5 * felizValue() : 0), // TODO Account for reaching a Yachtzee NC breakpoint
```
spaghetti-squash, line 34: this shouldn't be hard but is still worth postponing until after we've merged.

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -104,6 +105,8 @@ import { shouldMakeEgg } from "../resources";
 import { lavaDogsAccessible, lavaDogsComplete } from "../resources/doghouse";
 import { hotTubAvailable } from "../resources/clanVIP";
 import { meatMood } from "../mood";
+import { freeFishyAvailable, shouldYachtzee } from "./yachtzee/lib";
```
spaghetti-squash, line 108: given the number of places these two functions are being used it might make more sense for them to just be in the regular lib

packages/garbo/src/tasks/yachtzee/index.ts
```diff
@@ -0,0 +1,125 @@
...
+        bestFamUnderwaterGear(yachtzeeFamiliar),
+      ];
+      if (overdrunk) equips.push($item`Drunkula's wineglass`);
+      return {
+        equip: equips,
+        modifier: modifiers,
+        avoid: $items`anemoney clip, cursed magnifying glass, Kramco Sausage-o-Matic™, cheap sunglasses`,
+        familiar: yachtzeeFamiliar,
+      };
+    },
```
spaghetti-squash, line 56: This code can be simplified by using an `Outfit` instance, instead of just returning a spec imo

spaghetti-squash:
Right now I believe we have code to use the fishy pipe during copy targets, to fight a single copy target underwater for an envyfish egg, and to interpolate a couple yachtzees into our copy target fights if you do so.

I would fully sundown the interpolated yachtzees, and make the fishy pipe for envyfish contingent on not doing EOD yachtzees.

packages/garbo/src/tasks/yachtzee/lib.ts
```diff
@@ -0,0 +1,62 @@
...
+    ? $item`Apriling band tuba`.dailyusesleft
+    : 0);
+
+export function maximumYachtzees(): number {
+  return clamp(
+    freeNCs(),
+    0,
+    have($item`fishy pipe`) && !get("_fishyPipeUsed")
+      ? 10
+      : haveEffect($effect`Fishy`),
```
spaghetti-squash, line 49:
```suggestion
     haveEffect($effect`Fishy`) + (have($item`fishy pipe`) && !get("_fishyPipeUsed")
      ? 10
      : 0),
```
this will probably lint super differently to how I have it typed, but it's Technically More Correct

packages/garbo/src/tasks/yachtzee/lib.ts
```diff
@@ -0,0 +1,62 @@
...
+  );
+}
+
+export function yachtzeeBuffValue(obj: Item | Effect): number {
+  return clamp(
+    (2000 *
+      (getModifier("Meat Drop", obj) +
+        getModifier("Familiar Weight", obj) * 2.5)) /
```
spaghetti-squash, line 57: iirc we have similar code for this with fewer magic numbers somewhere in our bjorn outfit code.

Malibu-Stacey:
You can also get 30 turns of Fishy from the Skate Park for free 1/day if you've done the quest.
My ASH code to do so is
```
  if (get_property("skateParkStatus") == "ice" && !get_property("_skateBuff1").to_boolean()) {
    cli_execute("skate lutz");
  }
```

packages/garbo/src/resources/realm.ts
```diff
@@ -52,7 +52,10 @@ export function checkVolcanoQuest() {
     volcanoItemValue,
     true,
   );
-  if (bestItem.item === $item`fused fuse`) {
+  if (
+    bestItem.item === $item`fused fuse` &&
+    (!shouldYachtzee() || volcoinoValue > 2000 - get("valueOfAdventure"))
```
spaghetti-squash, line 57: I think it might be smarter to integrate this logic into `volcanoItemValue`

packages/garbo/src/tasks/yachtzee/index.ts
```diff
@@ -0,0 +1,138 @@
...
+        !(
+          yachtzeeFamiliar.underwater ||
+          have($effect`Driving Waterproofly`) ||
+          have($effect`Wet Willied`)
+        )
+      ) {
+        outfit.modifier.push("underwater familiar");
+      }
+      outfit.equip(getBestWaterBreathingEquipment(freeNCs()).item);
+      outfit.equip(bestFamUnderwaterGear(yachtzeeFamiliar));
```
spaghetti-squash, line 53: I'm not sure we need both this and the `Underwater Familiar` modifier

packages/garbo/src/tasks/yachtzee/outfit.ts
```diff
@@ -88,26 +81,3 @@ export function prepareOutfitAndFamiliar(): void {
     );
   }
 }
-
-export function stickerSetup(expectedYachts: number): void {
```
spaghetti-squash, line 92: is this not worth it anymore or did this get relocated and I just haven't seen it yet

spaghetti-squash (approved):

PR #2376, opened 2025-05-17
shrug songs to cast ode

### LINKED ISSUE #2375: [BUG] no check for song space for ode
the diet code's ode handler is found here:

https://github.com/loathers/garbage-collector/blob/fceb38321cc4d30cdc6d07c3e03101f9009a7da0/packages/garbo/src/diet.ts#L152

in the current cockroach-copying universe where you cast buffs early to do piraterealm, the fact that accordion thieves can cast thingfinder & chorale means that they enter diet with fat loot, polka, thingfinder, and chorale. this means there's no space to cast ode. i have my mafia configured to warn me if i am about to consume booze without ode. 

i expect some of our garbo users just do not have that warning active, so they just lose out on the turns when they are playing AT, but it is a silly set of turns to lose out on. that's like 20+ turns! my recommendation is to essentially drop buffs until there is space to cast ode; start with anything that's not in the loot/polka/thingfinder/chorale list then remove loot then remove polka and it should work even for accounts without the 4-AT-song perm.  

log attached here: [Captain_Scotch_20250515.txt](https://github.com/user-attachments/files/20236877/Captain_Scotch_20250515.txt) -- for ease of reading, here is where i manually opened the relay browser, shrugged fat loot on my character sheet, and cast ode (at which point i let it run)

> use 1 distention pill
> Preference _distentionPillUsed changed from false to true
> Preference _concoctionDatabaseRefreshes changed from 320 to 321
> > Considering item Sacramento wine.
> > Based on organ size, planning to consume 21.
> > Trying to acquire 21 bottles of Sacramento wine; max price 40266.
>
> cast 1 The Ode to Booze
>
> cast 1 The Ode to Booze
>
> cast 1 The Ode to Booze
> Preference lastAdventureTrail changed from PirateRealm Island to Sailing the PirateRealm Seas
>
> uneffect Fat Leon's Phat Loot Lyric
>
> cast 1 The Ode to Booze
> You acquire an effect: Ode to Booze (20)

spaghetti-squash:
resolves #2375

### DIFF
--- a/packages/garbo/src/diet.ts
+++ b/packages/garbo/src/diet.ts
@@ -173,8 +173,9 @@ function shrugForOde() {
 function drinkSafe(qty: number, item: Item) {
   const prevDrunk = myInebriety();
   if (have($skill`The Ode to Booze`)) {
-    if (!have($effect`Ode to Booze`) && getSongCount() >= getSongLimit())
-      {shrugForOde();}
+    if (!have($effect`Ode to Booze`) && getSongCount() >= getSongLimit()) {
+      shrugForOde();
+    }
     const odeTurns = qty * item.inebriety;
     const castTurns = odeTurns - haveEffect($effect`Ode to Booze`);
     if (castTurns > 0) {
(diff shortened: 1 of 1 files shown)


docrostov (approved):
tested it. appears to work!

PR #2388, opened 2025-05-27
Wander for ultra rares

Shiverwarp:


packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -105,6 +104,7 @@ import { shouldMakeEgg } from "../resources";
 import { lavaDogsAccessible, lavaDogsComplete } from "../resources/doghouse";
 import { hotTubAvailable } from "../resources/clanVIP";
 import { meatMood } from "../mood";
+import { getAvailableUltraRareZones } from "garbo-lib/dist/wanderer/lib";
```
spaghetti-squash, line 107: this isn't quite how we want to import this; make sure the function is exported from `garbo-lib`'s index, then run `yarn`, then import it just from `garbo-lib`

packages/garbo-lib/src/wanderer/lib.ts
spaghetti-squash, line 1: I don't want to let great stand in the way of good, but it'd be really nice if we could incorporate this logic into our existing skipLists for various wanderer types.

spaghetti-squash: YR, freekill, and backup I think all make sense as being able to go to these locations? we have a separate check for 100% combat, so that's not an issue

spaghetti-squash: let's test it as-is and I'll fiddle with it later

PR #2397, opened 2025-06-08
Yachtzee: check for active NC force, don't need sober check

Shiverwarp:


packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -506,7 +506,7 @@ const NonBarfTurnTasks: AlternateTask[] = [
     ...lavaDogs(() => !willDrunkAdventure(), {}),
     sobriety: "sober",
   },
-  ...yachtzeeTasks(), // Use NC forces and adventure to get the Yachtzee NC
+  ...(getTasks(yachtzeeQuest) as AlternateTask[]), // Use NC forces and adventure to get the Yachtzee NC
```
spaghetti-squash, line 509: if this type assertion is needed, you should instead use the type argument
```suggestion
  ...getTasks<never, AlternateTask>(yachtzeeQuest)), // Use NC forces and adventure to get the Yachtzee NC
```
but it shouldn't be needed?

spaghetti-squash (approved):

PR #2403, opened 2025-06-09
Fix empty array maxby in ultrarare wanderer

Shiverwarp:


capncastom:
Would this be what would fix today's constant "JavaScript exception: Error: Cannot call maxBy on an empty array!" crashes?
If so Ill sjip the bug report. And thanks for the work!

PR #2454, opened 2025-07-10
Beret busk for salty mouth WIP

Ignose:


packages/garbo/src/diet.ts
```diff
@@ -128,6 +131,32 @@ function consumeWhileRespectingMoonRestaurant(command: () => void, item: Item) {
...
+      {
+        uselessEffects: [],
+        buyItems: true,
+      },
+      i,
+    );
+
+    const buskEffects = Object.keys(
+      beretBuskingEffects(optimalPower, get("_beretBuskingUses")),
+    ).map(toEffect);
```
spaghetti-squash, line 150: This will error if it is ever called

PR #2476, opened 2025-08-08
Don't filter by times rested in the cinch post task

Shiverwarp:


spaghetti-squash (commented):
Can I ask why this is desirable?

PR #2502, opened 2025-08-29
Salty busk Alternate Form

Shiverwarp:
Sera already has a PR up, here's a different version if we want.

This is a first step just to get busking in, improvements I think could be doable:
-More robust predicting for hammertime to optimize the amount of Salty Mouth, and also for general buffs
-Generating a useless effects list that includes effects that we have more than our expected turns amount of, and also including spells that we can cast ourselves for MP


packages/garbo/src/diet.ts
```diff
@@ -184,9 +219,18 @@ function drinkSafe(qty: number, item: Item) {
...
+      buskForSaltyMouth();
+      consumeWhileRespectingMoonRestaurant(() => {
+        if (!drink(1, item)) throw "Failed to drink safely";
+      }, item);
+    }
+  } else {
+    consumeWhileRespectingMoonRestaurant(() => {
+      if (!drink(qty, item)) throw "Failed to drink safely";
+    }, item);
+  }
```
spaghetti-squash, line 233:
```suggestion
  consumeWhileRespectingMoonRestaurant(() => {  
    if (item.notes?.includes("BEER") && canBusk()) {
      for (let i = 0; i < qty; i++) {
        buskForSaltyMouth();
        if (!drink(1, item)) throw "Failed to drink safely";
      }
    } else if (!drink(qty, item)) throw "Failed to drink safely";
  }, item);
```
this will need linting, probably

PR #2515, opened 2025-09-02
Do the wave, enabling negative pressure and such shenanigans

Ignose:


packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -1317,6 +1319,15 @@ export const BarfTurnQuest: Quest<GarboTask> = {
         !(totalTurnsPlayed() % 11) &&
         meatMood().execute(estimatedGarboTurns()),
       post: () => {
+        if (
+          !get("_seadentWaveUsed") &&
+          // Seadent can change names, so we use the item ID to avoid mafia errors
```
gausie, line 1324: This doesn't make any sense. Address the item by its canonical name like we would with pantogram or any other name-changing item

spaghetti-squash:
Drafted postnerf

PR #2557, opened 2025-12-02
Skeleton of crimbo past

Ignose:


Ignose:
Opening this so others can use it as a starting point

PR #2562, opened 2025-12-06
start refactoring target code

spaghetti-squash:


packages/garbo/src/outfit/free.ts
```diff
@@ -55,34 +60,44 @@ const famExpValue = new Map<Familiar, Delayed<number>>([
...
 export function freeFightOutfit(
   spec: OutfitSpec = {},
+  destination: Location | WanderDetails,
   options: FreeFightOutfitMenuOptions = {},
 ): Outfit {
   cleaverCheck();
 
-  const computedSpec = computeOutfitSpec(spec, options);
+  const location =
+    destination instanceof Location
```
gausie: You do this enough that it could be a utility function

PR #2588, opened 2026-01-02
Exclude The Barracks from CookbookBat Quests as per #2565

### LINKED ISSUE #2565: [BUG] KoL erroneously gives cookbookbat quests in Island Barracks post AT nemesis boss defeat
The mini quests the CBB gives for extra ingredients usually name a monster and a zone where it can be fought, however sometimes an AT who has defeated Somerset Lopez, Demon Mariachi will be given a quest to fight one of the three mariachis who are native to the Barracks before the final Lopez fight, which is not a thing that can happen, all the mariachis are gone at that point.

Now, this is clearly a KoL bug, and I have reported it, however garbo's handling of the bug is problematic. The zone is 100% NC, and all NC choices result in an adv used but no combat occurs, so even when a new CBB mini quest window has opened, garbo doesn't know because it sees no combat text. I set the choice option to leave one time and garbo chewed up 50 adv trying to fight a mariachi that is no longer there.

To work around I take the CBB to Barf mountain and kill tourists until the next CBB mini quest comes up in the post combat messages..

I would suggest the mariachi mini quests be ignored (alert, sleepy and surprised), perhaps even before the final boss fight, but definitely after.  

Finish the AT nemesis quest, and run garbo. It won't happen every day, or even most days, maybe one a week on average for me so far.

**Expected behavior**
I would expect garbo to keep away from the Island Barracks NC, as no good can come from there

**Observed behavior**
Once the mariachi CBB quest is triggered, garbo will go no where else until manual intervention occurs

Tokoeka:
[BUG] KoL erroneously gives cookbookbat quests in Island Barracks post AT nemesis boss defeat

Fixes #2565

### DIFF
--- a/packages/garbo-lib/src/wanderer/cookbookbatquest.ts
+++ b/packages/garbo-lib/src/wanderer/cookbookbatquest.ts
@@ -11,7 +11,7 @@ export function cookbookbatQuestFactory(
   const questReward = get("_cookbookbatQuestIngredient");
   const questMonster = get("_cookbookbatQuestMonster");
   // TODO: These are fixed in mafia (we get the correct location), however we cannot use peridot to target them for cookbookbat specifically, because we cannot know which monster cookbookbat wants because of duplicate names
-  const blackListedLocations = $locations`The Orcish Frat House, The Orcish Frat House (Bombed Back to the Stone Age), The Island Barracks`;
+  const blackListedLocations = $locations`The Orcish Frat House, The Orcish Frat House (Bombed Back to the Stone Age)`;
   if (
     ["yellow ray", "freefight", "freerun"].includes(type) && // Runs still get you the quest reward
     questLocation &&
--- a/packages/garbo-lib/src/wanderer/lib.ts
+++ b/packages/garbo-lib/src/wanderer/lib.ts
@@ -218,7 +218,7 @@ function canWanderTypeFreeFight(location: Location): boolean {
   );
 }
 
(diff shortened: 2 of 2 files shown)


spaghetti-squash:
I don't think this quite does what you want, which means this is a good opportunity to make some of this code clearer.

The array you updated is used in calculating the expected likelihood that the cookbookbat gives a location we can safely adventure in. In particular, that array consists of locations that KoL will not pick for cookbookbat quests despite mafia's canAdventure saying they're open.

What we actually want to do is prevent garbo from ever wandering there, which would happen somewhere in the garbo-lib workspace's wanderer section. There, we have code that prevents us from adventuring places like the arid extra-dry desert--consider looking around there for where to add this.

Tokoeka:
> I don't think this quite does what you want, which means this is a good opportunity to make some of this code clearer.
> 
> The array you updated is used in calculating the expected likelihood that the cookbookbat gives a location we can safely adventure in. In particular, that array consists of locations that KoL will not pick for cookbookbat quests despite mafia's canAdventure saying they're open.
> 
> What we actually want to do is prevent garbo from ever wandering there, which would happen somewhere in the garbo-lib workspace's wanderer section. There, we have code that prevents us from adventuring places like the arid extra-dry desert--consider looking around there for where to add this.

Yeah, I wasn't entirely sure what this exclusion list was for, I was only going off the fact that the code appeared to not generate a wanderertarget "quest" if the questlocation was one of the blacklisted locations, which felt like it would do what was required. 

Agree its almost certainly generically better to avoid the location entirely. Have updated accordingly

PR #2589, opened 2026-01-07
Add clubbing into next week as a wanderer copier

Ignose:


packages/garbo/src/outfit/target.ts
```diff
@@ -146,5 +147,12 @@ export function meatTargetOutfit(
     outfit.equip($item`Roman Candelabra`);
   }
 
+  if (
+    get("clubEmNextWeekMonster", Monster.none) === Monster.none &&
```
spaghetti-squash: A few things:
1. I would use `$monster.none` over `Monster.none`; I think? that's more consistent with the rest of the codebase? we should really pick a convention here if we haven't
2. all of these `get`s default to null when it would otherwise be the MafiaClass's `none` value; just let it be null! null is falsy, so this could be `!get("clubEmNextWeekMonster")` or, more verbosely, `get("clubEmNextWeekMonster") === null`
3. I don't actually think that should be the check; I think we should check whether the `Counter` is Infinity (that is, not set) (`Infinity` is a number value in js) and whether we have casts remaining. Technically we shouldn't need the `have` check, because `equip` checks that for us, but follow whatever we do elsewhere in the code on this one. (randle looks like it doesn't check it, but it might be alone in this)
4. I'm not sure this should be this far down in the outfit. The more important something is, broadly speaking, the earlier it should be, to decrease the chances that some other equipment edges it out. This is more of a nitpick than anything else, and I recognize that this is probably equal in value to randle which it neighbors.'
5. Could have sworn I had a fifth but I got distracted.

spaghetti-squash:
https://github.com/loathers/garbage-collector/blob/7e8356bfbc3526af3074bd74a7d50b0e3a008afb/packages/eslint-config-garbo/index.mts#L7 btw this is what you bump if you want to upgrade eslint-plugin-libram

packages/garbo/src/combat.ts
```diff
@@ -232,6 +232,11 @@ export class Macro extends StrictMacro {
           get("_enamorangs") < 5 && !get("enamorangMonster"),
           Macro.tryHaveItem(itemOrSkill),
         );
+      case $skill`Club 'Em Into Next Week`:
+        return this.externalIf(
+          get("_clubEmNextWeekUsed", 0) < 5 && !get("clubEmNextWeekMonster"),
```
spaghetti-squash: Does that pref get cleared when you encounter your target?

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -947,6 +947,46 @@ const BarfTurnTasks: GarboTask[] = [
...
+                  : (11 - get("_mimicEggsObtained")) * 50)
+              ? {
+                  familiar: $familiar`Chest Mimic`,
+                  weapon: $item`legendary seal-clubbing club`,
+                }
+              : {},
+            wanderer().getTarget({
+              wanderer: "wanderer",
+              allowEquipment: false,
+            }).location,
```
spaghetti-squash: I really should adjust this code to allow you to pass a `Destination` in here rather than just a `Location`. I guess note for the future

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -947,6 +947,46 @@ const BarfTurnTasks: GarboTask[] = [
...
+      digitizedTarget()
+        ? meatTargetOutfit(
+            get("_mimicEggsObtained") < 11 &&
+              $familiar`Chest Mimic`.experience >
+                (digitizedMonstersRemaining() === 1
+                  ? 50
+                  : (11 - get("_mimicEggsObtained")) * 50)
+              ? {
+                  familiar: $familiar`Chest Mimic`,
+                  weapon: $item`legendary seal-clubbing club`,
```
spaghetti-squash: We do not need to and actively should not force us to equip the club during these fights

packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -947,6 +947,46 @@ const BarfTurnTasks: GarboTask[] = [
     spendsTurn: () =>
       !SourceTerminal.getDigitizeMonster()?.attributes.includes("FREE"),
   },
+  {
+    name: "Club Into Next Week Monster",
+    completed: () => Counter.get("Club 'Em Into Next Week Monster") > 0,
+    outfit: () =>
+      digitizedTarget()
```
spaghetti-squash: This should not reference digitize at all.

PR #2594, opened 2026-01-17
value expected turns more sanely

Ignose:


spaghetti-squash:
instead of having this be an arg, just add `30` in specifically the potion/mood code. maybe export a const from lib called like `TURNCOUNT_WIGGLE_ROOM` or something like that?

Issue #2444, opened 2025-07-01
[FEATURE] Preference to ignore Pilsners


xn:
**Is your feature request related to a problem? Please describe.**
I'd like to save my pilsners for day 2 of my loop when I have all of my Busks available for salty mouth. Since you can't closet nor display them, I can't see a way to hide them otherwise.

I think something like what was undone here would work: (removing the ascend boolean to pilsner consumption.)
https://github.com/loathers/garbage-collector/pull/1405

**Describe the solution you'd like**
This could be a command line arg or a preference.

**Describe alternatives and complications**
I don't think there is an alternative?

A consideration is thinking about whether garbo is going to incorporate Beret into the diet and if it makes sense to implement in such a way as to make that future change easier.

**Additional context**
Thank you!
Also, I'm happy to implement this myself.

spaghetti-squash:
This seems reasonable, I would accept this as a PR

PR #2599, opened 2026-01-21
Add usepilsners diet option

xn:
This was the smallest implementation I could think of. I don't have a good sense if yet another flag is overdoing it. Happy to re-implement. 

* Introduce a new configuration flag 'usepilsners' to allow users to opt-in for instant pilsners as part of their diet.
* Modify the diet logic to filter pilsners based on the new configuration setting.

Addresses https://github.com/loathers/garbage-collector/issues/2444


spaghetti-squash:
Looks like this is failing linting. I haven't taken too thorough a look at it, because the commas thing has made the diff bigger than it needs to be. Try a quick `yarn format` or set up Husky to do it for you

PR #2601, opened 2026-01-26
Allow wandering at Oasis if we have Ultrahydrated

Shiverwarp:


spaghetti-squash (approved):
this looks good to me, merge it if you feel you've tested it sufficiently

PR #2608, opened 2026-02-02
BCZ Refracted Gaze

Shiverwarp:
This is going to need some work, this is just initial stuff, and will be in progress

Things I think will need to be done:
1. Splitting up item drop bonuses from monsters and "other" bonuses (like CBB and bofa) in our wanderer evaluation
2. Somehow be able to pass equip requirements as well as a macro to use (how can we place the macro where we want within the task?)
3. Account for how if we're using BCZ we do not want to use peridot, and also if we use monodent to swap to fish, we will lose the "other" bonuses in favor of more drops. (If we don't use monodent, we'll still have a 1/monstersInZone chance of getting the bonus, so we can take some average there)

If we want to get fancy we can maybe do some other stuff, like only activating monodent if the monster we end up with is worth it to add the drops from, while losing the other bonuses, but I believe that's unnecessary for a first pass unless someone is feeling passionate


packages/garbo-lib/src/wanderer/index.ts
```diff
@@ -141,28 +192,63 @@ function bestWander(
...
-      targetedMonsterValue(monsterValues);
+
+    const shouldRefracted =
+      refractedGazeValue > monsterAverageValue &&
+      refractedGazeValue > monsterTargetedValue;
 
     const shouldPeridot =
       PeridotOfPeril.canImperil(location) &&
       !unperidotableZones.includes(location) &&
+      !shouldRefracted &&
```
spaghetti-squash: I feel like we should call this `shouldRefract`

spaghetti-squash (changes requested):
This looks broadly correct, but I will have an easier time following what all this does after these changes.

PR #2611, opened 2026-02-07
correct spendsTurns in some tasks

Shiverwarp:


packages/garbo/src/tasks/barfTurn.ts
```diff
@@ -586,7 +586,7 @@ const NonBarfTurnTasks: AlternateTask[] = [
             11 - get("_mimicEggsObtained"),
           )
         : 0,
-    spendsTurn: true,
+    spendsTurn: !globalOptions.target.attributes.includes("FREE"),
```
spaghetti-squash: This needs to be delayed

PR #2640, opened 2026-03-26
handle toilet alongside pantsgiving

spaghetti-squash:


packages/garbo/src/tasks/post/index.ts
```diff
@@ -413,7 +413,9 @@ function usePorkToilet(): GarboPostTask {
         throw new Error("Using toilet did not reduce our fullness!");
       }
     },
-    available: () => $item`Pork Elf toilet`.name in getCampground(),
+    available: () =>
+      $item`Pork Elf toilet`.name in getCampground() &&
+      (!CinchoDeMayo.have() || totalFreeRests() <= get("timesRested")),
```
Shiverwarp, line 418: Should this maybe be in ready rather than available?

PR #2647, opened 2026-04-16
fix: allow feeding robortender drinks from inventory

erhlee-bird:
Prices for fish heads and drive-by shootings have gone up. Made a small change so that you can stock up manually and have garbo use up your inventory to feed Robortender even when prices spike.

### DIFF
--- a/packages/garbo/src/tasks/dailyFamiliars.ts
+++ b/packages/garbo/src/tasks/dailyFamiliars.ts
@@ -114,7 +114,7 @@ export function prepRobortender(): void {
     }
     useFamiliar($familiar`Robortender`);
     const drink = toItem(drinkName);
-    if (retrievePrice(drink) > priceCap) {
+    if (!have(drink) && retrievePrice(drink) > priceCap) {
       if (mandatory) {
         setBestLeprechaunAsMeatFamiliar();
         if (


docrostov:
so... the problem i see with this is that the pricing still feels relevant even if you have available drinks. this is because you can always sell an item in your inventory for the price of whatever's in the mall. so, if the value you're going to derive from drinking the drink is less than the value of selling the drink, the general philosophy of garbo is that you shouldn't use your inventory drink when you could gain more meat by doing the mall-selling.

that said, i do get your point. perhaps this is a case for a garbo alias that feeds your robort drinks before you run garbo?  @spaghetti-squash what do you think

erhlee-bird:
That makes sense, I appreciate the nuance and just thought it might be convenient especially in a case where for pvp I valued fitegen over strict meat opportunity cost.

spaghetti-squash:
I'm afraid I agree with scotch; if you want garbo to drink robort drinks when it's not mathematically optimal to do so, you can do so before running garbo. This change, as implemented, would add _your_ intended behavior for _all_ users. I could maybe see putting this behind a pref or argument, but given that "chug a DBS beforehand" isn't particularly hard, I think that would be needlessly complex.

I appreciate the desire to fix this problem in the script, though!

erhlee-bird:
Makes total sense, thanks for reviewing and your work on the project :tada:

PR #2649, opened 2026-04-23
Use barrel shrine buff if Pastamancer

Shiverwarp:


docrostov (approved):
tested these conditions/commands w/ jsl. all works. good for the world.

PR #2651, opened 2026-04-24
Smash barrels

Shiverwarp:
Thank you Katarn via discord

Should we confirm whether we can actually access the barrel full of barrels? (Even if we can't, I don't think it would matter because the regex won't match)


spaghetti-squash (approved):
I would probably personally do this with an xpath but this is probably fine

PR #2655, opened 2026-05-04
Swap to Mayo from Asdon after buffing before diet

Ignose:


packages/garbo/src/diet.ts
```diff
@@ -1470,6 +1477,15 @@ export function runDiet(): void {
       }
       printDiet(dietBuilder.diet(), "FULL");
     } else {
+      if (asdonToMayo()) {
+        AsdonMartin.drive(
+          $effect`Driving Observantly`,
+          dietAdventures(dietBuilder.diet()) +
+            (globalOptions.ascend ? 0 : estimatedTurnsTomorrow),
+        );
+        workshedTasks();
```
spaghetti-squash, line 1486: This function doesn't do anything, it returns a list of tasks

Ignose: function PostQuest uses this, and based on my logs the first instance of PostQuest being called is when the buffing and subsequent swapping occurs. What am I missing?

spaghetti-squash: PostQuest uses this function, which creates a list of tasks that are part of the list of tasks in the quest. PostQuest then returns, as you might expect, a big ol' quest, which gets run as part of a `runGarboQuests` call every time it appears.

Ignose: Okay. So we'd want to `runGarboQuests([...workshedTasks()])`

spaghetti-squash: Yes but I also extremely think we don't want to do that. I'll put in a draft PR for how I would do this

PR #2662, opened 2026-05-12
Can't club giant giant crab into next week

Shiverwarp:


spaghetti-squash (commented):
can you move the randle code into this block as well, in addition to any other copiers you can find?

packages/garbo/src/outfit/target.ts
```diff
@@ -45,8 +45,21 @@ export function meatTargetOutfit(
     new Error(`Failed to construct outfit from spec ${JSON.stringify(spec)}`),
   );
 
-  if (nextWeekReady()) outfit.equip($item`legendary seal-clubbing club`);
   const { location } = toAdventure(adventureArgument ?? $location.none);
+  if (location !== $location`Crab Island`) {
```
spaghetti-squash, line 49: I guess longer term we should really have a "if this is _really_ a copy target fight" check, rather than just location, because we don't want to do this during embezzlers either

Shiverwarp: We can just take the `adventureArgument` and check whether the monster is copiable? I guess we would want to make sure we're always passing it in for these types of fights?

PR #2679, opened 2026-06-21
feat: use level 11 spice ghost before eating

midgleyc:
If we happen to have a level 11 Spice Ghost, bind it before eating.


packages/garbo/src/diet.ts
```diff
@@ -165,6 +167,14 @@ function eatSafe(qty: number, item: Item) {
   if (mallPrice($item`fudge spork`) < 3 * MPA && !get("_fudgeSporkUsed")) {
     eat($item`fudge spork`);
   }
+  if (
+    myClass() === $class`Pastamancer` &&
+    myThrall() !== $thrall`Spice Ghost` &&
+    !get("_legendarySpiceGhostFood") &&
+    $thrall`Spice Ghost`.level + (get("pumpkinSpiceWhorlUsed") ? 3 : 0) >= 11
```
spaghetti-squash, line 174: Is it worth having a check here for the legendary pasta wand?

spaghetti-squash (approved):
approving so you can merge if the answer to my question is "no"

midgleyc:
Indeed it's "no" because it's not necessary to have the wand -- if you have used the pumpkin spice whorl and your Spice Ghost is level 8 or higher, you get the effect.

Thanks!

packages/garbo/src/diet.ts
midgleyc: No -- with the pumpkin spice whorl and level 8, you don't need the wand.

PR #2683, opened 2026-07-08
Added a visit to the Speakeasy Pool Table if the player has access

TommyDingle:
**Changes**
I noticed Garbo doesn't visit the Speakeasy Pool Table for its free daily meat. Added a check for access, and a Pool Skill maximizer to get the most meat out of it before visiting.

**Notes**
I've never added to a GitHub project before, and thought I'd try my hand at it with this as my first addition, especially since it's small and low stakes.

Let me know if there's anything wrong with it, or if there are any best practices I missed!

### DIFF
--- a/packages/garbo/src/tasks/daily.ts
+++ b/packages/garbo/src/tasks/daily.ts
@@ -559,6 +559,15 @@ const DailyTasks: GarboTask[] = [
     spendsTurn: false,
   },
   {
+    name: "Speakeasy pool table",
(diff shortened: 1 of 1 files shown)


spaghetti-squash:
Is this part of breakfast? Does mafia not track it?

TommyDingle:
I searched through mafia's code a bit and it's not in breakfast and it doesn't look like it tracks it's state at all.

Issue #2684, opened 2026-07-09
[BUG] Garbo fails when Deep Machine Tunnels Snowglobe is > 2000 meat 


RooftopTal:
## Describe the bug
I keep hitting an error when Deep Machine Tunnel Snowglobes are too expensive (> 2000) AND you have a comma chameleon (from the code this looks like a chameleon only bug).

This fails during that step, and breaks prevents mafia from running. The relatively easy fix for this is to just buy a snowglobe manually and use it before rerunning.

## Potential fix

My suspicion is that the issue is here:
https://github.com/loathers/garbage-collector/blob/e63f85f110d4d51b86f43046dd7a75f60334647b/packages/garbo/src/tasks/freeFight.ts#L735

The `ready/completed` sections check if we we have a machine elf, or comma will be "cheap" enough to enter - but there's no actual check that the prices are as we expect below.

### Non-risky fix:
* Update `dmtCommaValuable` to enforce price maximums of `10000`, `2000` and return false if either exceed those benchmarks
* We then could eliminate those cost checks from the acquire steps in the machine elf test; it wouldn't be necessary? In theory? But it could prevent race conditions where the price jumps in between the value check/acquire step.

### Maybe risky fix?:
* eliminate any hard cost checks from the tunnel diving, and trust that `dmtCommaValuable` ensures the cost will be worth it for the garbo user
* This is mildly risky for two reasons:
  * A huge price jump at exactly the wrong time could mean the user is out of pocket when mafia buys the items
  * This potentially increases the price of snowglobes jumping up as garbo farms them, blocking out content for folks who just want to try out the machine tunnels. Which... might be against garbo philosophy? idk

I'm happy to make either of these changes as a PR, if someone with more garbo experience lets me know which would be the cleaner fix?
  
## To Reproduce

1. Have a comma chameleon, but no machine elf
2. Have no snowglobes or basketballs in your inventory
3. Ensure that running free turns in the machine tunnels is profitable for you, but EITHER:  a. the basketball is > 10k
  b. the snowglobe is > 2k
3. Run garbo
4. Wait for it to hit the machine tunnels free fights

It should then fail and break out of garbo. This will fail every time you run garbo unless you buy & manually use a snowglobe

## Expected behavior
Garbo should either just run the machine elf tunnel turns (if it finds it profitable) or skip the tunnel diving (if it doesn't want to buy the snowglobes/basketballs at over some hardcoded maximum)

## Session Log
Error section:
```
[...]

PR #2685, opened 2026-07-12
2684 - skip dmt if comma threshholds are exceeded

RooftopTal:
See: https://github.com/loathers/garbage-collector/issues/2684
...for context

This should fix an occasional issue where garbo breaks out if DMT tunnels are (a) profitable for garbo but (b) exceed garbo price threshholds

---

I've gone for a fix where we _enforce the existing_ price thresholds, rather than deleting them; we seem to mostly use a `maxPrice` in garbo when trying to acquire items, & I didn't want to go against what seems to be best practice?


packages/garbo/src/tasks/freeFight.ts
```diff
@@ -734,7 +737,7 @@ const RAW_FIGHTS: Parameters<typeof freeFightTask>[0][] = [
   {
     name: "Machine Elf",
     adventure: $location`The Deep Machine Tunnels`,
-    ready: () => have($familiar`Machine Elf`) || dmtCommaValuable(),
+    ready: () => have($familiar`Machine Elf`) || dmtCommaValuable(10000, 2000),
```
RooftopTal, line 740: I put these values in as variables, since it felt weird to hardcode the values _both_:

* in dmtCommaValuable
* On lines 743 & 748 below

It feels like a better way to do this might be to add consts at the top of this task file:

```
const MAX_DMT_SNOWGLOBE_PRICE = 2000;
CONST MAX_SELF_DRIBBLING_BASKETBALL_PRICE = 10000;
```

...but we don't seem to be doing that anywhere elsewhere, and I didn't want to break convention.

packages/garbo/src/tasks/freeFight.ts
RooftopTal, line 755: Not part of the review, but is this comment doing anything?

I can't see anything on the wiki suggestion comma users don't get to convert abstractions, and it looks like I'm doing it here? Shall I just delete?

Image

spaghetti-squash: cc @Ignose

packages/garbo/src/tasks/freeFight.ts
spaghetti-squash: Go ahead and do it, except also put `_`s in those numbers for legibility.

spaghetti-squash:
I think it makes sense to respect price caps, although I would consider also raising them if we're routinely hitting them. The price caps should be a sanity check--"oh my god, why did you spend a million meat on a snowglobe, of course no one is buying these abstractions for 10m"--not something we routinely hit.

packages/garbo/src/tasks/freeFight.ts
Ignose: Bad spading on how comma works, then. We don't need this else if it works.

PR #2686, opened 2026-07-24
Skip Assert your Authority on insta-kill-immune monsters

eegeeZA:
## Summary

Assert your Authority is capped at 3 uses/day.
— if that monster is `BOSS`/`LUCKY`/`ULTRARARE`
flagged, the cast still burns a daily charge even though the insta-kill
fails against it. Observed live: garbo expected "man with the red buttons"
but a `Ron "The Weasel" Copperhead` (BOSS-flagged) preempted the
wander, and the skill was cast at it anyway.

- Guard the skill cast with a fight-time `monsterid` allowlist built from
  `availableMonsters(location)` — the zone's native monsters, already
  excluding `LUCKY`/`ULTRARARE`/`BOSS`. This has to be a KoL macro-language
  condition rather than a JS-side check: `GarboStrategy` submits this macro
  as an autoattack, which grimoire's engine compiles before the actual
  monster is revealed.
- `availableMonsters` previously wasn't exported from the `garbo-lib`
  package entry point (only from `wanderer/lib.ts` internally); added it to
  the re-export list.

## Scope note

`Cheese Wizard Fondeluge`, `Spit Acid`, and `Pig Skinner Free-For-All` in the
same file share the same unconditional `.skill()` cast shape and plausibly
the same gap, but they're not daily-capped resources in the same way and
were left out of scope here — worth a follow-up if it's a live problem for
them too.

## Test plan

`packages/garbo` has no automated test suite; verified via:

- [x] `yarn check` (tsc across workspaces) — passes
- [x] `yarn lint` (eslint + prettier) — passes
- [x] `yarn build` — passes; confirmed the guard logic and `availableMonsters`
      are present in the compiled `garbo.js` output

No way to exercise KoL's live preemption from CI — the fix's
runtime effect isn't test-covered.

---
Assisted by Claude Code — reviewed by @eegeeZA; verified by the test suite.

### DIFF
--- a/packages/garbo-lib/src/index.ts
+++ b/packages/garbo-lib/src/index.ts
@@ -3,6 +3,7 @@ import type { ValueFunctions } from "./value";
 import { WandererManager } from "./wanderer";
 import type { DraggableFight, WanderDetails, WanderOptions } from "./wanderer";
 import {
+  availableMonsters,
   canAdventureOrUnlock,
   getAvailableUltraRareZones,
   hasNameCollision,
@@ -12,6 +13,7 @@ import {
 export {
   makeValue,
   WandererManager,
+  availableMonsters,
   canAdventureOrUnlock,
   getAvailableUltraRareZones,
   hasNameCollision,
--- a/packages/garbo/src/tasks/barfTurn.ts
+++ b/packages/garbo/src/tasks/barfTurn.ts
@@ -63,6 +63,7 @@ import {
 } from "libram";
 import { getTasks, Outfit, OutfitSpec, Quest } from "grimoire-kolmafia";
 import {
+  availableMonsters,
   canAdventureOrUnlock,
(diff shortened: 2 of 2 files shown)


spaghetti-squash:
There's no such thing as semirares anymore. Why were you lucky! in the zeppelin?

eegeeZA:
> There's no such thing as semirares anymore. Why were you lucky! in the zeppelin?

I think I misdiagnosed the 'semi-rare' part, but I still think this could protect against:
image

Rinn:
this seems like more of an issue of allowing wandering into the zeppelin when the quest isn't finished

eegeeZA:
Sounds fair. Should this be adjusted to check for `zeppelinProgress < 5` (or `questL11Ron == "finished"`)?

PR #2689, opened 2026-07-28
Cowo I guess

Ignose:


bigbongobeats:
Cowo rocks. I recommend adding additional handling to dailyFamiliars.ts for robortender to value Bloody Nora when running cowo mode.

I'd also recommend handling das boot in computeBarfOutfit(), where we can then do something like:
if (needsUnderwaterFamiliarEquipment(spec.familiar)) {
  outfit.equip($item`das boot`);

Ignose:
> Cowo rocks. I recommend adding additional handling to dailyFamiliars.ts for robortender to value Bloody Nora when running cowo mode.
> 
> I'd also recommend handling das boot in computeBarfOutfit(), where we can then do something like: if (needsUnderwaterFamiliarEquipment(spec.familiar)) { outfit.equip($item`das boot`);

Can you show what

>  jsl $familiar`robortender`

 returns in mafia after Robortender has been given a Bloody Nora?

bigbongobeats:
That works great! Garbo ate a bloody nora as expected.

In addition to the underwater familiar fix, one note on monodent banish - it seems to currently not evaluate banishes before combat, and, as a result, wears 'dent to prepare for a potential banish that is not needed. Possibly a mismatch of calling the banish source Monodent rather than skill itself here - 
`method.name === sourceItemOrSkill.name`

Setting bolts used to 11 or more suppresses this behavior.

bigbongobeats:
I think we can handle the robort underwater breathing case with one change in outfit/barf.ts.

I think it would look something like:

```ts
try {
  return computeBarfOutfit(
    {
      familiar,
      ...spec,
      famequip:
        spec.famequip === $item`das boot` && familiar.underwater
          ? undefined
          : spec.famequip,
    },
    sim,
  );
} finally {
```

Given that script isn't in this PR I can't add a comment there.

If we wanted to also handle bathysphere here we could, but not sure that's in scope.

packages/garbo/src/resources/cowoResources.ts
```diff
@@ -0,0 +1,134 @@
...
+  Macro,
+} from "libram";
+import { garboAverageValue } from "../garboValue";
+
+const monsters = $monsters`Mer-kin rustler, sea cowboy`;
+
+export function getCowoMonstersToBanish(): Monster[] {
+  const banishedMonsters = getBanishedMonsters();
+  const alreadyBanished = Array.from(banishedMonsters.values());
+  return monsters.filter((monster) => !alreadyBanished.includes(monster));
```
spaghetti-squash: why not just use `isBanished`

packages/garbo/src/resources/cowoResources.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+interface BanishMethod {
+  available: () => boolean;
+  macro: () => Macro;
```
spaghetti-squash: none of the macros here have any reason to be delayed

packages/garbo/src/resources/cowoResources.ts
```diff
@@ -0,0 +1,134 @@
...
+      $item`salinated mint julep`,
+      $item`sand dollar`,
+      $item`sea lace`,
+      $item`seaweed`,
+      $item`shark cartilage`,
+      $item`slick fish meat`,
+      $item`slug of rum`,
+      $item`slug of shochu`,
+      $item`slug of vodka`,
+      $item`soggy seed packet`,
```
spaghetti-squash: use $items, maybe put this list in the pulled taffy libram namespace

packages/garbo/src/tasks/farm/cowo.ts
```diff
@@ -0,0 +1,145 @@
...
+      combat: new GarboStrategy(() => Macro.meatKill()),
+      prepare: () =>
+        !(totalTurnsPlayed() % 11) && meatMood().execute(estimatedGarboTurns()),
+      post: () => {
+        if (!have($effect`Everything looks Beige`)) updateParachuteFailure();
+        completeBarfQuest();
+        trackMarginalMpa();
+      },
+      spendsTurn: true,
+    },
```
spaghetti-squash: could we genericize the barf one as opposed to duplicating?

packages/garbo/src/tasks/farm/cowo.ts
```diff
@@ -0,0 +1,145 @@
...
+    {
+      name: "Cowo",
```
spaghetti-squash:
```suggestion
      name: "Coral Corral",
```

packages/garbo/src/tasks/farm/cowo.ts
```diff
@@ -0,0 +1,145 @@
...
+      },
+      completed: () => myAdventures() === 0,
+      outfit: () => {
+        const outfit = barfOutfit({
+          pants: have($effect`Driving Waterproofly`)
+            ? undefined
+            : $item`really, really nice swimming trunks`,
+          famequip:
+            have($effect`Driving Waterproofly`) ||
+            barfOutfit({}).familiar?.underwater
```
spaghetti-squash: we aren't calling barfOutfit inside a barfOutfit call every turn. Worst case scenario just call barfFamiliar, but I feel like there's a better way

packages/garbo/src/tasks/farm/cowo.ts
```diff
@@ -0,0 +1,145 @@
...
+          AsdonMartin.drive(
+            $effect`Driving Waterproofly`,
+            estimatedGarboTurns(),
+          );
+        }
+      },
+      completed: () => myAdventures() === 0,
+      outfit: () => {
+        const outfit = barfOutfit({
+          pants: have($effect`Driving Waterproofly`)
```
spaghetti-squash: restructure, pull this condition and put it on the whole object in ternary

packages/garbo/src/tasks/farm/cowo.ts
```diff
@@ -0,0 +1,145 @@
...
+            : $item`really, really nice swimming trunks`,
+          famequip:
+            have($effect`Driving Waterproofly`) ||
+            barfOutfit({}).familiar?.underwater
+              ? undefined
+              : $item`das boot`,
+        });
+        const banishMethod = cowoChooseBanish();
+
+        if (banishMethod?.equip) {
```
spaghetti-squash: this needs to get equipped on the spec, so that it takes precedence over the barf outfit itself.

packages/garbo/src/tasks/farm/cowo.ts
```diff
@@ -0,0 +1,145 @@
...
+          outfit.equip(banishMethod.equip);
+        }
+
+        return outfit;
+      },
+      do: $location`The Coral Corral`,
+      combat: new GarboStrategy(() => {
+        const banishMethod = cowoChooseBanish();
```
spaghetti-squash: Use a grimoire context for this shared state

spaghetti-squash: and put the type guard on the engine itself, or in some context-handling method.

packages/garbo/src/tasks/farm/cowo.ts
```diff
@@ -0,0 +1,145 @@
...
+            banishMethod?.macro() ?? Macro.abort(),
+          )
+            .tryItem($item`pulled red taffy`)
+            .meatKill();
+        } else {
+          return Macro.if_(
+            $monsters`Mer-kin rustler, sea cowboy`,
+            banishMethod?.macro() ?? Macro.abort(),
+          ).meatKill();
+        }
```
spaghetti-squash: use externalIf

packages/garbo/src/tasks/farm/cowo.ts
```diff
@@ -0,0 +1,145 @@
...
+      }),
+      post: () => {
+        trackMarginalMpa();
+        postCombatActions();
+
+        const BARF_PLANTS = [
```
spaghetti-squash: 1. not barf
2. if you're doing a SCREAMING_SNAKE_CASE const, put it in as top-level a scope as you can for the file

packages/garbo/src/tasks/farm/cowo.ts
```diff
@@ -0,0 +1,145 @@
...
+        ];
+        if (
+          BARF_PLANTS.some((flower) =>
+            flower.available($location`The Coral Corral`),
+          )
+        ) {
+          BARF_PLANTS.filter((flower) =>
+            flower.available(garboFarmLocation()),
+          ).forEach((flower) => flower.plant());
+        }
```
spaghetti-squash: I think upon rereading this code you will see why I have flagged it.

packages/garbo/src/tasks/farm/cowo.ts
```diff
@@ -0,0 +1,145 @@
...
+
+        if (
+          getCowoMonstersToBanish().includes(toMonster(get("lastEncounter")))
```
spaghetti-squash: I've written guards like this before and found them fragile. I wonder if there's a better way to handle it.

packages/garbo/src/tasks/post/index.ts
```diff
@@ -92,26 +93,32 @@ function useStuff(): GarboPostTask {
...
+  ? [
+      FloristFriar.Crookweed,
+      FloristFriar.ElectricEelgrass,
+      FloristFriar.Duckweed,
+    ]
+  : [
+      FloristFriar.StealingMagnolia,
+      FloristFriar.AloeGuvnor,
+      FloristFriar.PitcherPlant,
+    ];
```
spaghetti-squash: 1. You've written a BARF_PLANTS const in two different places, sounds like it belongs somewhere else merged to both
2. by putting the ternary in the top scope here without any function call protecting it, it gets called before `globalOptions` gets populated

packages/garbo/src/tasks/post/worksheds.ts
```diff
@@ -124,15 +124,21 @@ const worksheds = [
   new GarboWorkshed({
     workshed: $item`Asdon Martin keyfob (on ring)`,
     done: () => {
+      const asdonEffect = globalOptions.cowo
```
spaghetti-squash: I'm starting to wonder if instead of these frequent ternaries and the `garboFarmingLocation` function, we should instead have two big objects--one representing barf, one representing cowo--that globalOptions gives us, based on our target. So rather than this, we'd have `globalOptions.farmZone.asdonEffect` or something like that. Not quite sure what the best move is here.

packages/garbo/src/tasks/dailyFamiliars.ts
```diff
@@ -97,8 +97,15 @@ export function prepRobortender(): void {
...
-        : 0,
+        ? targetMeat() * (0.5 + ((4 + Math.sqrt(110 / 100)) * 30) / 100) +
+          baseMeat() *
+            (0.5 + ((4 + Math.sqrt(110 / 100)) * 30) / 100) *
+            estimatedGarboTurns()
+        : globalOptions.cowo
+          ? baseMeat() *
+            (0.5 + ((4 + Math.sqrt(110 / 100)) * 30) / 100) *
+            estimatedGarboTurns()
+          : 0,
```
spaghetti-squash, line 121: this math has become too complex to legibly inline.

packages/garbo/src/combat.ts
```diff
@@ -496,6 +498,10 @@ export class Macro extends StrictMacro {
       stasisRounds = 20;
     }
 
+    if (globalOptions.cowo) {
+      stasisRounds = 5;
```
spaghetti-squash: I think stasisRounds is probably now long enough to slice out into a function or method that uses `return` rather than `let`

packages/garbo/src/config.ts
```diff
@@ -146,6 +146,11 @@ You can use multiple options in conjunction, e.g. "garbo nobarf ascend"',
       setting: "",
       help: "Print the current version and exit.",
     }),
+    cowo: Args.flag({
+      setting: "garbo_cowo",
+      help: "Flag to run Cowo instead of Barf",
+      default: false,
+    }),
```
spaghetti-squash: I don't love the internal-to-the-code use of the term `cowo`, but more relevantly, I'm not confident that the way we want to access this information in garbo's code is strictly as a boolean.

packages/garbo/src/fights.ts
```diff
@@ -1244,13 +1245,11 @@ const priorityFreeRunFightSources = [
...
       !have($effect`Citizen of a Zone`) &&
-      $locations`Barf Mountain, The Fun-Guy Mansion`.some((l) =>
-        canAdventure(l),
-      ),
+      $locations`Barf Mountain, The Dire Warren`.some((l) => canAdventure(l)),
     (runSource: ActionSource) => {
       const location = canAdventure($location`Barf Mountain`)
         ? $location`Barf Mountain`
-        : $location`The Fun-Guy Mansion`;
+        : $location`The Dire Warren`;
```
spaghetti-squash, line 1252: the reason we used the fun-guy mansion and barf rather than the dire warren is that the peagle can easily kill the monster you're fighting.

packages/garbo/src/fights.ts
```diff
@@ -1267,7 +1266,12 @@ const priorityFreeRunFightSources = [
       },
       location: canAdventure($location`Barf Mountain`)
         ? $location`Barf Mountain`
-        : $location`The Fun-Guy Mansion`,
+        : $location`The Dire Warren`,
+      postTask: () => {
+        if(have($effect`Fishy`, estimatedGarboTurns()) && have($item`Monodent of the Sea`) && !get("_seadentWaveUsed")) {
+          useSkill($skill`Sea *dent: Summon a Wave`)
+        }
```
spaghetti-squash, line 1277: This is really burying the lede in a function that's otherwise been seemingly devoted to peagle-priming. What's the deal?

packages/garbo/src/index.ts
```diff
@@ -104,7 +104,10 @@ import { acquire } from "./acquire";
 const TICKET_MAX_PRICE = 500000;
 
 function ensureBarfAccess() {
-  if (!(get("stenchAirportAlways") || get("_stenchAirportToday"))) {
+  if (
+    !(get("stenchAirportAlways") || get("_stenchAirportToday")) &&
```
spaghetti-squash: there's a libram function for this top check now, realmAvailable

packages/garbo/src/lib.ts
```diff
@@ -164,7 +164,8 @@ export const songboomMeat = () =>
     : 0;
 
 // all tourists have a basemeat of 250
-export const baseMeat = () => 250 + songboomMeat();
+export const baseMeat = () =>
+  globalOptions.cowo ? 300 + songboomMeat() : 250 + songboomMeat();
```
spaghetti-squash: add songboomMeat() just once, outside the ternary

packages/garbo/src/lib.ts
```diff
@@ -681,12 +682,13 @@ const touristFamilyRatio = touristFamilies / barfTourists;
 // 30 tourists till NC, with families counting as 3
 // Estimate number of turns till the counter hits 27
 // then estimate the expected number of turns required to hit a counter of >= 30
-export const turnsToNC =
-  (27 * barfTourists) /
-    (garbageTourists + angryTourists + 3 * touristFamilies) +
-  1 * touristFamilyRatio +
-  2 * (1 - touristFamilyRatio) * touristFamilyRatio +
-  3 * (1 - touristFamilyRatio) * (1 - touristFamilyRatio);
+export const turnsToNC = globalOptions.cowo
```
spaghetti-squash: top scope reference to a globalOptions variable, tale as old as time.

If we have to delay this, which it seems like we do, we should delay the entire calculation, which currently locks into the state you're in when you first start your garbo-run.

packages/garbo/src/lib.ts
```diff
@@ -1242,3 +1244,6 @@ export function mainStatLevel(level: number): number {
 export type RequireAtLeastOne<T, K = keyof T> = K extends keyof T
   ? Partial<T> & { [k in K]: T[K] }
   : never;
+
+export const garboFarmLocation = () =>
```
spaghetti-squash: I think this will be made moot by other restructures, but I don't think this function needs or wants `garbo` as part of its name.

packages/garbo/src/mood.ts
```diff
@@ -74,15 +74,22 @@ export function meatMood(
...
-  mood.skill(
-    urKels
-      ? $skill`Ur-Kel's Aria of Annoyance`
-      : $skill`Fat Leon's Phat Loot Lyric`,
-  );
   mood.skill($skill`Singer's Faithful Ocelot`);
   mood.skill($skill`The Spirit of Taking`);
-  mood.skill($skill`Drescher's Annoying Noise`);
-  mood.skill($skill`Pride of the Puffin`);
+  if (!globalOptions.cowo) {
```
spaghetti-squash: yeah I'm more and more convinced we want a framework closer to what we see in chrono or crimbo than just a flat boolean, despite only having two options. It's just messier this way. Maybe do a  `switch` statement on farming location, for example?

packages/garbo/src/potions.ts
```diff
@@ -187,12 +187,16 @@ export interface PotionOptions {
...
-  "Smithsness",
-  "Item Drop",
-] as const;
+export const VALUABLE_MODIFIERS = () => globalOptions.cowo
+  ? ([
+      "Meat Drop",
+      "Familiar Weight",
+      "Smithsness",
+      "Item Drop",
+      "Familiar Weight Hidden",
```
spaghetti-squash: is it `Familiar Weight Hidden`

packages/garbo/src/potions.ts
```diff
@@ -266,14 +270,20 @@ export class Potion {
...
     );
   }
 
   familiarWeight(): number {
     return (
       this.effectValues?.famWeight ??
-      getModifier("Familiar Weight", this.effect())
+      getModifier("Familiar Weight", this.effect()) +
+        (globalOptions.cowo
+          ? getModifier("Hidden Familiar Weight", this.effect())
```
spaghetti-squash, line 281: or is it `Hidden Familiar Weight`

packages/garbo/src/potions.ts
```diff
@@ -187,12 +187,16 @@ export interface PotionOptions {
   }>;
 }
 
-export const VALUABLE_MODIFIERS = [
-  "Meat Drop",
-  "Familiar Weight",
-  "Smithsness",
-  "Item Drop",
-] as const;
+export const VALUABLE_MODIFIERS = () => globalOptions.cowo
```
spaghetti-squash: doesn't this thang get used in `debuffPlanner`, why don't I see that file changed at all?

packages/garbo/src/tasks/farm/cowo.ts
Ignose: That's what I did originally before the move; I'm actually not convinced we need this for cowo. I'll remove. The benefit for Barf is you hit the NC less; there's no NC for cowo.

spaghetti-squash:
barfFamiliar's outfit planning doesn't seem to be touched virtually at all, I expect it to need a lot of handling for this to function. I also don't see anything for cmc or space jellyfish or several other places where we assume that we're adventuring in barf.

packages/garbo/src/potions.ts
Ignose: it is

packages/garbo/src/potions.ts
Ignose: Idk? I see a difference on line 70?

packages/garbo/src/mood.ts
```diff
@@ -74,15 +73,23 @@ export function meatMood(
...
-    urKels
-      ? $skill`Ur-Kel's Aria of Annoyance`
-      : $skill`Fat Leon's Phat Loot Lyric`,
-  );
   mood.skill($skill`Singer's Faithful Ocelot`);
   mood.skill($skill`The Spirit of Taking`);
-  mood.skill($skill`Drescher's Annoying Noise`);
-  mood.skill($skill`Pride of the Puffin`);
+  if (farmingStrategy().ensureML) {
+    mood.potion($item`How to Avoid Scams`, 3 * baseMeat);
```
Rinn, line 80: Should this be in the ensureML block?

packages/garbo/src/tasks/dailyFamiliars.ts
Rinn: I'd put this in the farmingStrategy() object

packages/garbo/src/tasks/post/index.ts
```diff
@@ -92,26 +94,32 @@ function useStuff(): GarboPostTask {
   };
 }
 
-const BARF_PLANTS = [
-  FloristFriar.StealingMagnolia,
-  FloristFriar.AloeGuvnor,
-  FloristFriar.PitcherPlant,
-];
+const BARF_PLANTS = farmingStrategy().underwater
```
Rinn: I'd move this into the farmingStrategy()

packages/garbo/src/tasks/farm/farmTurn.ts
```diff
@@ -0,0 +1,53 @@
...
+import { GarboTask } from "../engine";
+import {
+  canContinue,
+  shouldCheckParachute,
+  updateParachuteFailure,
+} from "./lib";
+import { farmingStrategy } from "../../lib";
+
+export const BarfTurnQuest: Quest<GarboTask> = {
+  name: "Barf Turn",
```
Rinn: need to rename this to match the name of the file

packages/garbo/src/tasks/farm/nonBarfTurn.ts
Rinn: Why does NonBarfTurnTasks and nonBarfTurns() need to be exported when we already export the quest? I don't see them being called from elsewhere.

packages/garbo/src/familiar/barfFamiliar.ts
```diff
@@ -106,6 +119,7 @@ function getCachedOutfitValues(fam: Familiar) {
     computeBarfOutfit(
       {
         familiar: fam,
+        equip: needsBoot ? $items`das boot` : [],
```
Rinn, line 149: I hate to imagine A Guy, but it's possible someone doesn't have a das boot. imo that should just be a failure condition of trying to opt into sea farming, maybe have a list of required equipment in the farmingStrategy object

packages/garbo/src/familiar/barfFamiliar.ts
```diff
@@ -53,6 +56,14 @@ import { garboValue } from "../garboValue";
 const ITEM_DROP_VALUE = 0.72;
 const MEAT_DROP_VALUE = baseMeat() / 100;
```
Rinn: This isn't safe to be at the module level because baseMeat() calls farmingStrategy() which may not be initialized yet, it'll need to be a function

Rinn (commented):
I tried to run this and got runtime errors, looks like farmingStrategy() is called a number of places at module init and it is still undefined

packages/garbo/src/tasks/farm/nonBarfTurn.ts
Ignose: idk, I didn't add or change this.

packages/garbo/src/mood.ts
Ignose: Yes, fixed

packages/garbo/src/tasks/post/index.ts
Ignose: I don't think it should be. It should be based on whether the farm location is underground/indoors/etc. no?

If we add more additional locations, we'd want to add plant logic for those environment types instead.

packages/garbo/src/tasks/farm/farmTurn.ts
Ignose: Resolved

packages/garbo/src/familiar/barfFamiliar.ts
Ignose: Added more conditions to running cowo; we just default back to barf in the main index.ts if you're A Guy (tm)

packages/garbo/src/tasks/farm/nonBarfTurn.ts
Rinn: kk i'd just remove those exports then if they're unneeded

packages/garbo/src/tasks/post/index.ts
Rinn: that's fine, this will need to be moved out of the module level scope though because farmingStrategy() could be called before it is initialized and be undefined

packages/garbo/src/mood.ts
Ignose: I think I resolved this with recent changes.

packages/garbo/src/config.ts
```diff
@@ -256,6 +266,32 @@ You can use multiple options in conjunction, e.g. "garbo nobarf ascend"',
           help: "At how many minutes before Rollover should we terminate to let you get ready for bed?",
           default: 5,
         }),
+        farmingMethod: Args.custom<FarmingMethod>(
```
Rinn, line 264: The help for this implies it should passed as an array index rather than a string
```
  farmingMethod Farming Method - Select the farming method to use. [default: 0] [setting: garbo_farmingMethod]
    farmingMethod 0 - barf mountain
    farmingMethod 1 - sea cows
```

packages/garbo/src/potions.ts
```diff
@@ -675,7 +677,7 @@ export const pawPotions = Array.from(validPawWishes.keys())
       }),
   );
 
-export const farmingPotions = [
+export const farmingPotions = () => [
```
Rinn, line 686: How often would farmingPotions get called? Since it's not initialized once now it will redo the item.all() filter logic up every call. If this isn't frequent like only on the bulk buffing it shouldn't matter but if this is called every adventure it will probably need to be cached

Ignose: Should just be the once during copyTarget setup?

Rinn (dismissed):

packages/garbo/src/familiar/barfFamiliar.ts
```diff
@@ -49,9 +50,18 @@ import {
 } from "./lib";
 import { meatFamiliar } from "./meatFamiliar";
 import { garboValue } from "../garboValue";
+import { farmingStrategy } from "../farmingStrategy";
 
 const ITEM_DROP_VALUE = 0.72;
```
spaghetti-squash: not blocking but there's no way this value is the same across targets too

packages/garbo/src/resources/banish.ts
```diff
@@ -0,0 +1,130 @@
...
+  [$item`shark cartilage`, 0.03],
+  [$item`slick fish meat`, 0.03],
+  [$item`slug of rum`, 0.03],
+  [$item`slug of shochu`, 0.03],
+  [$item`slug of vodka`, 0.03],
+  [$item`soggy seed packet`, 0.03],
+]);
+
+export function redTaffyWorth(): boolean {
+  const averageRedTaffyValue = [...RED_TAFFY_DROP_WEIGHTS.entries()].reduce(
```
spaghetti-squash: use `sum`

packages/garbo/src/tasks/dailyFamiliars.ts
```diff
@@ -30,42 +30,54 @@ import { meatFamiliar, setBestLeprechaunAsMeatFamiliar } from "../familiar";
...
 import { acquire } from "../acquire";
 import { amuletCoinValue } from "../familiar/lib";
+import { farmingStrategy, garbageTouristRatio } from "../farmingStrategy";
 
 function drivebyValue(targetCount = 0): number {
   const targets = targetCount;
-  const tourists =
-    ((estimatedGarboTurns() - targets) * turnsToNC) / (turnsToNC + 1);
+
+  const tourists = farmingStrategy().accountForNC()
```
spaghetti-squash: I don't think you want tourists set to 0 for cowo?

packages/garbo/src/tasks/dailyFamiliars.ts
spaghetti-squash: idk why this is marked as resolved?

packages/garbo/src/farmingStrategy.ts
```diff
@@ -0,0 +1,173 @@
...
+  touristFamilies = 1,
+  angryTourists = 1;
+const barfTourists = garbageTourists + touristFamilies + angryTourists;
+export const garbageTouristRatio = garbageTourists / barfTourists;
+const touristFamilyRatio = touristFamilies / barfTourists;
+// 30 tourists till NC, with families counting as 3
+// Estimate number of turns till the counter hits 27
+// then estimate the expected number of turns required to hit a counter of >= 30
+
+interface FarmingStrategy {
```
spaghetti-squash:
```suggestion
type FarmingStrategy = {
```

packages/garbo/src/farmingStrategy.ts
```diff
@@ -0,0 +1,173 @@
...
+  stasisRounds(): number;
```
spaghetti-squash: this seems to be a static value, why is this a function?

packages/garbo/src/farmingStrategy.ts
```diff
@@ -0,0 +1,173 @@
...
+  asdonEffect(): Effect;
```
spaghetti-squash: this seems to be a static value, why is this a function?

spaghetti-squash: okay actually an overwhelming majority of these shouldn't be functions.

packages/garbo/src/fights.ts
spaghetti-squash: How is this relevant to this PR?

packages/garbo/src/potions.ts
```diff
@@ -187,12 +187,8 @@ export interface PotionOptions {
   }>;
 }
 
-export const VALUABLE_MODIFIERS = [
-  "Meat Drop",
-  "Familiar Weight",
-  "Smithsness",
-  "Item Drop",
-] as const;
+export const VALUABLE_MODIFIERS = () =>
```
spaghetti-squash: this never changes, why is it a function

spaghetti-squash (changes requested):
Also, I don't see anything done to the outfit caching code in barfFamiliar

packages/garbo/src/familiar/barfFamiliar.ts
Ignose: Genericized this, helps for changes in sniff strategy too.

packages/garbo/src/farmingStrategy.ts
```diff
@@ -0,0 +1,170 @@
...
+  ensureBarfAccess: boolean;
+  baseMeat: number;
+  accountForNC: boolean;
+  turnsToNC(): number;
+  bonusModifiers: Modifier[];
+  location: Location;
+  ensureML: boolean;
+  bonusEffects: Effect[];
+  monstersToBanish: Monster[];
+  targetMonster(): Monster;
```
spaghetti-squash:
```suggestion
  targetMonster: Delayed<Monster>;
```

That way we can have the cowo one be a fixed value.

packages/garbo/src/farmingStrategy.ts
```diff
@@ -0,0 +1,170 @@
...
+      : $monster`garbage tourist`,
+  shouldOlfact: true,
+
+  outfit: () => {
+    const lubing = get("dinseyRollercoasterNext") && have($item`lube-shoes`);
+
+    return barfOutfit(lubing ? { equip: $items`lube-shoes` } : {});
+  },
+
+  combat: () =>
```
spaghetti-squash: I'm not sure I understand why these are a function either, but let me read through

packages/garbo/src/farmingStrategy.ts
```diff
@@ -15,12 +24,55 @@ import {
...
-import { getMonstersToBanish, redTaffyWorth } from "./resources/banish";
 import { completeBarfQuest } from "./resources/realm";
 import { GarboContext } from "./tasks/context";
+import { garboValue } from "./garboValue";
+
+export function getMonstersToBanish(monstersToBanish: Monster[]): Monster[] {
+  return monstersToBanish.filter((monster) => !isBanished(monster));
+}
+
+const RED_TAFFY_DROP_WEIGHTS = new Map<Item, number>([
```
spaghetti-squash: if you bump libram you can import this now

packages/garbo/src/farmingStrategy.ts
```diff
@@ -133,7 +185,10 @@ const THE_CORAL_CORRAL: FarmingStrategy = {
         : macro.meatKill();
     }
 
-    if (!banish && getMonstersToBanish().length > 0) {
+    if (
+      !banish &&
+      getMonstersToBanish($monsters`Mer-kin rustler, sea cowboy`).length > 0
```
spaghetti-squash: i would remove this block and add a throw somewhere to the engine, which I may already have done

packages/garbo/src/lib.ts
```diff
@@ -144,7 +146,19 @@ export function modeValueOfMeat(mode: BonusEquipMode): number {
 }
 
 export function modeValueOfItem(mode: BonusEquipMode): number {
-  return mode === BonusEquipMode.BARF ? 0.72 : 0;
+  const ITEM_DROP_VALUE = () =>
```
spaghetti-squash, line 149: instead of defining this identically across two files, define it in one place and export it

spaghetti-squash (dismissed):

PR #2695, opened 2026-08-13
Maximize MP after PirateRealm setup

### LINKED ISSUE #2652: [FEATURE] Get a higher max MP before buffing
When running piraterealm for cockroaches, garbo leaves piraterealm eyepatch on when buffing up, which ends up causing wasteful MP usage due to the much lower MP cap.

I recommend a maximizer statement before buffing.

bigbongobeats:
Maximize MP after CockroachSetup finishes and pointing to Trash Island before buffing.

This prevents the PirateRealm outfit from leaving the player with a low maximum MP when Garbo begins buffing.

Closes #2652.

### DIFF
--- a/packages/garbo/src/index.ts
+++ b/packages/garbo/src/index.ts
@@ -564,6 +564,7 @@ export function main(argString = ""): void {
           withProperty("removeMalignantEffects", false, () =>
             runGarboQuests([CockroachSetup]),
           );
+	maximize("MP", false);
         }
         // 0. diet stuff.
         if (


Rinn:
It would probably be better to add a task at the end of CockroachSetup which sets the adventuring location to none and does any outfit changes before the quest ends.

Rinn:
https://github.com/loathers/garbage-collector/blob/e9746add5b485c049fce00f3a3f461b14980628d/packages/garbo/src/tasks/cockroach/prep.ts#L293 eyepatch is also already getting removed

bigbongobeats:
While the eyepatch change helped, the lower maxmp still makes eating magical sausages, a common MP restorer, inefficient given how low your stats are post-piraterealm. I'm happy to defer to folks more familiar with the codebase on placement.

Rinn:
there should be a specific location where the sausages are being eaten and it should maximize if needed

bigbongobeats:
Circling back on this, and we can definitely close this one out - it seems like the sausage-wasting behavior I have seen is specific to the giant crab route and not dessert. Will reassess that portion and aim a fix for there.

PR #2698, opened 2026-08-22
Arrrbor Day

Rinn:


Rinn:
I couldn't test this because I got the saplings before I ascended so guess we're waiting 3 months.

PR #2699, opened 2026-08-23
Move a bunch of stuff to garbo-lib that was self-contained

Rinn:


packages/garbo-lib/src/resources/latte.ts
spaghetti-squash, line 1: I think `desirableIngredients` is garbo-opinionated enough that if we're going to start exporting this we might want to let these functions accept args for the ingredients

packages/garbo-lib/src/garboValue.ts
spaghetti-squash, line 1: I don't know that I see the value in pulling this specifically into garbo-lib when `value.ts` already exists. In particular, I think we actively don't want to invoke `garboValue` inside any garbo-lib resource (which doesn't happen in this PR, fwiw), and moving it in here makes that possible.

(The reason we don't want to invoke `garboValue` inside a garbo-lib resource is that the point of having the generic/extensible functions in value.ts is that we can override them or modify them as necessary for another script, and so we want other modules to be able to benefit from that).

I see something about circular imports in the commit history, so maybe this is a necessary, core load-bearing change, but idk that it is

spaghetti-squash: It looks like we pull it in for a single invocation in acquire.ts?

packages/garbo-lib/src/acquire.ts
spaghetti-squash, line 1: I feel like acquire is one of garbo's more footgun-y functions, and I'm surprised at the desire to use it outside garbo, but not opposed to making it accessible, especially if that means more eyes on it to make it less brittle.

Rinn:
This is all fair, i'll reevaluate what im doing here late

PR #2703, opened 2026-09-01
Add cup of 13s as a drops gear

Ignose:


Rinn (dismissed):

packages/garbo/src/outfit/dropsgear.ts
```diff
@@ -525,3 +527,36 @@ export function toyCupidBow(familiar: Familiar): Map<Item, number> {
...
+function cupOfThirteens(): Map<Item, number> {
+  if (!CupOfThirteens.have()) {
+    return new Map();
+  }
+
+  const dropsToday = get("_cupOf13sDrops");
+
+  // A drop occurs at 6 charges if we haven't gotten 3 drops today.
+  // Otherwise, the next drop occurs at 10 charges.
+  const chargeRequired = dropsToday < 3 ? 6 : 10;
```
Rinn: wait the wiki says first drop is 6, the rest are 10

Rinn: is the wiki wrong?

Ignose: Nas said it's the first two drops; but checking my logs it's definitely only the first drop. Fixing.

packages/garbo/src/outfit/dropsgear.ts
```diff
@@ -525,3 +527,36 @@ export function toyCupidBow(familiar: Familiar): Map<Item, number> {
...
+      : dropsToday <= 3
+        ? ["Awesome", "good"]
+        : dropsToday <= 5
+          ? ["good", "decent"]
+          : ["decent", "crappy"];
+
+  const possibleDrops = Item.all().filter(
+    (item) =>
+      item.tradeable &&
+      item.discardable &&
```
spaghetti-squash: does one of these conditions filter out speakeasy drinks?

packages/garbo/src/outfit/dropsgear.ts
```diff
@@ -525,3 +527,36 @@ export function toyCupidBow(familiar: Familiar): Map<Item, number> {
...
+  const qualities =
+    dropsToday <= 1
+      ? ["EPIC", "Awesome"]
+      : dropsToday <= 3
+        ? ["Awesome", "good"]
+        : dropsToday <= 5
+          ? ["good", "decent"]
+          : ["decent", "crappy"];
+
+  const possibleDrops = Item.all().filter(
```
spaghetti-squash: Typically with sums this large we cache the value, because we don't want to iterate over Item.all() too many times per turn. But maybe that's dumb?

packages/garbo/src/outfit/dropsgear.ts
Ignose: No, that's new information we got from CF today. It was previously unknown.

I can add that, though.

packages/garbo/src/outfit/dropsgear.ts
```diff
@@ -525,3 +536,39 @@ export function toyCupidBow(familiar: Familiar): Map<Item, number> {
...
+    (mode !== BonusEquipMode.BARF && mode !== BonusEquipMode.FREE)
+  ) {
+    return new Map();
+  }
+
+  const dropsToday = get("_cupOf13sDrops");
+
+  // A drop occurs at 6 charges if we haven't gotten a drop today.
+  // Otherwise, the next drop occurs at 10 charges.
+  const chargeRequired = dropsToday < 2 ? 6 : 10;
```
Rinn:
```suggestion
  const chargeRequired = dropsToday === 0 ? 6 : 10;
```

packages/garbo/src/outfit/dropsgear.ts
```diff
@@ -525,3 +536,39 @@ export function toyCupidBow(familiar: Familiar): Map<Item, number> {
...
+          ? ["good", "decent"]
+          : ["decent", "crappy"];
+
+  const possibleDrops = CUP_OF_THIRTEENS_DROPS.filter((item) =>
+    qualities.includes(item.quality),
+  );
+
+  const cupBonus = garboAverageValue(...possibleDrops) / chargeRequired;
+
+  return new Map([[$item`Cup of 13s`, cupBonus]]);
```
Rinn: could cache with
```
let cupOfThirteensBonus: [string, number] | undefined;
```

and do something like

```suggestion
  if (cupOfThirteensBonus?.[0] !== qualities[0]) {
    const possibleDrops = CUP_OF_THIRTEENS_DROPS.filter((item) =>
      qualities.includes(item.quality),
    );
    cupOfThirteensBonus = [qualities[0], garboAverageValue(...possibleDrops)];
  }

  const cupBonus = cupOfThirteensBonus[1] / chargeRequired;
  return new Map([[$item`Cup of 13s`, cupBonus]]);
```

Ignose: What's the benefit to changing how this is cached? Just curious.

Rinn: just prevents looping through the CUP_OF_THIRTEENS_DROPS to filter it and generate the vaule every time this function is called, which I think is between every adventure

packages/garbo/src/outfit/dropsgear.ts
```diff
@@ -525,3 +536,43 @@ export function toyCupidBow(familiar: Familiar): Map<Item, number> {
...
+  // A drop occurs at 6 charges if we haven't gotten a drop today.
+  // Otherwise, the next drop occurs at 10 charges.
+  const chargeRequired = dropsToday === 0 ? 6 : 10;
+
+  const qualities =
+    dropsToday <= 1
+      ? ["EPIC", "Awesome"]
+      : dropsToday <= 3
+        ? ["Awesome", "good"]
+        : dropsToday <= 5
```
Rinn: Just checked the mafia data files and awesome is lowercase
```suggestion
      ? ["EPIC", "awesome"]
      : dropsToday <= 3
        ? ["awesome", "good"]
        : dropsToday <= 5
```

Rinn (changes requested):

Rinn (approved):

PR #2705, opened 2026-09-02
If we can predict the next fruit, just value that, otherwise use the …

Ignose:
…statistical variant


packages/garbo/src/outfit/dropsgearAccessories.ts
```diff
@@ -148,6 +150,17 @@ function cinchoDeMayo(mode: BonusEquipMode) {
 }
 
 function calculateLaughingStockBonus() {
+  const allFruit = LaughingStock.expectedDropsToday(estimatedGarboTurns());
+
+  if (allFruit) {
+    const nextFruit = allFruit[0][1];
+
+    const averageHorizon =
+      allFruit.reduce((sum, [, fights]) => sum + fights, 0) / allFruit.length;
```
spaghetti-squash: use `sum`

packages/garbo/src/outfit/dropsgearAccessories.ts
```diff
@@ -148,6 +150,17 @@ function cinchoDeMayo(mode: BonusEquipMode) {
 }
 
 function calculateLaughingStockBonus() {
+  const allFruit = LaughingStock.expectedDropsToday(estimatedGarboTurns());
```
spaghetti-squash: I think we probably want to cache this somehow, so we aren't predicting the full day ahead ever turn

packages/garbo/src/outfit/dropsgearAccessories.ts
```diff
@@ -148,6 +150,17 @@ function cinchoDeMayo(mode: BonusEquipMode) {
...
+
+  if (allFruit) {
+    const nextFruit = allFruit[0][1];
```
spaghetti-squash: are we really returning tuples in this function? That's terrible, we shouldn't be doing that, we should be returning something useful and good instead.

packages/garbo/src/outfit/dropsgearAccessories.ts
```diff
@@ -148,6 +150,17 @@ function cinchoDeMayo(mode: BonusEquipMode) {
...
+
+    const averageHorizon =
+      allFruit.reduce((sum, [, fights]) => sum + fights, 0) / allFruit.length;
+
+    return Math.max(nextFruit, averageHorizon);
```
spaghetti-squash: I don't know that this is actually the value you want. This doesn't include any garboValue calls in its computation.

packages/garbo/src/outfit/dropsgearAccessories.ts
Ignose: I am using sum?

packages/garbo/src/outfit/dropsgearAccessories.ts
Ignose: This comment makes me laugh for several reasons; thank you for bringing me joy on this day.

Yes, we're returning tuples.

packages/garbo/src/outfit/dropsgearAccessories.ts
spaghetti-squash: use the `sum` utility function used all over the codebase

packages/garbo/src/outfit/dropsgearAccessories.ts
```diff
@@ -147,7 +149,23 @@ function cinchoDeMayo(mode: BonusEquipMode) {
   return new Map<Item, number>([[$item`Cincho de Mayo`, 3 * felizValue()]]);
 }
 
+const allFruit = LaughingStock.expectedDropsToday(estimatedGarboTurns());
```
spaghetti-squash: we don't want to call this at top-level

PR #2709, opened 2026-09-03
Ranch refurbishing

Shiverwarp:
General thoughts while I was doing this:

- Are we running underwater potions during embezzlers/copy targets? What about meat mood? We don't want to run pressure penalty stuff here even if we're flooding because the pressure penalty is already 0

- Pill keeper opportunity cost? Is that old yachtzee?

-For enabling underwater wandering, how can we check whether our farming strategy is underwater from within wanderer?

- Are we properly setting location before maximizing? Are we just purely relying on +sea? This is important for wandering probably

- Some way to value resistances while wandering for pearls?

- What happened to our dinseyRollercoasterNext check? I can't seem to find it anymore

- Is there anything we might want to do during nobarf if we do have fishy, but our farming strategy isn't cows?

- Can we better integrate the combat NC forcers for yachtzee somehow with farmingStrategy? can't use farmingStrategy().combat to put something after the skill usage

How can I make it so VALUABLE_MODIFIERS only includes Hidden Familiar Weight if we're going to go underwater?

### DIFF
--- a/packages/garbo-lib/src/wanderer/lib.ts
+++ b/packages/garbo-lib/src/wanderer/lib.ts
@@ -141,9 +141,9 @@ const ILLEGAL_PARENTS = [
   "A Monorail Station",
   "Memories",
 ];
-const ILLEGAL_ZONES = ["The Drip", "Suburbs"];
+const ILLEGAL_ZONES = ["The Drip", "Suburbs", "The Mer-Kin Deepcity"];
 const canAdventureOrUnlockSkipList = [
-  ...$locations`The Bubblin' Caldera, Barrrney's Barrr, The F'c'le, The Poop Deck, Belowdecks, The Secret Government Laboratory, The Dire Warren, Inside the Palindome, The Haiku Dungeon, An Incredibly Strange Place (Bad Trip), An Incredibly Strange Place (Mediocre Trip), An Incredibly Strange Place (Great Trip), El Vibrato Island, The Daily Dungeon, Trick-or-Treating, Seaside Megalopolis, The Orcish Frat House, Through the Spacegate, Mt. Molehill`,
+  ...$locations`The Skate Park, The Bubblin' Caldera, Barrrney's Barrr, The F'c'le, The Poop Deck, Belowdecks, The Secret Government Laboratory, The Dire Warren, Inside the Palindome, The Haiku Dungeon, An Incredibly Strange Place (Bad Trip), An Incredibly Strange Place (Mediocre Trip), An Incredibly Strange Place (Great Trip), El Vibrato Island, The Daily Dungeon, Trick-or-Treating, Seaside Megalopolis, The Orcish Frat House, Through the Spacegate, Mt. Molehill`,
   ...Location.all().filter(
     ({ parent, zone }) =>
       ILLEGAL_PARENTS.includes(parent) || ILLEGAL_ZONES.includes(zone),
@@ -188,7 +188,7 @@ export function canAdventureOrUnlock(
       (z) => loc.zone === z.zone && (z.available() || !z.noInv),
     );
   return (
-    !underwater(loc) &&
+    (!underwater(loc) || have($effect`Fishy`)) && // Can we check from wanderer somehow whether our farming strategy is an underwater zone?
     !skiplist.includes(loc) &&
     (canAdventure(loc) || canUnlock)
   );
--- a/packages/garbo-lib/src/wanderer/lib.ts
+++ b/packages/garbo-lib/src/wanderer/lib.ts
@@ -70,6 +70,7 @@ export type WandererFactoryOptions = {
   valueOfAdventure?: number;
   takeTurnForProfit?: boolean;
   canRefractedGaze?: boolean;
+  underwaterAllowed?: boolean;
 };
 
 export type WandererFactory = (
@@ -152,6 +153,7 @@ const canAdventureOrUnlockSkipList = [
 export function canAdventureOrUnlock(
   loc: Location,
   includeUnlockable = true,
+  underwaterAllowed = false,
 ): boolean {
   const skiplist = [...canAdventureOrUnlockSkipList];
 
@@ -188,7 +190,7 @@ export function canAdventureOrUnlock(
       (z) => loc.zone === z.zone && (z.available() || !z.noInv),
     );
   return (
-    (!underwater(loc) || have($effect`Fishy`)) && // Can we check from wanderer somehow whether our farming strategy is an underwater zone?
+    (!underwater(loc) || (have($effect`Fishy`) && underwaterAllowed)) &&
     !skiplist.includes(loc) &&
     (canAdventure(loc) || canUnlock)
   );
--- a/packages/garbo-lib/src/wanderer/pearl.ts
+++ b/packages/garbo-lib/src/wanderer/pearl.ts
@@ -5,12 +5,21 @@ import {
   WandererTarget,
   wandererTurnsAvailableToday,
 } from "./lib";
-import { $effect, $element, $item, $location, get, have } from "libram";
+import {
+  $effect,
+  $element,
+  $item,
+  $location,
+  BooleanProperty,
+  get,
+  have,
+  NumericProperty,
+} from "libram";
 
 type PearlTarget = {
   location: Location;
-  completionPref: string;
-  progressPref: string;
+  completionPref: BooleanProperty;
+  progressPref: NumericProperty;
   estimatedProgress: number;
   maximizeElement: Element;
 };
--- a/packages/garbo-lib/src/wanderer/pearl.ts
+++ b/packages/garbo-lib/src/wanderer/pearl.ts
@@ -1,11 +1,11 @@
-import { Element, Location, numericModifier } from "kolmafia";
+import { Element, Location } from "kolmafia";
 import {
   DraggableFight,
   WandererFactoryOptions,
   WandererTarget,
   wandererTurnsAvailableToday,
 } from "./lib";
-import { $effect, $element, $item, $location, $modifier, get, have } from "libram";
+import { $effect, $element, $item, $location, get, have } from "libram";
 
 type PearlTarget = {
   location: Location;
@@ -22,40 +22,40 @@ const PearlTargets: PearlTarget[] = [
     completionPref: "_unblemishedPearlTheBriniestDeepests",
     progressPref: "_unblemishedPearlTheBriniestDeepestsProgress",
     estimatedProgress: 3.3,
-    maximizeElement: $element`Cold`
+    maximizeElement: $element`Cold`,
   },
   {
     location: $location`Madness Reef`,
     completionPref: "_unblemishedPearlMadnessReef",
     progressPref: "_unblemishedPearlMadnessReefProgress",
     estimatedProgress: 3.3,
-    maximizeElement: $element`Stench`
+    maximizeElement: $element`Stench`,
   },
   {
     location: $location`Anemone Mine`,
     completionPref: "_unblemishedPearlAnemoneMine",
     progressPref: "_unblemishedPearlAnemoneMineProgress",
     estimatedProgress: 3.3,
-    maximizeElement: $element`Spooky`
+    maximizeElement: $element`Spooky`,
   },
   {
     location: $location`The Dive Bar`,
     completionPref: "_unblemishedPearlDiveBar",
     progressPref: "_unblemishedPearlDiveBarProgress",
     estimatedProgress: 3.3,
-    maximizeElement: $element`Sleaze`
+    maximizeElement: $element`Sleaze`,
   },
   {
     location: $location`The Marinara Trench`,
     completionPref: "_unblemishedPearlMarinaraTrench",
     progressPref: "_unblemishedPearlMarinaraTrenchProgress",
     estimatedProgress: 3.3,
-    maximizeElement: $element`Hot`
+    maximizeElement: $element`Hot`,
   },
 ];
 
 // We need to throw this to the maximizer somehow so that we properly value elemental bonuses
-function elementalBonus(element: Element): number {
+/* function elementalBonus(element: Element): number {
   const resistance = numericModifier(
     $modifier`${element.toString()} Resistance`
   );
@@ -67,7 +67,7 @@ function elementalBonus(element: Element): number {
     Math.ceil(1 / progress(resistance));
 
   return get("valueOfAdventure", 4000) * (turns(resistance) - turns(resistance + 1));
-}
+} */
 
 function turnsRemainingToComplete(
   progressPref: string,
--- a/packages/garbo-lib/src/wanderer/pearl.ts
+++ b/packages/garbo-lib/src/wanderer/pearl.ts
@@ -1,0 +1,82 @@
+import { Location } from "kolmafia";
+import {
+  DraggableFight,
+  WandererFactoryOptions,
+  WandererTarget,
+  wandererTurnsAvailableToday,
+} from "./lib";
+import { $effect, $item, $location, get, have } from "libram";
+
+type PearlTarget = {
+  location: Location;
+  completionPref: string;
+  progressPref: string;
+  estimatedProgress: number;
+};
+const PearlTargets: PearlTarget[] = [
+  // estimating progress at some lower rates, 3.3 is with just passives.
+  // TODO value resistances when targeting these, maybe make a pref or something that records our expected amount?
+  {
+    location: $location`The Briniest Deepests`,
+    completionPref: "_unblemishedPearlTheBriniestDeepests",
+    progressPref: "_unblemishedPearlTheBriniestDeepestsProgress",
+    estimatedProgress: 3.3,
+  },
+  {
+    location: $location`Madness Reef`,
+    completionPref: "_unblemishedPearlMadnessReef",
+    progressPref: "_unblemishedPearlMadnessReefProgress",
+    estimatedProgress: 3.3,
+  },
+  {
+    location: $location`Anemone Mine`,
+    completionPref: "_unblemishedPearlAnemoneMine",
+    progressPref: "_unblemishedPearlAnemoneMineProgress",
+    estimatedProgress: 3.3,
+  },
+  {
+    location: $location`The Dive Bar`,
+    completionPref: "_unblemishedPearlDiveBar",
+    progressPref: "_unblemishedPearlDiveBarProgress",
+    estimatedProgress: 3.3,
+  },
+  {
+    location: $location`The Marinara Trench`,
+    completionPref: "_unblemishedPearlMarinaraTrench",
+    progressPref: "_unblemishedPearlMarinaraTrenchProgress",
+    estimatedProgress: 3.3,
+  },
+];
+
+function turnsRemainingToComplete(
+  progressPref: string,
+  estimatedProgress: number,
+) {
+  const progressRemaining = 100 - get(progressPref, 0);
+  return Math.ceil(progressRemaining / estimatedProgress);
+}
+
+export function pearlFactory(
+  type: DraggableFight,
+  _locationSkiplist: Location[],
+  options: WandererFactoryOptions,
+): WandererTarget[] {
+  if (have($effect`Fishy`) && type !== "freerun") {
+    return PearlTargets.filter(
+      (t) =>
+        !get(t.completionPref) &&
+        wandererTurnsAvailableToday(options, t.location, true) >=
+          turnsRemainingToComplete(t.progressPref, t.estimatedProgress),
+    ).map(
+      (t) =>
+        new WandererTarget({
+          name: `${t.location} Pearl`,
+          location: t.location,
+          zoneValue:
+            options.itemValue($item`unblemished pearl`) *
+            (t.estimatedProgress / 100),
+        }),
+    );
+  }
+  return [];
+}
--- a/packages/garbo-lib/src/wanderer/pearl.ts
+++ b/packages/garbo-lib/src/wanderer/pearl.ts
@@ -1,17 +1,18 @@
-import { Location } from "kolmafia";
+import { Element, Location, numericModifier } from "kolmafia";
 import {
   DraggableFight,
   WandererFactoryOptions,
   WandererTarget,
   wandererTurnsAvailableToday,
 } from "./lib";
-import { $effect, $item, $location, get, have } from "libram";
+import { $effect, $element, $item, $location, $modifier, get, have } from "libram";
 
 type PearlTarget = {
   location: Location;
   completionPref: string;
   progressPref: string;
   estimatedProgress: number;
+  maximizeElement: Element;
 };
 const PearlTargets: PearlTarget[] = [
   // estimating progress at some lower rates, 3.3 is with just passives.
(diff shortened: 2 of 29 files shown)


packages/garbo-lib/src/wanderer/lib.ts
```diff
@@ -188,7 +188,7 @@ export function canAdventureOrUnlock(
       (z) => loc.zone === z.zone && (z.available() || !z.noInv),
     );
   return (
-    !underwater(loc) &&
+    (!underwater(loc) || have($effect`Fishy`)) && // Can we check from wanderer somehow whether our farming strategy is an underwater zone?
```
spaghetti-squash: right now we can't, but we can add a like `underwaterAllowed` arg to it?

packages/garbo-lib/src/wanderer/pearl.ts
```diff
@@ -0,0 +1,82 @@
...
+  DraggableFight,
+  WandererFactoryOptions,
+  WandererTarget,
+  wandererTurnsAvailableToday,
+} from "./lib";
+import { $effect, $item, $location, get, have } from "libram";
+
+type PearlTarget = {
+  location: Location;
+  completionPref: string;
```
spaghetti-squash: use some sort of property type for this rather than string. Are these all `StringProperty`s?

spaghetti-squash: looks like it should be `BooleanProperty`, based on some code later.

packages/garbo-lib/src/wanderer/yachtzee.ts
```diff
@@ -0,0 +1,48 @@
...
+    get("encountersUntilYachtzeeChoice") !== 0 &&
+    [
+      "backup",
+      "wanderer",
+      "yellow ray",
+      "freefight",
+      "freerun",
+      "conditional freefight",
+      "freefight (no items)",
+    ].includes(type) &&
```
spaghetti-squash, line 27: this is a really long list. It might be worth it to start tracking properties of the different fighttypes somewhere, as opposed to relying on these bare lists.

packages/garbo/src/resources/seaDent.ts
```diff
@@ -0,0 +1,30 @@
...
+  get,
+  have,
+  withChoice,
+} from "libram";
+import { copyTargetCount } from "../target";
+import { farmingStrategy } from "../farmingStrategy";
+
+export function waveDireWarren() {
+  if (
+    farmingStrategy().location === $location`The Coral Corral` &&
```
spaghetti-squash: i think most of these checks should actually check for whether the location is underwater

packages/garbo/src/resources/seaDent.ts
```diff
@@ -0,0 +1,30 @@
...
+    have($item`Monodent of the Sea`) &&
+    haveEffect($effect`Fishy`) > copyTargetCount() + 50 && // Let's be very careful
+    !get("_seadentWaveUsed") &&
+    get("_seadentWaveZone") !== "The Dire Warren" &&
```
spaghetti-squash, line 20: we should let libram know this is a LocationProperty

packages/garbo/src/tasks/yachtzee/index.ts
```diff
@@ -92,6 +97,47 @@ export const yachtzeeQuest: Quest<AlternateTask, GarboContext>[] = [
...
+        name: "Parka Spikes NC Forcer",
+        completed: () => get("noncombatForcerActive"),
+        ready: () =>
+          have($item`Jurassic Parka`) && get("_spikolodonSpikeUses") < 5,
+        outfit: () =>
+          barfOutfit({
+            equip: $items`Jurassic Parka`,
+            modes: { parka: "spikolodon" },
+          }),
+        do: farmingStrategy().location,
```
spaghetti-squash: I assume that the yachtzeeQuest function gets called after farmingStrategy() is initialized, but it still gives me the heebie jeebies not to do `() => farmingStrategy().location`

packages/garbo/src/tasks/yachtzee/index.ts
```diff
@@ -92,6 +97,47 @@ export const yachtzeeQuest: Quest<AlternateTask, GarboContext>[] = [
         sobriety: () => (willDrunkAdventure() ? "drunk" : "sober"),
         spendsTurn: false,
       },
+      {
+        name: "Parka Spikes NC Forcer",
+        completed: () => get("noncombatForcerActive"),
+        ready: () =>
+          have($item`Jurassic Parka`) && get("_spikolodonSpikeUses") < 5,
+        outfit: () =>
```
spaghetti-squash: we also want to do the same context stuff as the existing farm turns although maybe I wrote the code to make that automatically wrap our outfit? idk.

packages/garbo/src/farmingStrategy.ts
```diff
@@ -145,20 +145,35 @@ const THE_CORAL_CORRAL: FarmingStrategy = {
   },
 
   combat: new GarboStrategy(({ banish }) => {
+    const baseMacro = Macro.externalIf(
+      !get("seahorseName"),
+      Macro.if_(
+        $monster`wild seahorse`,
+        Macro.item($item`sea cowbell`)
+          .item($item`sea cowbell`)
```
spaghetti-squash, line 252: are we ensured to have these?

packages/garbo/src/farmingStrategy.ts
```diff
@@ -145,20 +145,35 @@ const THE_CORAL_CORRAL: FarmingStrategy = {
...
-        ? macro.tryItem($item`pulled red taffy`).meatKill()
-        : macro.meatKill();
+        ? banishMacro.tryItem($item`pulled red taffy`).meatKill()
+        : banishMacro.meatKill();
     }
 
     return redTaffyWorth()
-      ? Macro.tryItem($item`pulled red taffy`).meatKill()
-      : Macro.meatKill();
+      ? baseMacro.tryItem($item`pulled red taffy`).meatKill()
```
spaghetti-squash: so it's worth noting that all of these macro methods are mutative. I think the code would look a little cleaner if instead of doing a bunch of ternary here, we do a bunch of mutations on the base object and then return it at the end.

so like

```
const macro = stuff();

if (condition) macro.addMoreSteps();
if (otherCondition) macro.addOtherSteps();

return macro;
```

packages/garbo/src/potions.ts
```diff
@@ -192,6 +192,7 @@ export const VALUABLE_MODIFIERS = [
   "Familiar Weight",
   "Smithsness",
   "Item Drop",
+  "Hidden Familiar Weight", // How can I add this only if farming strategy is underwater? It's a const
```
spaghetti-squash: the reason this is a top level const with an `as const` is type-safety. I think this will be easier when I refactor farming strategies. I'd probably omit this for now?

packages/garbo/src/farmingStrategy.ts
Shiverwarp: We can check whether we have a seahorse name already, and then acquire these if we don't. Where would be the best place to do so? Same place we buy our emergency banishers?

packages/garbo-lib/src/wanderer/pearl.ts
```diff
@@ -0,0 +1,82 @@
...
+  progressPref: string;
+  estimatedProgress: number;
```
Ignose: `maximizeElement`: Element;

packages/garbo-lib/src/wanderer/pearl.ts
```diff
@@ -0,0 +1,82 @@
...
+];
+
+function turnsRemainingToComplete(
+  progressPref: string,
+  estimatedProgress: number,
+) {
+  const progressRemaining = 100 - get(progressPref, 0);
+  return Math.ceil(progressRemaining / estimatedProgress);
+}
+
```
Ignose: function to value more elemental res goes here, I guess, for the wanderer maximizer.

packages/garbo/src/farmingStrategy.ts
Shiverwarp: Yeah this was the way I wanted to do it, wasn't sure if the current ternary and such was on purpose

packages/garbo/src/outfit/dropsgearAccessories.ts
```diff
@@ -139,8 +140,9 @@ function cinchoDeMayo(mode: BonusEquipMode) {
     !monsterManuelAvailable() ||
     // If we're doing Yachtzees, only use up excess cincho.
     maximumPinataCasts() <= 0 ||
-    // If we have more than 50 passive damage, we'll never be able to cast projectile pinata without risking the monster dying
-    maxPassiveDamage() >= 50
+    // If we have more than 50 passive damage, we'll never be able to cast projectile pinata without risking the monster dying. Cows are tankier though.
+    maxPassiveDamage() >=
+      (farmingStrategy().location === $location`The Coral Corral` ? 150 : 50)
```
Ignose: We should probably tie this to the target monster defined by farmingStrategy, that way we can adjust to stronger or weaker targets using some predefined function.

packages/garbo/src/resources/aprilingband.ts
```diff
@@ -33,6 +35,14 @@ const instruments: {
     instrument: "Apriling band saxophone",
     value: () => getBestLuckyAdventure().value() * 3,
   },
+  {
+    instrument: "Apriling band tuba",
+    value: () =>
+      realmAvailable("sleaze") &&
+      farmingStrategy().location === $location`The Coral Corral`
+        ? (20000 - get("valueOfAdventure")) * 3 // Yachtzee
+        : 0, // Are there any other valuable NCs?
```
Ignose: shadow waters buff

packages/garbo/src/potions.ts
Ignose: const = () => whatever

Then change the import over on the cockroach side.

The thing is though, I didn't see this mattering; there weren't any hidden familiar weight and also stat buffs. Did I miss some?

spaghetti-squash:
> Are we running underwater potions during embezzlers/copy targets? What about meat mood? We don't want to run pressure penalty stuff here even if we're flooding because the pressure penalty is already 0

I haven't actually run cowo at all, so this is something someone who's running it should do a log check for. I would assume the answer is "no", but the hidden familiar weight thing is tricky. I also assume that 'hidden familiar weight' isn't strictly 'familiar weight at sea', which might be tricky.

> Pill keeper opportunity cost? Is that old yachtzee?

yes, although the premise of it still stands: we can use our free pillkeeper for an nc-force, an embezzler, or a potion-doubler.

> For enabling underwater wandering, how can we check whether our farming strategy is underwater from within wanderer?

We'd need either to reference something shared, like you're doing with fishy right now, or add an `allowedUnderwater: () => boolean` to the wanderer constructor and pass it on in.

> Are we properly setting location before maximizing? Are we just purely relying on +sea? This is important for wandering probably

I don't believe so. if anything, we're doing the opposite. I would wrap `super.dress` in `withLocation` maybe? Or just add it to each of the functions. dealer's choice.

> What happened to our dinseyRollercoasterNext check? I can't seem to find it anymore

searching for `dinseyRollercoasterNext` in the codebase still yields a lube shoes thing in the farming strategy. Not sure if there was a secondary check you're talking about.

> Is there anything we might want to do during nobarf if we do have fishy, but our farming strategy isn't cows?

I would argue yachtzee iff we're doing meat copies, but maybe taht's silly

> Can we better integrate the combat NC forcers for yachtzee somehow with farmingStrategy? can't use farmingStrategy().combat to put something after the skill usage

Yes, but I plan to refactor farming strategy soon:tm:, so I wouldn't sweat it for the immediate future. I'll try to get a draft up right away with the shape of it.

> How can I make it so VALUABLE_MODIFIERS only includes Hidden Familiar Weight if we're going to go underwater?

It can't be a top-level array anymore, which is fine; the `as const` is there for type safety, ensuring that stuff like the `isValuable` method knows that it's all numeric modifiers. I think this can be part of the farming strategy refactor, I'm going to give a lot of stuff that gently varies a default value and bonus values and we'll access stuff through either getter methods or regular methods that act as getters.

Shiverwarp:
For dinseyRollercoasterNext we used to have this in the prepare, which would mean we only run the meat mood if we're going to be fighting a monster, not riding the coaster

```     prepare: () =>
        !get("dinseyRollercoasterNext") &&
        !(totalTurnsPlayed() % 11) &&
        meatMood().execute(estimatedGarboTurns()), ```

File: packages/garbo-lib/src/wanderer/yachtzee.ts (hunk 3 above)
Comment (Shiverwarp) [reply packages/garbo-lib/src/wanderer/yachtzee.ts:27]: Could you expand on what you mean? Are we not already tracking these? Or do you mean to organize them in some new way?

File: packages/garbo/src/potions.ts (hunk 10 above)
Comment (Shiverwarp) [reply packages/garbo/src/potions.ts]: I think I'll need some nudging for how farming strategies might allow me to modify this

**Comment by Shiverwarp** [comment]
Alright I think I've addressed the comments I can so far, things left are addressing the location setting issue so that maximizer isn't equipping underwater only stuff on land, etc

And try to look into when and how we're using potions

File: packages/garbo-lib/src/wanderer/pearl.ts [hunk 15]
```diff
@@ -0,0 +1,112 @@
...
+  {
+    location: $location`The Marinara Trench`,
+    completionPref: "_unblemishedPearlMarinaraTrench",
+    progressPref: "_unblemishedPearlMarinaraTrenchProgress",
+    estimatedProgress: 3.3,
+    maximizeElement: $element`Hot`,
+  },
+];
+
+// We need to throw this to the maximizer somehow so that we properly value elemental bonuses
```
spaghetti-squash, line 66: yeah we don't yet have a framework for assessing the "cost" of outfit onuses in here. I think for stuff like the 8-bit realm we just assume your current value is correct.

packages/garbo-lib/src/wanderer/pearl.ts
```diff
@@ -0,0 +1,112 @@
...
+export function pearlFactory(
+  type: DraggableFight,
+  _locationSkiplist: Location[],
+  options: WandererFactoryOptions,
+): WandererTarget[] {
+  if (have($effect`Fishy`) && type !== "freerun") {
+    return PearlTargets.filter(
+      (t) =>
+        !get(t.completionPref) &&
+        wandererTurnsAvailableToday(options, t.location, true) >=
```
spaghetti-squash, line 98: technically, we don't strictly need this check if we aren't `ascending`, but I don't know that that actually makes sense to implement

packages/garbo/src/outfit/dropsgearAccessories.ts
```diff
@@ -139,8 +139,8 @@ function cinchoDeMayo(mode: BonusEquipMode) {
     !monsterManuelAvailable() ||
     // If we're doing Yachtzees, only use up excess cincho.
     maximumPinataCasts() <= 0 ||
-    // If we have more than 50 passive damage, we'll never be able to cast projectile pinata without risking the monster dying
-    maxPassiveDamage() >= 50
+    // If we have more than 50 passive damage, we'll never be able to cast projectile pinata without risking the monster dying. Cows are tankier though.
+    maxPassiveDamage() >= (FarmingStrategy.isUnderwater() ? 150 : 50)
```
spaghetti-squash, line 143: I think rather than checking which `FarmingStrategy` you're on, we should have this value be on the `FarmingStrategyOptions` associated with the strategy. Not sure if 50 should be a default with 150 as the override, or if it should always be supplied. If we can somehow compute this value from the zone, or the targetMonster, that's even better (but should be a method on the class instead)

packages/garbo/src/resources/aprilingband.ts
```diff
@@ -33,6 +34,13 @@ const instruments: {
     instrument: "Apriling band saxophone",
     value: () => getBestLuckyAdventure().value() * 3,
   },
+  {
+    instrument: "Apriling band tuba",
+    value: () =>
+      realmAvailable("sleaze") && FarmingStrategy.isUnderwater()
+        ? (20000 - get("valueOfAdventure")) * 3 // Yachtzee
```
spaghetti-squash, line 41: this made me look at pillkeeperopportunitycost and discover we're still valuing yachtzees at 40k there lol

packages/garbo/src/resources/seaDent.ts
```diff
@@ -0,0 +1,30 @@
...
+import { FarmingStrategy } from "../farmingStrategy";
+
+export function waveDireWarren() {
+  if (
+    FarmingStrategy.isUnderwater() &&
+    have($item`Monodent of the Sea`) &&
+    haveEffect($effect`Fishy`) > copyTargetCount() + 50 && // Let's be very careful
+    !get("_seadentWaveUsed") &&
+    get("_seadentWaveZone") !== "The Dire Warren" &&
+    get("lastAdventure") === $location`The Dire Warren`
```
spaghetti-squash, line 21: Do we ever go to the dire warren outside of copy targets? Does this update the `$location.environment`, to keep wanderer stuff sane? Would it be insane or sensible to add a copytarget-bofa-fishy check, that overrides the haveEffect?

packages/garbo/src/resources/yachtzee.ts
```diff
@@ -54,14 +56,22 @@ export const nonCinchNCs = () =>
         ? $item`Apriling band tuba`.dailyusesleft
         : 0);
 
+export const combatNCs = () =>
+  have($item`McHugeLarge left ski`)
+    ? Math.max(0, 3 - get("_mcHugeLargeAvalancheUses"))
+    : 0 +
```
spaghetti-squash: the indentation here scans as being nested rather than consecutive, to me. I know the linter chooses the indentation, but that makes me think it might actually _be_ nested rather than consecutive. If you put parenthesis around the first section, so it's

```suggestion
  (have($item`McHugeLarge left ski`)
    ? Math.max(0, 3 - get("_mcHugeLargeAvalancheUses"))
    : 0) +
```
does the formatter keep the parenthesis and change the indentation?

packages/garbo-lib/src/wanderer/pearl.ts
Shiverwarp: From my understanding, the progress on these resets at rollover, not on ascension

packages/garbo/src/resources/aprilingband.ts
Shiverwarp: Yeah that was the source of my question about pillkeeper earlier lol, I don't have a pillkeeper so not sure about setting that one up

packages/garbo/src/resources/seaDent.ts
Shiverwarp: I don't think dire warren is ever a wanderer location as far as I'm aware.

I have no idea if using wave updates `$location.environment` but I suspect it does not.

I don't think there's a reason to check anything other than the Fishy count, because we aren't gaining fishy while we're doing our copy targets, only when we're killing cows

packages/garbo/src/resources/yachtzee.ts
Shiverwarp: Nice catch

packages/garbo/src/outfit/dropsgearAccessories.ts
Shiverwarp: Something like `lowestMonsterHealth`? Would we include ML in it somehow?

packages/garbo/src/farmingStrategy.ts
```diff
@@ -233,22 +227,34 @@ const THE_CORAL_CORRAL: FarmingStrategyOptions = {
...
+    }
+
+    // Cows are tough! Let's delevel them to be safe
+    baseMacro.delevel().tryHaveItem($item`cow poker`);
 
-      return redTaffyWorth()
-        ? macro.tryItem($item`pulled red taffy`).meatKill()
-        : macro.meatKill();
+    if (redTaffyWorth()) {
+      baseMacro.tryItem($item`pulled red taffy`);
```
spaghetti-squash: this could be an `externalIf`; the banish is just weird because it doesn't work as a typeguard.

packages/garbo/src/mood.ts
```diff
@@ -87,12 +91,44 @@ export function meatMood(
         ? $skill`Ur-Kel's Aria of Annoyance`
         : $skill`Fat Leon's Phat Loot Lyric`,
     );
-  } else {
-    // This could be optimized a bit better using the new setup
-    mood.skill($skill`Donho's Bubbly Ballad`);
+  }
+
+  if (FarmingStrategy.location === $location`The Coral Corral`) {
```
spaghetti-squash, line 98: Do we think this is cowo-specific logic (in which case it should maybe?  live somewhere closer to farming strategy) or underwater-specific logic (in which case this check can be FarmingStrategy.isUnderwater() and this code can remain here)

packages/garbo/src/mood.ts
```diff
@@ -87,12 +91,44 @@ export function meatMood(
...
+      useSkill($skill`Donho's Bubbly Ballad`, 50 - get("_donhosCasts"));
+    } else if (
+      get("_donhosCasts") < 50 &&
+      availableDonhoTurnsFromSkill > estimatedGarboTurns()
+    ) {
+      mood.skill($skill`Donho's Bubbly Ballad`);
+    } else {
+      useSkill($skill`Donho's Bubbly Ballad`, 50 - get("_donhosCasts"));
+      mood.potion($item`recording of Donho's Bubbly Ballad`, 0.2 * meat);
+    }
```
spaghetti-squash, line 120: it MIGHT make sense to add this to the `mood.skill` logic innately, wherever it handles AT stuff

packages/garbo/src/tasks/yachtzee/index.ts
```diff
@@ -91,6 +95,53 @@ export const yachtzeeQuest: Quest<AlternateTask>[] = [
...
+          Macro.skill($skill`Launch spikolodon spikes`).step(
+            FarmingStrategy.macro(context),
+          ),
+        ),
+        prepare: farmPrepare,
+        post: farmPost,
+        turns: () => 2 * Math.max(0, 5 - get("_spikolodonSpikeUses")), // Need one turn to cast the NC, and one to do the yachtzee
+        sobriety: "sober",
+        spendsTurn: true,
+      },
```
spaghetti-squash, line 121: I think it makes more sense for this not to be a separate task, but rather to fold parka/spikolodon stuff into either barfOutfit or the outfit layer for farmingstrategy, and to similarly fold spikolodon spikes into either the base meatKill or the macro layer for farmingstrategy. I'm not married to this, but that's my instinct.

packages/garbo/src/tasks/yachtzee/index.ts
```diff
@@ -112,13 +112,11 @@ export const yachtzeeQuest: Quest<
           return barfOutfit(baseOutfit.spec());
         },
         do: () => FarmingStrategy.location,
-        combat: new GarboStrategy((context) =>
-          Macro.skill($skill`Launch spikolodon spikes`).step(
-            FarmingStrategy.macro(context),
-          ),
+        combat: FarmingStrategy.combat.clone().startingMacro(
```
spaghetti-squash: This is a regression. Fix whatever the actual issue is.

Shiverwarp:
PR has been split up and merged separately

PR #2711, opened 2026-09-04
access FarmingStrategy in a different way; implement methods on the class

spaghetti-squash:


packages/garbo/src/familiar/barfFamiliar.ts
```diff
@@ -279,10 +264,11 @@ function extraValue(
 ) {
   const targetValue = totalFamiliarValue(target);
   const meatFamiliarValue = totalFamiliarValue(meat);
-  const jelly =
-    farmingStrategy().location === $location`Barf Mountain`
-      ? $item`stench jelly`
-      : Item.none;
+  const jelly = FarmingStrategy.monsters().every(
+    (monster) => monster.attackElement === $element`Stench`,
```
Ignose: What if we extracted the available jelly? Does that even matter? I guess stench jelly is the only sneak.

PR #2723, opened 2026-09-09
Fix circular dependencies (there are still a bunch left but these were the easy ones)

gausie:
Fixed a bunch of circular dependencies, mostly by removing index.ts barrels


spaghetti-squash (approved):
a bummer that we're losing all our index files, but these are the sacrifices we have to make

PR #2724, opened 2026-09-09
Add IDMG Farming

Ignose:
I don't know if we actually want to merge this, but people requested it.

### DIFF
--- a/packages/garbo/src/config.ts
+++ b/packages/garbo/src/config.ts
@@ -5,6 +5,7 @@ import { $item, $items, $monster } from "libram";
 export enum FarmingMethod {
   BARF_MOUNTAIN,
   THE_CORAL_CORRAL,
+  UR_FARM,
 }
 
 const workshedAliases = [
@@ -268,6 +269,7 @@ You can use multiple options in conjunction, e.g. "garbo nobarf ascend"',
             options: [
               [FarmingMethod.BARF_MOUNTAIN, "barf mountain"],
               [FarmingMethod.THE_CORAL_CORRAL, "sea cows"],
+              [FarmingMethod.UR_FARM, "IDMG"],
             ],
           },
           (value) => {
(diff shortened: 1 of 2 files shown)


packages/garbo/src/config.ts
```diff
@@ -268,6 +269,7 @@ You can use multiple options in conjunction, e.g. "garbo nobarf ascend"',
             options: [
               [FarmingMethod.BARF_MOUNTAIN, "barf mountain"],
               [FarmingMethod.THE_CORAL_CORRAL, "sea cows"],
+              [FarmingMethod.UR_FARM, "IDMG"],
```
Rinn, line 272: please use like TREASURY or COBBS_KNOB_TREASURY rather than UR_FARM so it matches the other farming methods naming convention

packages/garbo/src/farmingStrategy.ts
```diff
@@ -264,8 +264,40 @@ function currentStrategy(): FarmingStrategyOptions {
...
+    case FarmingMethod.UR_FARM:
+      return UR_FARM;
+
     case FarmingMethod.BARF_MOUNTAIN:
     default:
       return BARF_MOUNTAIN;
   }
 }
+
+const UR_FARM: FarmingStrategyOptions = {
```
Rinn, line 276: i think this should increase combat rate to try and avoid https://wiki.kingdomofloathing.com/Meat_Score! and https://wiki.kingdomofloathing.com/Treasury_Chest

packages/garbo/src/farmingStrategy.ts
```diff
@@ -264,8 +264,40 @@ function currentStrategy(): FarmingStrategyOptions {
...
+  stasisRounds: 20,
+  asdonEffect: $effect`Driving Observantly`,
+  ensureBarfAccess: false,
+  baseMeat: 98,
```
Rinn, line 280: how is this value calculated?

Rinn (commented):
a very optimistic farming method but who am i to judge

packages/garbo/src/farmingStrategy.ts
spaghetti-squash: yeah increasingly I think baseMeat should be derived either from targetMonster or the zone itself, and just be a method on the class

Ignose:
I decided I don't care enough to PR this one actually

PR #2726, opened 2026-09-10
misc improvements (mostly cowo related)

tottington:
### Changes

- **`garbo_skipOverdrunkAdventures`** (default off): stops garbo adventuring with Drunkula's wineglass while overdrunk. Everything that costs no turn still runs, so the day is finished  afterwards. Without this, I would often attempt to go fight sea cows while overdrunk to poor results.

- **`garbo_autoWishProfitThreshold`** (default `-1`, always ask): skips the wish confirmation dialog when expected profit, after paying for the wish, is at least this many meat.

- **Time-Spinner refusal handling:** garbo now reads the offer list (`<select name="monid">` in choice 1196) before submitting a travel, escapes the choice when one is refused, and latches the source off for the run. Without the latch `available()` re-offers forever. An open choice blocks every later equipment change, so this previously surfaced as `Error: Failed to maximize properly!` Also fixes `targetInCombatQueue()` substring-matching `sea cow` against a queued `sea cowboy`.

- **Meat accessories priced against a stale location:** `usingThumbRing()` called `setLocation(Barf Mountain)` and never restored it, leaving every later `loc()`/`zone()`/`env()` modifier resolving against Barf Mountain. This would be bad if you're say, running in a different location like cowo wants to. It is memoized and called from diet and potion setup and persists for the run; it surfaced for me as PirateRealm setup maximizing against the wrong zone. Now wrapped in the existing `withLocation()` helper.

- **PirateRealm setup:**  two changes: Improving the 40 adventure failsafe a bit. Additionally, the setup outfit now maximizes `-mus/-mys/-mox`. `DebuffPlanner` only spends potions and effect removal, it didn't consider equipment, so stat gear the maximizer picks is flat overhead. This cropped up in a stat being pushed over 100 while I had the eyepatch on due to flat +mus.

### DIFF
--- a/packages/garbo/src/config.ts
+++ b/packages/garbo/src/config.ts
@@ -246,6 +246,11 @@ You can use multiple options in conjunction, e.g. "garbo nobarf ascend"',
           help: "This is used only when autoUserConfirm is true, will automatically use resources (such as pocket wishes, 11-leaf clovers, etc) up to this threshold to source a target monster for chaining before requesting user interference.",
           default: 1,
         }),
+        autoWishProfitThreshold: Args.number({
+          setting: "garbo_autoWishProfitThreshold",
+          help: "If 0 or greater, skip the confirmation dialog and wish for your copy target automatically whenever we have copies banked but no way to start a chain, and the expected profit (in meat, after paying for the wish) is at least this large. Negative values (the default) always ask. Genie fights remain capped at 3 per day.",
+          default: -1,
+        }),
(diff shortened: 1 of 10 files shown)


spaghetti-squash:
This has merge conflicts, and also is a lot of distinct changes for one pull request. You'll have an easier time getting them merged as separate PRs.

tottington:
noted, closing to break these out

PR #2727, opened 2026-09-10
Add garbo_skipOverdrunkAdventures

tottington:
Opt-in toggle (default false) to stop garbo adventuring with Drunkula's wineglass while overdrunk. Everything that costs no turns still runs, so the day is finished. 

Currently, I get beat up by cows when running cowo with the wineglass, and I'm unsure there's anything my account has access to that would stop that from happening, so this is a backstop against that. 

Additionally could be useful if one thinks their marginal VoA tomorrow will be more valuable than their marginal VoA today and is not already overcapping their 200 adventure limit, but that gets more complex. Perhaps a future improvement atop this.


spaghetti-squash:
Maybe I'm missing something but can't you just not run garbo after getting overdrunk?

tottington:
I think you're right. I think this is me mistakenly putting this feature into garbo when it'd be better served as a part of a wrapper.

PR #2729, opened 2026-09-10
Handle the Time-Spinner refusing to travel

tottington:
garbo currently submits choice 1196 unconditionally and calls `runCombat()`. When the monster is not on the game's recent-fight list the travel is declined, 1196 comes straight back with no fight, and garbo returns leaving mafia inside the open choice.

This blocks every later equipment change, so this surfaced as `Error: Failed to maximize properly!` in `barfOutfit → barfFamiliar → … → dress`, which is unclear as to what's the actual error here, that we couldn't find our target. 

It also does not self-correct. `available()` gates on `combatQueue`, which is isn't always in line with the Time-Spinner's list, and the phantom `_timeSpinnerMinutesUsed` increment is corrected back when mafia re-reads 1195, so the gate goes true again and it loops.

**Three parts:**

1. **Read the offer list before submitting.** It is a `<select name="monid">` in choice 1196, rather than approximating from `combatQueue`.
2. **Escape the choice on refusal.** 1196 is absent from `canWalkFromChoice()` and needs its explicit "Maybe Later"; 1195 *is* present, so any non-choice request drops it. Escaping one can land on the other, so it drains in a bounded loop and aborts if still stuck.
3. **Latch the source off for the run.** Refusals spend no minutes, so `available()` would otherwise re-offer forever. The latch is module state, so it resets if you rerun garbo. Currently, I think this handling could be improved, but I'm not sure what else we could do here to prevent just trying to go back into the choice again and failing to find the monster we're looking for. 

Also fixes `targetInCombatQueue()` substring-matching: `combatQueue` is `"; "`-joined, so `.includes("sea cow")` matched a queued `sea cowboy`.


packages/garbo/src/target/fights.ts
```diff
@@ -262,22 +266,86 @@ export const chainStarters = [
...
+  while (handlingChoice() && attempts++ < 3) {
+    const choice = lastChoice();
+    if (choice === TRAVEL_TO_A_RECENT_FIGHT) {
+      runChoice(2); // Maybe Later
+    } else if (choice === SPINNING_YOUR_TIME_SPINNER) {
+      visitUrl("main.php");
+    } else {
+      break;
+    }
+  }
```
spaghetti-squash: I'm not sure I understasnd why this is structured as a loop. Shouldn't it always be the case that "travel to a recent fight" choice 2 brings you to "spinning your time spinner"?

packages/garbo/src/target/fights.ts
```diff
@@ -286,9 +354,28 @@ export const copySources = [
...
+          }
+          if (offered === null) {
+            print(
+              "Could not find the Time-Spinner's recent-fight list on the page; attempting the travel anyway.",
+              HIGHLIGHT,
+            );
+          }
           visitUrl(
-            `choice.php?whichchoice=1196&monid=${globalOptions.target.id}&option=1`,
+            `choice.php?whichchoice=${TRAVEL_TO_A_RECENT_FIGHT}&monid=${globalOptions.target.id}&option=1`,
```
spaghetti-squash: I know this started as a visitUrl, but might as well change it to a `runChoice` here while we're in the area.

spaghetti-squash (commented):
I think this bloats the time-spinner code enough that it makes sense to extract these functions (and optionally other shared code) into a resource file instead of having it all live in fights.

spaghetti-squash (approved):

PR #2737, opened 2026-09-14
Add Copperhead as another farming method

Ignose:


packages/garbo/src/index.ts
```diff
@@ -141,6 +142,13 @@ export function main(argString = ""): void {
     globalOptions.prefs.farmingMethod = FarmingMethod.BARF_MOUNTAIN;
   }
 
+  if (
+    globalOptions.prefs.farmingMethod === FarmingMethod.THE_COPPERHEAD_CLUB &&
+    (!have($familiar`Red-Nosed Snapper`) || !(step("questL11Shen") === 999))
```
spaghetti-squash, line 147: These should be refactored into a requirement parameter on the farming strategy objects

packages/garbo/src/farmingStrategy.ts
```diff
@@ -265,11 +267,49 @@ const THE_CORAL_CORRAL: FarmingStrategyOptions = {
...
+  combat: new GarboStrategy(
+    () => Macro.meatKill(),
+    () =>
+      Macro.if_(
+        `(monsterid ${globalOptions.target.id}) && !gotjump && !(pastround 2)`,
+        Macro.meatKill(),
+      ).abort(),
+  ),
+
+  post: () => {
```
spaghetti-squash, line 298: I don’t know that this is the best place for this logic to occur; we don’t want it ”fighting” with any other snapper logic, and there’s no reason it should be contingent on completing a task. Realistically, it should live in daily familiar tasks.

packages/garbo/src/farmingStrategy.ts
```diff
@@ -265,11 +267,49 @@ const THE_CORAL_CORRAL: FarmingStrategyOptions = {
   }),
 };
 
+const THE_COPPERHEAD_CLUB: FarmingStrategyOptions = {
+  stasisRounds: 20,
+  asdonEffect: $effect`Driving Observantly`,
+  ensureBarfAccess: false,
+  baseMeat: 200, // Mob Penguin Capo
+  location: $location`The Copperhead Club`,
+  banishMonsters: $monsters`fan dancer, Copperhead Club bartender, ninja dressed as a waiter, waiter dressed as a ninja`,
```
spaghetti-squash, line 276: This is a long list of banish monsters; I don’t think we should keep our abort logic for failing to banish in this circumstance.

packages/garbo/src/farmingStrategy.ts
```diff
@@ -265,11 +267,49 @@ const THE_CORAL_CORRAL: FarmingStrategyOptions = {
...
+  targetMonster: $monster`sea cow`,
```
spaghetti-squash, line 277: Hotthonk; also, I think that, like basemeat, we probably want banish monsters to by default be derived as “every monster except the target”, maybe?

packages/garbo/src/farmingStrategy.ts
```diff
@@ -265,11 +267,49 @@ const THE_CORAL_CORRAL: FarmingStrategyOptions = {
   }),
 };
 
+const THE_COPPERHEAD_CLUB: FarmingStrategyOptions = {
+  stasisRounds: 20,
+  asdonEffect: $effect`Driving Observantly`,
```
spaghetti-squash, line 272: This should be moved to be the default value of this

packages/garbo/src/farmingStrategy.ts
```diff
@@ -265,11 +267,49 @@ const THE_CORAL_CORRAL: FarmingStrategyOptions = {
...
+  ensureBarfAccess: false,
```
spaghetti-squash, line 273: This should be moved to be the default value of this

packages/garbo/src/farmingStrategy.ts
```diff
@@ -265,11 +267,49 @@ const THE_CORAL_CORRAL: FarmingStrategyOptions = {
...
+  outfit: ({ banish }) => {
+    const banishItem = banish?.equip;
+    if (banishItem) {
+      print(`Planning to banish equipping ${banishItem?.name}`);
+    }
+
+    return banishItem ? { equip: [banishItem] } : {};
+  },
+
+  combat: new GarboStrategy(
```
spaghetti-squash, line 289: We should refactor the combat parameter to be a bare () = Macro rather than a combat strategy, and we should make meatkill the default

packages/garbo/src/farmingStrategy.ts
```diff
@@ -265,11 +267,49 @@ const THE_CORAL_CORRAL: FarmingStrategyOptions = {
...
+  stasisRounds: 20,
+  asdonEffect: $effect`Driving Observantly`,
+  ensureBarfAccess: false,
+  baseMeat: 200, // Mob Penguin Capo
+  location: $location`The Copperhead Club`,
+  banishMonsters: $monsters`fan dancer, Copperhead Club bartender, ninja dressed as a waiter, waiter dressed as a ninja`,
+  targetMonster: $monster`sea cow`,
+  shouldOlfact: false,
+
+  outfit: ({ banish }) => {
```
spaghetti-squash, line 280: We should extract this and make it a reused constant

packages/garbo/src/farmingStrategy.ts
Rinn: also it should be [Mob Penguin Capo](https://wiki.kingdomofloathing.com/Mob_Penguin_Capo) right?

packages/garbo/src/farmingStrategy.ts
Ignose: It is! The farming strategy doesn't really work without banishing everything, is universally worse than running barf mountain. Maybe we need a check for making sure enough banishes exist?

Issue #2716, opened 2026-09-06
[BUG] Cowo does not ensure maximization and fails


Zasqdddas:
[zasqdddas_20260906 cowo.txt](https://github.com/user-attachments/files/31875796/zasqdddas_20260906.cowo.txt)
when setting the garbo_farmingMethod to cowo, garbo crashes out with the error:
JavaScript exception: Error: Failed to maximize properly! (file:/home/leo/.kolmafia/scripts/garbage-collector/garbo.js#21094)

I suspect that the "+sea" is not working correctly for some reason, as when I try to re-run the maximizer, it suggests an outfit without a method of waterbreathing.  

My last few lines of output 
Maximizer: +sea, -tie, 24.836874999999974 Item Drop, 3.25 Meat Drop, -"equip Amulet of Perpetual Darkness", -"equip Spooky Putty ball", -"equip Spooky Putty leotard", -"equip Spooky Putty mitre", -"equip Spooky Putty snake", -"equip angelbone chopsticks", -"equip angelbone dice", -"equip angelbone totem", -"equip broken champagne bottle", -"equip devilbone corset", -"equip devilbone greaves", -"equip devilbone rosary", -"equip dice-shaped backpack", -"equip papier-masque", -"equip papier-mitre", -"equip smoke ball", -"equip stinky fannypack", -"equip ¶5868", -"equip ¶5869", -"equip ¶5872", -acc1, -acc2, -back, -buddy-bjorn, -crown-of-thrones, -familiar, -off-hand, -weapon, 131.19 "bonus June cleaver", 135.06 "bonus Mr. Cheeng's spectacles", 140.26 "bonus can of mixed everything", 153.85 "bonus cursed magnifying glass", 180.74 "bonus lucky gold ring", 20 "bonus toy Cupid bow", 241.67 "bonus mafia thumb ring", 2679.57 "bonus small peppermint-flavored sugar walking crook", 2900 "bonus bat wings", 312.53 "bonus aviator goggles", 361 "bonus Daylight Shavings Helmet", 56.7 "bonus rake", 56.7 "bonus tiny rake", 7.75 "bonus Portable Laughing Stock", 88.17 "bonus Cup of 13s"
Loading character status...
Preference _concoctionDatabaseRefreshes changed from 2015 to 2016
Maximizing...
385 combinations checked, best score 28,355.59 (FAIL)
Saving outfit: Script Outfit 1
Preference _concoctionDatabaseRefreshes changed from 2016 to 2017
[...]

PR #2740, opened 2026-09-15
I think this solves the problem with Cowo JOAT

Ignose:


Ignose:
Solves #2716

spaghetti-squash:
I don't believe these changes are sufficient to fix the problem at hand.

1. JOAT is still hardcoded as a meatFamiliar choice due to its conditional leprechaun multiplier.
2. `Familiar` is no longer a valid return type for the Cache Key function, so we should remove it
3. The problem fundamentally lies on outfit code, right? We're force-equipping the candle while also passing `+sea`. In the event that we still end up selecting JOAT (because, for example, it's still hardcoded as the meat familiar), the bug will recur. So while changes to barfFamiliar are necessary for optimal behavior, we still need changes to outfit code in order to ensure that the crash doesn't happen.
4. While JOAT is the only familiar to straight up force-equip its familiar equipment, I think this probably still messes with TCB familiars a lot.

